PR #44, opened 2017-01-06
Implemented custom images

### LINKED ISSUE #42: Add support for custom docker images
Cross is lacking support for custom docker images. Custom images should be used either by default or optionally.

## Option 1:
Cross looks on launch for an image called `<target>:<tag>` where `<tag>` is optionally specified with a flag `--tag` (if none then default tag, i.e `latest` or latest tagged version). If successful, cross will later launch docker with the specified image. If unsucessful, cross will load `japaric/<target>:<tag>` (`<tag>` is also the default here). Cross will also take a `docker-image` flag to fully specify what image to be used, i.e using `other-user/foo:<tag>` or `<local-image>:<tag>` where the user simply gives the full path to cross.

## Option 2:
Cross does the same as Option 1 but doesn't look for another image but `japaric/<target>:<tag>`unless any is specified.

### Things to consider
- `cargo --tag` is currently undefined, however it could be used in the feature, using `--docker-tag` could be safer to use for future compatability reasons.
- Instead of just using flags for `cross --docker-image myimage`, also consider using enviromental variables for this, possible even using [dotenv]. 
- When user specifies a full path to an image, parse it with `docker_image.replace("{target}", target)` so that when using `cross --target=<target> --docker-image={target}:v0.1.0`, the docker-image will be expanded and set to `<target>:v0.1.0`.

[dotenv]: https://crates.io/crates/dotenv

Emilgardis:
See issue #42

~~These changes follow option 1 on the issue and has~~
~~support for expansion as such: `--docker-image {target}`.~~
**EDIT::** Reworked to work with Cross.toml instead


src/main.rs
```diff
@@ -220,6 +220,10 @@ impl From<Host> for Target {
     }
 }
 
+pub type Tag = Option<String>;
+
+pub struct DockerImage(Option<String>);
```
Emilgardis: I should probably just make this into a newtype like I did with `Tag`.

japaric:
Thanks for the PR, @Emilgardis.

I rather not extend the CLI; I want to keep it exactly the same as Cargo. Could you move the image overrides into a Cross.toml file that sits next to Cargo.toml? That file would apply to all the `cross` calls inside the project and that way you don't have to pass e.g. `--docker-image` to each `cross` call.

I'm thinking of something like this:

``` toml
# Cross.toml
[target.armv7-unknown-linux-gnueabihf]
image = "user/repo:tag"  # or "user/repo" (implies :latest) or just "repo" (doesn't look up the image in DockerHub)
```

Emilgardis:
Fair enough. I have however a question about your example here. 

> just "repo" (doesn't look up the image in DockerHub)
Does that mean cross should parse `image` and see if it is of form `user/repo` or just `repo`.
(Or is it ok to actually let docker do the checking, i.e docker tries to use `-it repo`)

japaric:
Just passing the `image` string as it is to Docker is fine. Docker will print an error if it can't find it.

Emilgardis:
Ok, I'll remove the check on local images aswell, I feel like it is uneeded with this change.

Emilgardis:
I've implemented the images now using a Cross.toml file.

japaric:
Thanks for updating the PR, @Emilgardis.

I'd rather not expose an easy way to use a different tag of the `japaric/$TARGET` image as e.g. tag v0.1.2 is not guaranteed to work with cross v0.1.1 and viceversa. And I definitely don't want an easy way for someone to use :latest with an stable cross release. No end user should ever use :latest as it is a moving target (it changes everytime a PR lands in master).

If you remove target.$target.tag, I'll r+.

japaric:
Thanks again, @Emilgardis!

@homunkulus r+

homunkulus:
:pushpin: Commit a32f4f0 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit a32f4f0 with merge 938efc9...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cross/builds/189939952)
Approved by: japaric
Pushing 938efc99dd2dd6c07fc1e4a1b6bcb23b9960e407 to master...

PR #54, opened 2017-01-13
Individually set OPENSSL_INCLUDE_DIR and OPENSSL_LIB_DIR instead of OPENSSL_DIR

jedisct1:
This makes cross compatible with a wider range of openssl versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ install:
   - curl https://sh.rustup.rs -sSf |
     sh -s -- -y --default-toolchain $TRAVIS_RUST_VERSION
   - source ~/.cargo/env
-
+  - cargo clone -V || cargo install cargo-clone --vers 0.1.0
+  
 script:
   - bash ci/script.sh
 
--- a/README.md
+++ b/README.md
@@ -125,9 +125,10 @@ where libc was extracted.
   ever write to `$OUT_DIR` and never modify `$CARGO_MANIFEST_DIR` though.
 
 - Versions `0.7.*` and older of the `openssl` crate are NOT supported. `cross`
-  supports `openssl` via the `OPENSSL_DIR` "feature", which seems to have been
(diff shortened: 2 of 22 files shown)


japaric:
Nice! Do you know what's the oldest version of openssl that can be supported with this patch?

Also, could you add a test for this in `ci/script.sh`. Something like this:

```
# you'll need to `cargo install cargo-clone -f` in ci/install.sh first
$ cargo clone openssl --vers 0.6.0

$ cd openssl

$ cross build --target $TARGET
```

jedisct1:
The oldest version supporting this appears to be 0.3.6.

jedisct1:
I can confirm that this supports openssl down to version 0.4.0.

But openssl 0.4.0 is not compatible with current rust versions any more (it was released around the time `std::io` was reorganized, and depends on `std::old_io`).

jedisct1:
The pull request was updated with a test, against openssl-sys 0.5.5.

jedisct1:
```
# you'll need to `cargo install cargo-clone -f` in ci/install.sh first
```

Do you mean in trust's `install.sh` script?

japaric:
@homunkulus r+

Thanks, @jedisct1!

homunkulus:
:pushpin: Commit 0d53943 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 0d53943 with merge 0d53943...

japaric:
> Do you mean in trust's install.sh script?

There's no install.sh here, I forgot. I added that to Travis' install phase.

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cross/builds/191804308)

japaric:
Looks like that newer versions of openssl need the old OPENSSL_DIR variable. So I just opened #55 readding that variable.

Closing this PR in favor of that one.

PR #77, opened 2017-03-11
Add support for whitelisting environment variables to pass through (closes #76)

### LINKED ISSUE #76: Propagate environment variables
I noticed that `cross` clears all environment variables when building a project with Docker.  This breaks some build scripts and code that uses the `env!` macro.

dflemstr:
This adds a fairly straight-forward section to `Cross.toml` like so:

```toml
[build.env]
whitelist_all = true
# ...or...
whitelist = ["PATH", "RVM_VERSION", "LD_LIBRARY_PATH"]
```

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -78,6 +78,26 @@ pub fn run(target: &Target,
         docker.args(&["-e", &format!("QEMU_STRACE={}", strace)]);
     }
 
+    if let Some(toml) = toml {
+        if toml.env_whitelist_all()?.unwrap_or(false) {
+            for (var, _) in env::vars() {
+                // Only specifying the environment variable name in the "-e"
+                // flag forwards the value from the parent shell
+                docker.args(&["-e", &var]);
+            }
+        } else {
+            for var in toml.env_whitelist()? {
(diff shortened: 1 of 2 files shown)


japaric:
Thank you for the PR, @dflemstr, and sorry for taking so long to review.

The implementation looks good to me but I'm a bit concerned about `whitelist_all` as that makes more likely to break the operation of Cross by overriding one of the env variables its docker container internally uses. Ideally, I'd like Cross to warn you if you were to override one of such internal variables but that would make the implementation of this feature way more complicated.

So, how about this: let's implement just `whitelist` and document this feature on the README noting that Cross requires some env variables to be set to certain values, perhaps also list the common ones in the README, to work correctly and that you should think twice about overriding those.

What do you think?

kamalmarhubi:
I would like this to enable passing TravisCI env vars into the build. Let me know if I can help nudge this along!

Susurrus:
@kamalmarhubi I think all we need to do is remove the `whitelist_all` portion of this PR and then I think it's good to merge. Would you be able to do that in the short-term? Otherwise I can push this over the finish line.

PR #78, opened 2017-03-11
Preserve project working directory

dflemstr:
Some build scripts rely on references into the working directory of the project being built, so the project directory should be the same inside and outside the docker container.


japaric:
Thanks for the PR, @dflemstr. Could you elaborate on this:

> Some build scripts rely on references into the working directory of the project being built

Surely those build scripts are using relative paths rather than absolute paths; otherwise, they wouldn't be portable across machines.

> so the project directory should be the same inside and outside the docker container.

That it's already the case, right? Everything within the project directory gets mounted into the Docker container. The project can't refer to anything outside of it because it's not mounted. Changing on which directory the Cargo project will get mounted inside the docker cointainer won't change that.

jamesmunns:
@dflemstr: You may want to consider making a Cargo Workspace containing all of your code. Check out this documentation: https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html

Once you have made a cargo workspace, you can make the following call from the root of the workspace:

`cross build --manifest-path ./your_main_crate/Cargo.toml --target <CROSS TARGET>`

I was able to use this to build my crate with local source dependencies.

dflemstr:
The reason why this is needed is because I need to link some (C++) libraries that are built outside of the container. Sadly I can't do this with relative paths since the libraries are needed for a dependency far down the dependency path of my crate, and each dependency crate is built with its own manifest directory as the current working directory.

gnzlbg:
@dflemstr can you build those crates inside the container using a `build.rs` script ?

dflemstr:
@gnzlbg the building of the particular C++ codebase takes about 30-50 minutes so I would prefer to not do it on every cargo build if possible :)

gnzlbg:
@dflemstr so your build script then depends on how the directories outside of your crate are set up ?

Dylan-DPC:
Closing this as there are existing alternative ways of doing this. If you are still keen on including it, you can open a new PR.

PR #106, opened 2017-06-07
Add support for emscripten targets

malbarbo:
Emscripten 1.37.13 provides pre-build binaries, so we do not need to build it. 

Node 8.0.0 can exec wasm files, so we can have test support for both asmjs and wasm32. We use a workaround for https://github.com/kripken/emscripten/issues/4542.

### DIFF
--- a/docker/node-wasm
+++ b/docker/node-wasm
@@ -6,14 +6,42 @@ file="$(basename $1)"
 # Workaround for
 # https://github.com/kripken/emscripten/issues/4542
 
-if [ "$(basename $path)" != "deps" ]; then
-    path="$path/deps"
-fi
+# Consider a project with this struct
+# ├── src
+# │   ├── bin
+# │   │   └── prog.rs
+# │   └── lib.rs
+# ├── benches
+# │   └── b.rs
+# ├── examples
+# │   └── a.rs
+# └── tests
+#     └── t.rs
+#
+# We expect that the artifacts will be generated in
+# (where ? = release or debug)
+#
+# target/wasm32-unknown-emscripten/?/deps/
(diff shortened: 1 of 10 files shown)


malbarbo:
This was inspired by https://github.com/japaric/cross/pull/36, but I expected that the PR would be different that I wrote it from the beginning.

malbarbo:
Tested this with some local projects and it worked fine. Projects using libc generally fails, it seems that libc for emscripten targets are broken, I'm working on it.

docker/node-wasm
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+path="$(dirname $1)"
+file="$(basename $1)"
+
+# Workaround for
+# https://github.com/kripken/emscripten/issues/4542
+
+if [ "$(basename $path)" != "deps" ]; then
```
japaric: Could you add a comment here indicating what kind of paths we expect to see here and why we have to branch here and in the if below?

malbarbo:
Done.

japaric:
Thanks @malbarbo ❤️ 

@homunkulus r+

homunkulus:
:pushpin: Commit d1d7253 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit d1d7253 with merge d1d7253...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cross/builds/241423777?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing d1d72537a9052f205e9d1339e139e3751393f5a2 to master...

PR #113, opened 2017-07-09
Add armv5 support

jamesmunns:
Hey @japaric, heres a PR to integrate my work on ARMv5. Please let me know if you have any questions.

The changes to `cross` source should prevent the need for a `Cross.toml`, however as `jemalloc` and `unwind` are not available, it is still necessary to define a `Xargo.toml`, and make changes to `Cargo.toml` (as well as `main.rs` in Rust >= 1.20.0). These are detailed in `TARGET-NOTES.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,8 +191,8 @@ worst, "hang" (never terminate).
 | `arm-linux-androideabi`              | N/A    | 4.9     | 1.0.2k  | ✓   | N/A   |        |
 | `arm-unknown-linux-gnueabi`          | 2.19   | 4.8.2   | 1.0.2k  | ✓   | 2.8.0 |   ✓    |
 | `arm-unknown-linux-musleabi`         | 1.1.15 | 5.3.1   | N/A     |     | 2.8.0 |   ✓    |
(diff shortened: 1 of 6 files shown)


japaric:
@jamesmunns Thanks for the PR and sorry for not commenting earlier.

I'd prefer if the official target Just Worked and `cross` used that instead of creating a new target.

I think that the official target specification should be tweaked: the target does support atomic operations (the OS provides them) so max-atomic-width should be changed accordingly; it also seems reasonable to change the default allocator to system if jemalloc doesn't compile for the target (iirc, the default allocator will be changed to libc's anyways).

It's harder to make the case for abort = panic though; all the "std targets" support unwinding AFAIK and setting abort = panic means that you can't use the `test` crate either so you can't do unit testing. So, ideally, libunwind should be made work -- I don't know what the exact problems are in that case though; does the library not compile?

jamesmunns:
If https://github.com/rust-lang/rust/pull/44978 is merged, this PR will no longer be necessary.

homunkulus:
:umbrella: The latest upstream changes (presumably #132) made this pull request unmergeable. Please resolve the merge conflicts.

jamesmunns:
Closing as https://github.com/rust-lang/rust/pull/44978 has merged :tada:

PR #153, opened 2017-11-13
Add support to use qemu-system to run linux aarch64 binaries

malbarbo:
Fixes https://github.com/japaric/cross/issues/146


malbarbo:
@japaric If you consider this approach viable, we can write some docs add support for other archs in another PR.

docker/aarch64-unknown-linux-gnu/Dockerfile
```diff
@@ -12,15 +12,32 @@ RUN apt-get update && \
...
+RUN bash /linux-image.sh \
+        aarch64 \
+        gnu \
+        $MIRROR/dists/stretch/main/installer-arm64/current/images/netboot/debian-installer/arm64/linux \
+        $MIRROR/dists/stretch/main/installer-arm64/current/images/netboot/debian-installer/arm64/initrd.gz \
+        $MIRROR/pool/main/l/linux/linux-image-4.9.0-4-arm64_4.9.51-1_arm64.deb
+
+COPY linux-runner /
+
+ENV CROSS_RUNNER="qemu-user" \
```
malbarbo: The default runner is qemu-user, but the user can set CROSS_RUNNER=qemu-system to change the runner.

docker/linux-runner
```diff
@@ -0,0 +1,79 @@
...
+
+  qemu-system-$arch \
+    -m 1024 \
+    -smp 4 \
+    -nographic \
+    $opt \
+    -kernel /qemu/vmlinuz \
+    -initrd /qemu/initrd.gz \
+    -serial tcp:127.0.0.1:4444,server,nowait \
+    -virtfs local,path=/target,security_model=mapped,mount_tag=target \
```
malbarbo: We use virtfs to share /target with the guest running in qemu. The files do not need to be explicit copied!

docker/linux-runner
```diff
@@ -0,0 +1,79 @@
...
+
+# use a random prefix so we can differenciate the exit code tag from the
+# program output
+begin_tag="oach8aobiePheew1kahl-start"
+end_tag="oach8aobiePheew1kahl-end"
+
+begin=no
+
+# execute the program and then print the exit code
+echo "echo $begin_tag; $prog; echo $end_tag \$?" | nc -q -1 127.0.0.1 4444 | while read LINE; do
```
malbarbo: We can run $prog directly on the guest!

docker/linux-runner
gnzlbg: In `stdsimd` we use `objdump` from a procedural macro to check the assembly generated by SIMD intrinsics. `objdump` isn't typically installed by default in the distros that are usually loaded into `qemu-sytem`.

I think we could hack around this by copying whichever `objdump` we want to use into `target/` and passing its path in an environment variable to qemu system, but maybe there would be a better way of sharing resources and binaries with the code running in `qemu-system` ?

malbarbo: qemu-system is used to run the target binaries, the build happens in the docker container, which has `objdump`. Does this makes sense?

malbarbo:
Using `CROSS_RUNNER` is simpler, but we can consider using `CROSS_TARGET_..._RUNNER`, like cargo does.

docker/linux-runner
gnzlbg: I think so. So what happens if my target binary tries to start a new process that runs `objdump` ? Does the one from the docker container get picked up?

malbarbo: In this case there would be no `objdump`... The qemu-system image is really small, it is basically a busybox rootfs... If this is necessary (sorry if I misunderstood  you), I think you could put extra binaries in /target and use [passthrough](https://github.com/japaric/cross#passing-environment-variables-into-the-build-environemnt) to pass env variable to inform where the binaries are.

gnzlbg: I see, that's exactly what I meant in my first comment.

docker/linux-runner
```diff
@@ -0,0 +1,79 @@
...
+    -smp 4 \
+    -nographic \
+    $opt \
+    -kernel /qemu/vmlinuz \
+    -initrd /qemu/initrd.gz \
+    -serial tcp:127.0.0.1:4444,server,nowait \
+    -virtfs local,path=/target,security_model=mapped,mount_tag=target \
+    -append init=/init-alt > /dev/null &
+
+  sleep 5
```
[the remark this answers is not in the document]
malbarbo: I added a mechanism to wait until the OS finishes to boot. Also add boot messages including the time to boot.

docker/linux-runner
```diff
@@ -0,0 +1,79 @@
...
+# program output
+begin_tag="oach8aobiePheew1kahl-start"
+end_tag="oach8aobiePheew1kahl-end"
+
+begin=no
+
+# execute the program and then print the exit code
+echo "echo $begin_tag; $prog; echo $end_tag \$?" | nc -q -1 127.0.0.1 4444 | while read LINE; do
+  LINE=$(echo "$LINE" | tr -d '\r')
+  if [[ "$LINE" == "oach8aobiePheew1kahl-start" ]]; then
```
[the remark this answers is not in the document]
malbarbo: Fixed.

docker/qemu.sh
```diff
@@ -106,18 +109,35 @@ diff -ur qemu-2.10.0/linux-user/syscall_defs.h qemu-2.10.0.new/linux-user/syscal
...
+   case "$softmmu" in
+      softmmu)
+         targets="$targets,$arch-softmmu"
+         virtfs="--enable-virtfs"
+         ;;
+      "")
+         true
+         ;;
+      *)
+         echo "Invalid option $softmmu=$softmmu"
```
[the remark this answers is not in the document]
malbarbo: Fixed.

docker/qemu.sh
```diff
@@ -106,18 +109,35 @@ diff -ur qemu-2.10.0/linux-user/syscall_defs.h qemu-2.10.0.new/linux-user/syscal
...
-        --target-list=$arch-linux-user
+        $virtfs \
+        --target-list=$targets
     nice make -j$(nproc)
     make install
 
     # HACK the binfmt_misc interpreter we'll use expects the QEMU binary to be
     # in /usr/bin. Create an appropriate symlink
     ln -s /usr/local/bin/qemu-$arch /usr/bin/qemu-$arch-static
+    ln -s /usr/local/bin/qemu-system-$arch /usr/bin/qemu-system-$arch || true
```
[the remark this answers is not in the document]
malbarbo: It is not necessary. Thanks!

malbarbo:
I tried to address all pointed issues. For now I continue to use the runner term. I made each target have at least one runner, witch simplified the code but also brings the possibility of adding new generic runners (like a ssh remote runner, for example). I do not have preference for the runner term, but I think using node, native, etc as a kind of emulator is strange.

> Finally, before landing this we should document the downsides and advantages of using qemu-system instead of qemu-user. For example with qemu-system you can't modify the host filesystem (that's probably a good thing though) or invoke binaries in the host $PATH (what @gnzlbg mentioned about objdump) but threading support is better.

I  will need help with this. I have some difficulties writing in english, so it will take a lot of my time.

> P.S. Any idea of what other architectures could get qemu-system support? I think the main requirement is getting our hands on a compiled kernel and root filesystem for the target so it's up to what we can find and what's provided by distributions.

I've tried s390x, sparc64 and arm. I think Debian provides kernel and initrd for all rust supported archs, so we can offers qemu-system for many targets!

gnzlbg:
> I will need help with this. I have some difficulties writing in english, so it will take a lot of my time.

If you just write "the contents" in schematic way I can go over it; feel free to use words in your mother tongue if you don't know the english ones. 

> I've tried s390x, sparc64 and arm.

This is more like a wishlist for `stdsimd` but it would be cool if we could have:  

- some arm build bots:
  - `arm-unknown-linux-gnueabihf`
  - `armv7-unknown-linux-gnueabihf` 
  - `aarch64-unknown-linux-gnu`
- x86:
  - `i586-unknown-linux-gnu`     
  - `i686-unknown-linux-gnu`      
  - `x86_64-unknown-linux-gnu` with:
    - a TBM enabled AMD CPU to test TBM
    - an AVX-512 CPU to test AVX/AVX2/AVX-512..
- `powerpc64le-unknown-linux-gnu` to test AltiVec SIMD
- `mips64el-unknown-linux-gnuabi64` to test MSA SIMD
- `sparc64-unknown-linux-gnu` to test Sparc V SIMD

malbarbo:
I rewrote linux-image.sh and linux-runner to be more robust and make it easier to add other archs (I did preliminary test with 10 archs!). Instead of using debian installer kernel and initrd, we are using the normal debian packages and creating an initrd with busybox and some libs (this initrd is smaller and the boot is faster). Color is missing (adding ncurses-term package in the qemu image did not help).

@gnzlbg Thanks for your help. When the PR is ready to merge I will try to write some docs.  I will add the other targets after that.

@japaric After this rewrite I think a new review from zero is need... but it is for better!

malbarbo:
See the reason for this change in the last commit message.

This is a prerequisite for https://github.com/japaric/cross/pull/166

japaric:
Thanks for the patience @malbarbo @gnzlbg and sorry for the overdue review.

This looks great to me. Let's land it now and punt on documenting the internals.

bors r+

japaric:
@malbarbo The aarch64 build [got stuck](https://travis-ci.org/japaric/cross/jobs/351580908#L2755) at:

> Booting QEMU virtual machine with 2 cpus...

It's hard to tell what the problem was but if I had to guess I'd say that it could've been an OOM problem -- the Travis builders don't have much memory available.

gnzlbg:
@japaric thank you!

malbarbo:
After many tries i got it working on [travis](https://travis-ci.org/malbarbo/cross/builds/354564008). The problems seems to be with the output being buffered and the script hanging waiting for it. I've changed the approach and it worked.

But there is problems with others targets: [arm](https://travis-ci.org/malbarbo/cross/jobs/354511267) [i686](https://travis-ci.org/malbarbo/cross/jobs/354482479). Any ideas?

gnzlbg:
So `travis_wait ` can be used to extend the travis time limit for "no output" from 10 to 20 minutes.

The index out of bounds errors are a regression bug in rustc: https://github.com/rust-lang/rust/issues/49065

The linking errors do not ring a bell :/

Dylan-DPC:
@malbarbo can you resolve the conflicts? We can give this a run after that

malbarbo:
@Dylan-DPC I think it is better to wait for https://github.com/rust-embedded/cross/pull/217

If https://github.com/rust-embedded/cross/pull/217 get merged I will rebase and simplify this branch.

Dylan-DPC:
@malbarbo resolve the conflict so we can move ahead with this :+1:

malbarbo:
Rebased and adapted to https://github.com/rust-embedded/cross/pull/217. We are ready to go.

Dylan-DPC:
bors: try

Dylan-DPC:
bors: r+

malbarbo:
Network issue https://travis-ci.org/rust-embedded/cross/jobs/441838459#L2929

Dylan-DPC:
bors: r-

Dylan-DPC:
bors: r+

PR #157, opened 2017-11-14
add support for i586-unknown-linux-gnu

### LINKED ISSUE #156: Support i586-unknown-linux-gnu

gnzlbg:
Closes #156 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ worst, "hang" (never terminate).
 | `armv7-unknown-linux-gnueabihf`      | 2.15   | 4.6.2   | 1.0.2m  | ✓   | 2.8.0 |   ✓    |
 | `armv7-unknown-linux-musleabihf`     | 1.1.15 | 5.3.1   | N/A     |     | 2.8.0 |   ✓    |
 | `asmjs-unknown-emscripten` [4]       | 1.1.15 | 1.37.13 | N/A     | ✓   | N/A   |   ✓    |
+| `i586-unknown-linux-gnu`             | 2.23   | 5.3.1   | 1.0.2m  | ✓   | N/A   |   ✓    |
 | `i686-linux-android` [5]             | N/A    | 4.9     | 1.0.2m  | ✓   | N/A   |   ✓    |
 | `i686-pc-windows-gnu`                | N/A    | 6.2.0   | N/A     | ✓   | N/A   |   ✓    |
 | `i686-unknown-freebsd` [1]           | 10.2   | 5.3.0   | 1.0.2m  |     | N/A   |        |
--- a/docker/i586-unknown-linux-gnu/Dockerfile
+++ b/docker/i586-unknown-linux-gnu/Dockerfile
@@ -1,0 +1,26 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    cmake \
+    gcc \
+    libc6-dev \
+    make \
+    pkg-config
+
+COPY xargo.sh /
+RUN bash /xargo.sh
+
+COPY cmake.sh /
+RUN apt-get purge --auto-remove -y cmake && \
+    bash /cmake.sh 2.8.11
+
+COPY openssl.sh /
+RUN apt-get install -y --no-install-recommends \
+    g++-multilib && \
+    bash /openssl.sh linux-elf "" -m32
+
+ENV OPENSSL_DIR=/openssl \
+    OPENSSL_INCLUDE_DIR=/openssl/include \
+    OPENSSL_LIB_DIR=/openssl/lib
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
     - env: TARGET=arm-unknown-linux-musleabi                    STD=1                RUN=1
     - env: TARGET=armv7-unknown-linux-gnueabihf   CPP=1 DYLIB=1 STD=1 OPENSSL=0.5.5  RUN=1
     - env: TARGET=armv7-unknown-linux-musleabihf                STD=1                RUN=1
+    - env: TARGET=i586-unknown-linux-gnu          CPP=1 DYLIB=1 STD=1 OPENSSL=0.5.5  RUN=1
     - env: TARGET=i686-unknown-linux-gnu          CPP=1 DYLIB=1 STD=1 OPENSSL=0.5.5  RUN=1
     - env: TARGET=i686-unknown-linux-musl                       STD=1 OPENSSL=0.5.5  RUN=1
     - env: TARGET=mips-unknown-linux-gnu          CPP=1 DYLIB=1 STD=1 OPENSSL=0.5.5  RUN=1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add support for the `i586-unknown-linux-gnu` target.
+
 ## [v0.1.13] - 2017-11-08
 
 ### Added
--- a/src/main.rs
+++ b/src/main.rs
@@ -98,6 +98,7 @@ pub enum Target {
     ArmUnknownLinuxMusleabi,
     Armv7UnknownLinuxGnueabihf,
     Armv7UnknownLinuxMusleabihf,
+    I586UnknownLinuxGnu,
     I686UnknownLinuxGnu,
     I686UnknownLinuxMusl,
     Mips64UnknownLinuxGnuabi64,
@@ -199,6 +200,7 @@ impl Target {
             Target::ArmUnknownLinuxMusleabi |
             Target::Armv7UnknownLinuxGnueabihf |
             Target::Armv7UnknownLinuxMusleabihf |
+            Target::I586UnknownLinuxGnu |
             Target::I686UnknownLinuxGnu |
             Target::I686UnknownLinuxMusl |
             Target::Mips64UnknownLinuxGnuabi64 |
@@ -236,6 +238,7 @@ impl Target {
                        !triple.starts_with("i586") &&
                        !triple.starts_with("i686")
             }
+            Target::I586UnknownLinuxGnu |
             Target::I686UnknownLinuxGnu |
             Target::I686UnknownLinuxMusl |
             Target::X86_64UnknownLinuxGnu |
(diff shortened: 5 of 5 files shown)


gnzlbg:
I don't have a linux machine, is there a way to get the tests to run? Or do I have to test it manually (e.g. by forking cargo and trying to compile it with cross for i586 ?)

docker/i586-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:17.04
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    cmake \
+    gcc \
+    libc6-dev \
+    make \
+    file \
```
malbarbo: Is file dependency required?

docker/i586-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:17.04
```
malbarbo: This file is based on i686-unknown-linux-gnu, right? If the ubuntu version is kept in 12.04, the docker layers can be shared between i686-unknown-linux-gnu and i586-unknown-linux-gnu.

README.md
```diff
@@ -198,6 +198,7 @@ worst, "hang" (never terminate).
 | `armv7-unknown-linux-gnueabihf`      | 2.15   | 4.6.2   | 1.0.2m  | ✓   | 2.8.0 |   ✓    |
 | `armv7-unknown-linux-musleabihf`     | 1.1.15 | 5.3.1   | N/A     |     | 2.8.0 |   ✓    |
 | `asmjs-unknown-emscripten` [4]       | 1.1.15 | 1.37.13 | N/A     | ✓   | N/A   |   ✓    |
+| `i586-unknown-linux-gnu`             | 2.15   | 4.6.2   | 1.0.2m  | ✓   | N/A   |   ✓    |
```
japaric: the glibc version is wrong. Ubuntu 17.04 ships with glibc 2.24.

docker/i586-unknown-linux-gnu/Dockerfile
gnzlbg: I don't really know my way around docker. Should I downgrade the version in this file, upgrade the version in the `i686` docker file, or do I need to do something else to reuse the docker layers? 

(FWIW in the `stdsimd` crate we use 17.04 for everything and I am trying to replace our testing infrastructure there with `cross` now that you are adding support here for qemu-system).

docker/i586-unknown-linux-gnu/Dockerfile
gnzlbg: No, this is a c&p-error from the `i586` file in the `stdsimd` crate.

docker/i586-unknown-linux-gnu/Dockerfile
japaric: This should be as low as possible, i.e. 12.04 as the i686 target, to make the binaries produced with Cross compatible with legacy distributions like centos.

malbarbo:
I don't known how to run the test in a non Linux machine... Maybe you can install Linux in a virtual machine?

Instructions to run the tests in a Linux machine:

First you need to build the docker image:

```
./build-docker-image.sh i586-unknown-linux-gnu
```

Then you can run the tests:

```
TRAVIS_OS_NAME=linux TARGET=i586-unknown-linux-gnu CPP=1 DYLIB=1 STD=1 OPENSSL=0.5.5 RUN=1 bash ci/script.sh
```

japaric:
> I don't have a linux machine, is there a way to get the tests to run?

I can throw this into homu as a try build but test times will be in the order of hours due to the large number of travis builders. If you temporarily comment all the other travis builds I can start a try build.

docker/i586-unknown-linux-gnu/Dockerfile
gnzlbg: Fixed. This should make the version in the readme file correct right?

gnzlbg: Per @japaric 's comment below I've downgraded this to 12.04.

gnzlbg:
@japaric I have another PR with an ARM build, maybe after that one passes review I can merge both PRs and then we can do a homu build of both together?

README.md
gnzlbg: I've downgraded the ubuntu version to 12.04, so that this should now be correct.

japaric:
@gnzlbg Fine by me.

gnzlbg:
@japaric I've merged both PR here and disabled the other builds so that homu can be run.

gnzlbg:
@japaric is there a way to force qemu-arm to emulate an armv6 processor ?

gnzlbg:
@japaric I've removed the arm target since I don't really know how to do that without a machine where I can test it, the i586 parts should be good to go if you want to try with homu.

japaric:
> is there a way to force qemu-arm to emulate an armv6 processor ?

I think I tried setting a specific CPU before using QEMU_CPU but I think I didn't manage to get a SIGILL when supossedly emulating ARMv6 and running a binary that contained ARMv7-only instructions.

> I don't really know how to do that without a machine where I can test it

You can probably copy paste some of the Docker build stuff from rust-lang/rust. There they use crosstool-ng to build an ARMv6 toolchain.

@homunkulus try

homunkulus:
:hourglass: Trying commit 81f03cd510640c341df0005cc7f21b86cadae43a with merge 7e225bd6303bde0bc2f99e44270a47aa1763b883...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cross/builds/302095949?utm_source=github_status&utm_medium=notification)

gnzlbg:
It fails to link `cargo`: https://travis-ci.org/japaric/cross/builds/302095949#L8670

japaric:
That looks slightly familiar. IIRC that error is caused because you used a relatively new binutils / gcc when compiling the std facade but used an old binutils (ld) when linking a Rust binary against the facade. This has happened before with 32-bit targets and it was fixed upstream, perhaps this target was not fixed? I believe the solution was to downgrade the binutils version used in the docker image (in rust-lang/rust) where std is built.

gnzlbg:
@japaric you mean this here, and the comment above?

https://travis-ci.org/japaric/cross/builds/302095949#L8670

It seems like that is done for i686 but not for i586.

japaric:
I was referring to this linker error:

>   = note: /usr/bin/ld: /rust/lib/rustlib/i586-unknown-linux-gnu/lib/liballoc_jemalloc-208c05c6a7233cbb.rlib(jemalloc.pic.o): unrecognized relocation (0x2b) in section `.text.malloc_conf_init'

Specifically the "unrecognized relocation" you can probably find issues with that phrase in rust-lang/rust.

gnzlbg:
Ok I c&p the wrong link, this is the right one: https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-i586-gnu-i686-musl/Dockerfile#L36

japaric:
Yes, that's the issue. The problem is that std is built in Ubuntu 16.04 and we are using a linker from 12.04.

> ENV CFLAGS_i686_unknown_linux_musl=-Wa,-mrelax-relocations=no

This flag should also be applied to i586-gnu otherwise only i686-musl would be "supercompatible". This has to be fixed upstream.

gnzlbg:
https://github.com/rust-lang/rust/pull/45991

gnzlbg:
@homunkulus try

gnzlbg:
@japaric might be worth to retry this today or tomorrow, the fix was merged yesterday in rustc :)

japaric:
@homunkulus retry

homunkulus:
:hourglass: Trying commit 81f03cd510640c341df0005cc7f21b86cadae43a with merge 6a7bee866502c7ebeb9533155c6f57ff352c70e3...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cross/builds/303978506?utm_source=github_status&utm_medium=notification)

japaric:
linker error:

> /checkout/src/liballoc_jemalloc/../jemalloc/src/jemalloc.c:797: undefined reference to `secure_getenv'

that sounds like the std facade is being built against a newer libc than the one we are using. The easiest fix would be to bump our Ubuntu version to match the one that's being used in rust-lang/rust. The fix that would make i586 binaries more compatible would be to lower the libc version used in rust-lang/rust.

gnzlbg:
I've bumped the ubuntu version to 16.04 (the one used in rust-lang/rust). If that works we could try lowering it to 14.04 and see if that works.

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 7f3bfb6b42f2420806e6ba451abaa1e86f065fb5 with merge 04e7fecc00f8d3784ebcab5b051643e48247a2b1...

japaric:
@gnzlbg you can `try` this PR, and any other PR, by the way. It didn't work before because the branch was in a `try (failed)` stage. In that stage you have to `retry`. I think you should be able to retry `try`-ed PRs but I've never tested that before.

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cross/builds/303993985?utm_source=github_status&utm_medium=notification)
State: approved= try=True

japaric:
We shouldn't use a libc older than the one used by rust-lang/rust because there's the risk that a new libc symbol will be used in the std facade in the future and that will cause linker error due to us having an older libc without that symbol.

If you uncomment the other build jobs I'll r+ this.

gnzlbg:
@japaric I've rebased but I think the versions of libc in the readme are still wrong.

gnzlbg:
@japaric so i've updated the version of glibc in the readme to 2.23 and gcc to 5.3.1 (those seem to be the values for ubuntu 16.04). IIUC the openssl version is 1.0.2m for all right?

gnzlbg:
@homunkulus try

homunkulus:
:hourglass: Trying commit ae1f3d5a7f094969525fe9a6b53398abda635766 with merge 999cc8945a9a62157c23ae20d9a8fb8d8571b349...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cross/builds/304387395?utm_source=github_status&utm_medium=notification)
State: approved= try=True

japaric:
>  IIUC the openssl version is 1.0.2m for all right?

Yes. The OpenSSL is the same on all the images.

Thank you, @gnzlbg.

gnzlbg:
@japaric could you do a release of this? I'd like to move `stdsimd` to cross asap, even if that means loosing `arm-unknown-linux-gnueabihf` support temporarily for the time being.

japaric:
@gnzlbg  v0.1.14 has been released. You'll have to wait like 1 hour until propely tagged images make it to dockerhub and the release works properly. When [this build](https://travis-ci.org/japaric/cross/builds/305538558) is done the new release should work.

gnzlbg:
Thanks, I'll send a PR to `stdsimd` to start using `cross` probably tomorrow. I'll ping you in there.

PR #164, opened 2017-11-28
Upgrade 12.04-based containers to 14.04

### LINKED ISSUE #149: Unable to add packages for other architectures
The documentation states that it's possible to create derivative images that include additional packages required for building.  For example, libasound2-dev is required to build the asound-dev package.  README.md states that it is possible to do this:

````
FROM japaric/aarch64-unknown-linux-gnu:v0.1.4

RUN dpkg --add-architecture arm64 && \
    apt-get update && \
    apt-get install libfoo:arm64
````

However this doesn't actually work, because the distribution that is in use doesn't support other architectures:

````
xobs@nas /d/x/d/cross-build> docker run --rm -it japaric/aarch64-unknown-linux-gnu:v0.1.4 bash
root@a2a1fcfefc05:/# dpkg --add-architecture arm64 && \
>     apt-get update
Ign http://archive.ubuntu.com trusty InRelease
Get:1 http://archive.ubuntu.com trusty-updates InRelease [65.9 kB]
...
Err http://archive.ubuntu.com trusty-updates/restricted arm64 Packages
  404  Not Found [IP: 91.189.88.149 80]
Err http://archive.ubuntu.com trusty-updates/universe arm64 Packages
  404  Not Found [IP: 91.189.88.149 80]
...
W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty/restricted/binary-arm64/Packages  404  Not Found [IP: 91.189.88.149 80]

W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty/universe/binary-arm64/Packages  404  Not Found [IP: 91.189.88.149 80]

E: Some index files failed to download. They have been ignored, or old ones used instead.
root@a2a1fcfefc05:/#
````

ticky:
Aims to fix #149 for `armv7-unknown-linux-gnueabihf`, `i686-unknown-linux-gnu` and `x86_64-unknown-linux-gnu` containers, by upgrading them to Ubuntu 14.04, where multiarch is properly supported.

### DIFF
--- a/docker/armv7-unknown-linux-gnueabihf/Dockerfile
+++ b/docker/armv7-unknown-linux-gnueabihf/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:12.04
+FROM ubuntu:14.04
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
@@ -11,10 +11,6 @@ RUN apt-get update && \
 
 COPY xargo.sh /
 RUN bash /xargo.sh
-
-COPY cmake.sh /
-RUN apt-get purge --auto-remove -y cmake && \
-    bash /cmake.sh 2.8.11
 
 COPY openssl.sh qemu.sh /
 RUN apt-get install -y --no-install-recommends \
--- a/docker/i686-unknown-linux-gnu/Dockerfile
+++ b/docker/i686-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:12.04
+FROM ubuntu:14.04
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
@@ -11,10 +11,6 @@ RUN apt-get update && \
 
 COPY xargo.sh /
 RUN bash /xargo.sh
-
-COPY cmake.sh /
-RUN apt-get purge --auto-remove -y cmake && \
-    bash /cmake.sh 2.8.11
 
 COPY openssl.sh /
 RUN apt-get install -y --no-install-recommends \
--- a/docker/x86_64-unknown-linux-gnu/Dockerfile
+++ b/docker/x86_64-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:12.04
+FROM ubuntu:14.04
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
@@ -11,10 +11,6 @@ RUN apt-get update && \
 
 COPY xargo.sh /
 RUN bash /xargo.sh
-
-COPY cmake.sh /
-RUN apt-get purge --auto-remove -y cmake && \
-    bash /cmake.sh 2.8.11
 
 COPY openssl.sh /
 RUN apt-get install -y --no-install-recommends \


japaric:
Thanks for the PR.

Unfortunately, this change makes the binaries produced by cross less portable because it increases the glibc version of the (cross) C toolchains. For instance, the binaries produced with these changes won't run on Ubuntu 12.04 ("undefined reference to $SYMBOL" error at runtime).

We strive to make Cross-generated binaries as portable as the rustc binary shipped by rust-lang/rust so I won't be accepting this change. Sorry but you'll have to find another solution to #149. (Idea: perhaps there's some debian release that supports multiarch *and* also has an old glibc version?).

ticky:
Sorry to hear that. Unfortunately, that’s not really something I understand enough to contribute.

As it stands, the instructions for adding library dependencies which is [in the readme](https://github.com/japaric/cross/tree/5f0477004f00217ab72a34e2a1a5afc6a7c42dcc#custom-docker-images) simply _do not work_, for confusing reasons, on a whole bunch of common architectures. Adding another invisible dependency on a different OS seems like an even more confusing change!

It might be good to add the OS and version pair to the table in the Readme, so these differences are clear. Moving any of these to Debian may also mean package names change subtly.

briansmith:
What will happen when Ubuntu 12.04 becomes completely unsupported in April 2019 (see https://wiki.ubuntu.com/Releases#Extended_Security_Maintenance)? Is the plan to switch to CentOS or some other LTS that supports old glibc and is still supported?

Disasm:
As far as I can see, Rust uses Ubuntu 16.04 for `armv7-unknown-linux-gnueabihf` and CentOS 5 with `vault.centos.org` repository mirror for `i686-unknown-linux-gnu` and `x86_64-unknown-linux-gnu` targets. I hope we can update Dockerfiles for these targets.

Disasm:
cc @rust-embedded/tools

japaric:
To reiterate: we are not going to bump the glibc versions in the v0.1.x line
because that's a breaking change. We'll bump the glibc versions when we are
ready to release v0.2.0.

> As far as I can see, Rust uses Ubuntu 16.04 for `armv7-unknown-linux-gnueabihf`

But rust-lang/rust builds their own toolchains from scratch. For example, they
use [glibc 2.16] + gcc 5.2 for armv7-unknown-linux-gnueabihf. We use Ubuntu 12.04
which comes with glibc 2.15 for armv7-unknown-linux-gnueabihf. Bumping our glibc
beyond 2.16 (i.e. moving to 14.04 or 16.04) would be a breaking change so it
can't be done in the v0.1.x line.

[glibc 2.16]: https://github.com/rust-lang/rust/blob/082c86175fcf72c355e6a889956fbea59e65bcdb/src/ci/docker/dist-armv7-linux/armv7-linux-gnueabihf.config#L308

> What will happen when Ubuntu 12.04 becomes completely unsupported in April 2019

Nothing's going to happen to the existing releases; our tagged docker images
will still be there. If someone is concerned about using an image based on
Ubuntu 12.04 they are free to use a custom image, Cross supports this.

---

I should also add that we are *not* going to increase the CI load of this
repository because it's already ridiculously large (and affects the whole
rust-embedded org). This means:

- We are *not* going to build custom toolchains. Not here nor in another
  repository under the rust-embedded org. If a third party provides nicely
  packaged toolchains that meet our needs we'll gladly use them but we are not
  going to build them ourselves.

- We are *not* going to provide variants of existing images so no
  `armv7-unknown-linux-gnueabihf-with-recent-glibc`. One image per target is
  plenty of CI time already.

With that say I'm going to close this PR since it's not actionable until we have
decided what to do with glibc versions for v0.2.0. If someone feels like
proposing a plan of action for v0.2.0 please post it in an RFC issue.

PR #199, opened 2018-05-02
Update Ubuntu version to 18.04

bgermann:
The images that do not rely on an old glibc being available are updated to the new LTS version.

### DIFF
--- a/docker/aarch64-linux-android/Dockerfile
+++ b/docker/aarch64-linux-android/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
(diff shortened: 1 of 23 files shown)


Dylan-DPC:
HI Thanks for the PR.

We had a similar PR that changed from 12.04 to 14.04 but wasn't merged yet (#164). The comment explains it. I wonder if this PR is affected by the same issue.

bgermann:
It is not affected as it only touches the targets that do not build binaries that use glibc symbols.

bgermann:
Closing as this is starving and now partially applied in another PR. I will not put any work in this anymore.

PR #201, opened 2018-05-02
allow build.rs to write files to the volume

### LINKED ISSUE #145: Relax insistence that cargo project is mounted read-only
I'm having a go at cross-compiling [rust-c-ares](https://github.com/dimbleby/rust-c-ares).

To build the underlying C library, the first step is to run [buildconf](https://github.com/c-ares/c-ares/blob/master/buildconf) which:

* must be run from the root of the `c-ares` checkout
* does stuff in that checkout eg creating a `configure` script

... which is unfortunate for trying to get it to work with `cross`, because the file system is mounted read-only.

Any chance of relaxing that, perhaps via some opt-in option if you want to maintain current behaviour as default?

(Also: it would be nice if the Docker images would `apt-get install automake libtool m4` so that I didn't have to make my own docker image, but this is not a big deal).

gnzlbg:
Closes #145 .

### DIFF
--- a/ci/crates/build_rs_write/Cargo.toml
+++ b/ci/crates/build_rs_write/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "build_rs_write"
+version = "0.1.0"
+authors = ["gnzlbg <gonzalobg88@gmail.com>"]
+
+[dependencies]
--- a/ci/crates/build_rs_write/build.rs
+++ b/ci/crates/build_rs_write/build.rs
@@ -1,0 +1,8 @@
+fn main() {
+    let p = std::path::PathBuf::from("src/main.rs");
+    assert!(p.exists());
+    let mut cmd = std::process::Command::new("cp");
+    cmd.arg("src/main.rs")
+        .arg("lib.rs")
+        .output().unwrap();
+}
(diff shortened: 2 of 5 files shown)


gnzlbg:
@homunkulus try

jamesmunns:
Hey @gnzlbg, I know this had been discussed before, however at the time it was decided that it was better practice to do all building in the `target/` directory, using the `OUT_DIR` environment variable exposed to build.rs scripts. You can see an example of this [in one of my repos](https://github.com/jamesmunns/nrf52dk-sys/blob/master/build.rs#L13).

Since this does impose some problems for people that are porting existing makefiles, it might make sense to allow disabling this check, but I would prefer this be an explicit opt-in.

Would you consider adding a cli flag to toggle this change with a "loud" name, like `--allow-container-to-modify-files` (okay, thats maybe too long, but you get the idea hopefully), so that the current behavior is not changed by default?

Also for reference, only collaborators can trigger CI at the moment.

Also also, thanks for all the attention on this project lately!

gnzlbg:
I think it would be better to try to detect
this and suggest that everything must happen in the OUT_DIR, but I don’t
know a good place to try to catch this.

gnzlbg:
This is written in the readme, and I have no idea how to detect this easily, so I am closing it. I agree that it is better to never modify the project directory.

PR #217, opened 2018-10-09
Simplify Target to use only to cases: BuiltIn and Custom

malbarbo:
In the current form, cross needs to known every target its supports.
This makes things too restrict. For example, its not possible to use a
custom target image for a target that cross does not provides a docker
image (like aarch64-unknown-linux-musl, for example). Cross refuses to
execute saying it does not support the specified target. Besides that,
to add a new docker image the cross code needs to be changed.

This changes simplifies the Target enum to have only two variants:
BuiltIn and Custom. Cross needs to known if a target is a custom target
so it can call xargo, otherwise, all BuitIn targets can be uniformly
handled.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,6 @@
 use std::env;
 use std::error::Error;
-use std::fs::File;
+use std::fs::{read_dir, File};
 use std::io::Write;
 use std::path::PathBuf;
 use std::process::Command;
@@ -21,6 +21,11 @@ fn main() {
     File::create(out_dir.join("commit-info.txt"))
         .unwrap()
         .write_all(commit_info().as_bytes())
+        .unwrap();
+
+    File::create(out_dir.join("docker-images.rs"))
+        .unwrap()
+        .write_all(docker_images().as_bytes())
         .unwrap();
 }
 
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -53,3 +58,19 @@ fn commit_date() -> Result<String, Some> {
...
     }
 }
+
+fn docker_images() -> String {
+    let mut images = String::from("[");
+    let mut dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+    dir.push("docker");
+    for entry in read_dir(dir).unwrap() {
+        let path = entry.unwrap().path();
+        if path.is_dir() {
```
Dylan-DPC: we should be using Path(Buf) here instead of depending on string concatenation. Much safer.

Dylan-DPC: Ah :+1: the path variable misled me.

Dylan-DPC:
Appreciate a second review from @jamesmunns or @rust-embedded/tools

Dylan-DPC:
@malbarbo  Fails on one target
```
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-dragonfly` target may not be installed
```

malbarbo:
The target x86_64-unknown-dragonfly is not available on rustup, so it also needs xargo. Can we try again?

Dylan-DPC:
Sure. though i feel that having uses_xargo` and `needs_xargo` variables feels a odd and confusing to me. We can refactor it later though.

Dylan-DPC:
bors: r+

Dylan-DPC:
Your latest change broke every platform :smile:

malbarbo:
I've adjusted AvaliableTargets (see the commit message) and remove the needs_xargo flag.

Dylan-DPC:
Dragonfly strikes again. Same issue as we had previously

Dylan-DPC:
bors: try

malbarbo:
The same issue, but a different reason. I tested dragonfly locally. Let's see how is goes.

PR #230, opened 2018-11-13
Update emscripten

CryZe:
rustc switched to emscripten with LLVM 6. emscripten versions with LLVM 4 are not supported anymore. This upgrades emscripten to the same version as rustc is using in its CI.

### DIFF
--- a/docker/emscripten.sh
+++ b/docker/emscripten.sh
@@ -26,8 +26,8 @@ main() {
     export HOME=/emsdk-portable/
 
     ./emsdk update
-    ./emsdk install sdk-1.37.21-64bit
-    ./emsdk activate sdk-1.37.21-64bit
+    ./emsdk install sdk-1.38.15-64bit
+    ./emsdk activate sdk-1.38.15-64bit
 
     # Compile and cache libc
     source ./emsdk_env.sh


Dylan-DPC:
Thanks @CryZe 

bors: try

Dylan-DPC:
bors: try

CryZe:
That seems unrelated to my PR.

CryZe:
So it seems like there's some spurious errors in here. But there's an actual problem with my PR in that the CI uses an old rust compiler for the wasm32-unknown-emscripten target due to https://github.com/rust-lang/cargo/issues/4689

CryZe:
I would want this to be fixed, but in order for that to happen, the rustc version in the test needs to be bumped. However it is explicitly set to an old version, because cargo run is apparently broken with emscripten files in more recent rust versions. Can I just kill the cross run for the emscripten target and then this can be merged or how should this progress?

CryZe:
Can somebody tell me if I can just remove the tests for that target? The target is on the verge of dying anyway, so it's unlikely that anyone fixes the test runner bugs anytime soon. So I'd rather have this merged without the tests running than the target being broken on cross' side altogether.

reitermarkus:
bors retry

Issue #228, opened 2018-11-10
master is broken


gnzlbg:
The following targets fail to pass tests on master (see https://travis-ci.org/rust-embedded/cross/builds/453241344):

* s390x-unknown-linux-gnu 
* i686-unknown-freebsd 
* x86_64-unknown-freebsd 
* TARGET=i686-pc-windows-gnu

PR #231, opened 2018-11-16
temporarily disable broken targets

gnzlbg:


gnzlbg:
Can someone please add a cron job to run cross CI daily ? 

It takes 2 seconds to add such a cron job to travis, and it would allow us to detect when master breaks cross, as well as to save some headaches to those submitting PRs.

jamesmunns:
@gnzlbg I also added a daily cron job: https://screenshots.firefox.com/W3my389ZfDoiFoWr/travis-ci.org

gnzlbg:
Awesome, thank you @jamesmunns, it would be great if you would get an email on failure, and could open an issue on the repo so that we can try to quickly find which rust PR broke things.

jamesmunns:
@gnzlbg will do. Would you mind opening issues to fix each of the targets that are disabled here?

I'd like to close https://github.com/rust-embedded/cross/issues/228, and instead have one issue per arch that doesn't work.

gnzlbg:
Yeah I'll do that later today and close #228 myself.

Dylan-DPC:
bors: retry

gnzlbg:
So I've also disabled `asmjs`. The emscripten backend of rustc was updated recently (https://github.com/rust-lang/rust/pull/55626), and there have been a couple of related PRs since then, so maybe these broke it (emscripten is using an older LLVM version for asm.js).

Dylan-DPC:
@gnzlbg cool i'll create a tracking issue for that

Dylan-DPC:
bors: r+

Dylan-DPC:
I'd prefer to have a release soon so that people can use the tagged version instead of master which will be a lesser burden for us

gnzlbg:
What we need is a cron job every time there is a new nightly. Given that I just had to amend this PR because asmjs broke on nightly, I'd say a daily job is the minimum required. One can configure the cron jobs so that they not run if a job already ran on master on the same day.

Dylan-DPC:
@gnzlbg yes a daily cron is needed but that could mess with existing builds since each build takes roughly 2-3 hours, which means it could block other PRs. We could sort this out by setting a priority (bors makes it easy with `bors: p=..`) but that will mean ensuring we do it for every PR (or at least those that run during the cron) which won't scale well.

PR #244, opened 2018-12-28
Fix toolchain path on macOS.

reitermarkus:
This automatically installs the Linux toolchain on macOS, so it can be used by the Docker images, and then installs the targets for the Linux toolchain instead of the default one.

### DIFF
--- a/build-docker-image.sh
+++ b/build-docker-image.sh
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 set -ex
 
 run() {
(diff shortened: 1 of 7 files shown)


Dylan-DPC:
bors: try

Dylan-DPC:
oops, sorry for tthe delay

bors: r+

Dylan-DPC:
bors: r+

reitermarkus:
Would be great to have a new release so that “zero setup” is also true for macOS.

PR #263, opened 2019-03-21
Check for updates on base image during build

bltavares:
It is important to keep Docker images up-to date, as upstream images
could have important security fixes, even using the same tag and
version.

When building images, it is possible to pass in a `--pull` flag to check
if the base image has updates, and rebuild as necessary. This feature is
available [since 2014], and will still retain the image cache if there are
no updates on the base image.

Travis might need updated Docker version. It is documented how to update it on their docs if it becomes necessary.

[since 2014]: https://github.com/moby/moby/pull/9281

### DIFF
--- a/docker/linux-image.sh
+++ b/docker/linux-image.sh
@@ -46,7 +46,7 @@ main() {
             kernel=4.19.0-4-powerpc64
             debsource="deb http://ftp.ports.debian.org/debian-ports/ unreleased main"
             debsource="$debsource\ndeb http://ftp.ports.debian.org/debian-ports/ unstable main"
-            # sid version of dropbear requeries this dependencies
+            # sid version of dropbear requires these dependencies
             deps="libtommath1:ppc64 libtomcrypt1:ppc64 libgmp10:ppc64"
             libssl="libssl1.1"
             ;;
@@ -63,7 +63,7 @@ main() {
             kernel=4.19.0-4-sparc64
             debsource="deb http://ftp.ports.debian.org/debian-ports/ unreleased main"
             debsource="$debsource\ndeb http://ftp.ports.debian.org/debian-ports/ unstable main"
-            # sid version of dropbear requeries this dependencies
+            # sid version of dropbear requires these dependencies
             deps="libtommath1:sparc64 libtomcrypt1:sparc64 libgmp10:sparc64"
             libssl="libssl1.1"
             ;;
--- a/docker/linux-image.sh
+++ b/docker/linux-image.sh
@@ -33,21 +33,22 @@ main() {
             ;;
         powerpc)
             # there is no stretch powerpc port, so we use jessie
-            # use a more recent kernel from backports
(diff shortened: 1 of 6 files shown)


bltavares:
The images have been quite outdated, and while developing a GitHub Action similar to `goreleaser`, I've noticed that last Travis run has been 1 year ago.

There has been a recent commit on master this year, in January, which didn't trigger new updates. This could be likely due to the change on user host the code, and Travis not being properly configured for the new `rust-embedded` user.

I've also noticed that this project has changed maintainers, but the images still live under the previous Docker Hub account. I would be able to help change the code to point to the new user on Docker Hub, but I can't help with credential management and setup of Travis.

There are other issues (eg: https://github.com/rust-embedded/cross/issues/261) that have noticed the missing image as well. How could I help to make new images available?

Dylan-DPC:
Thanks 

bors: try

bltavares:
Some images had outdated packages not available on Debian repository anymore. I've pointed to existing packages on a follow up commit on this PR, in case you want to retry building it.

Dylan-DPC:
bors: try

bltavares:
I think it will be necessary to change the url of Dropbear to point to their new releases path.
I could take a look later this week.

bltavares:
@Dylan-DPC would you mind trying again? I've updated the url for `dropbear` instalation

Dylan-DPC:
@bltavares sure. some of the builds are failing due to a unrelated issue. Will still try this 

bors: try

docker/linux-image.sh
```diff
@@ -33,21 +33,22 @@ main() {
...
             ;;
         powerpc64)
             # there is no stable port
             arch=ppc64
-            kernel=4.19.0-1-powerpc64
+            kernel=4.19.0-4-powerpc64
             debsource="deb http://ftp.ports.debian.org/debian-ports/ unreleased main"
             debsource="$debsource\ndeb http://ftp.ports.debian.org/debian-ports/ unstable main"
-            # sid version of dropbear requeries this depencendies
+            # sid version of dropbear requeries this dependencies
```
spl:
```suggestion
            # sid version of dropbear requires these dependencies
```

docker/linux-image.sh
```diff
@@ -59,11 +60,12 @@ main() {
             ;;
         sparc64)
             # there is no stable port
-            kernel=4.19.0-1-sparc64
+            kernel=4.19.0-4-sparc64
             debsource="deb http://ftp.ports.debian.org/debian-ports/ unreleased main"
             debsource="$debsource\ndeb http://ftp.ports.debian.org/debian-ports/ unstable main"
-            # sid version of dropbear requeries this depencendies
+            # sid version of dropbear requeries this dependencies
```
spl:
```suggestion
            # sid version of dropbear requires these dependencies
```

reitermarkus:
Most of these changes are now in `master`.

PR #271, opened 2019-05-20
Minor improvements to wording

spl:
* Fix a typo in "QEMU bug sand"
* Add a link for QEMU
* Clarify and simplify various other things about the wording

The diff looks worse than the changes are. I tried to keep to the 80-column width since it looked like the rest of the text was that way. GitHub's “rich diff” seems to help here.


Dylan-DPC:
Thanks. Will merge this once we sort out some unrelated issues in the build which will block this PR.

Issue #229, opened 2018-11-11
Remove OpenSSL


briansmith:
I propose that all the OpenSSL functionality be removed from cross. I noticed a lot of the issues in the cross issue tracker are related to cross trying to maintain OpenSSL; all of these issues would just go away if OpenSSL support were removed. And, the experience of using cross for applications that do not use OpenSSL would be improved.

It's not clear why cross includes OpenSSL for some (but not all) targets in the first place. Making OpenSSl easier to use seems out of scope for a tool like this. The rust-openssl crate already has support for building a vendored copy of OpenSSL if it isn't available. If that mechanism and other ways of installing OpenSSL don't work well, that's really OpenSSL's and/or rust-openssl's problem.

In any case, I would like to use cross specifically for applications that often make some effort to avoid depending on OpenSSL. Unfortunately, cross's OpenSSL stuff gets in the way here. For example, does a Rustls application, specifically one that uses the webpki-roots crate for root certificates, work correctly if OpenSSL isn't installed and/or was never installed?

(FWIW, I am the creator of [*ring*](https://github.com/briansmith/ring) and [webpki](https://github.com/briansmith/webpki).)

therealprof:
Big agreement from my side here.

TotalKrill:
I actually used this because it meant i did not have to compile a version of openssl myself. So would be sad to see it go. It is quite easy to remove yourself though, so a suggestion would be to create a docker image without it and upload and use that instead in the Cross.toml file.

SirVer:
I just ran into issues because OpenSSL was removed from cross. Every single of my tools I use cross (and trust) for are using OpenSSL and since this happens somewhere deep down in my dependencies, I have little options of not using OpenSSL. Removing it means that cross has lost all value to me overnight :(.

While I fully acknowledge that the maintenance burden is on the maintainers of this repository and I can see that OpenSSL is a pita to maintain, the users of https://github.com/japaric/trust and many cross users that just want to do a web request in their code will have a lot more to deal with now. I hope this move will be reconsidered going forward.

CryZe:
Yeah this is likely going to cause significant problems for me as well.

surban:
Is it possible to revert to an older Docker image that still include OpenSSL?

Emilgardis:
@surban to use another image, use following config: https://github.com/rust-embedded/cross/blob/master/README.md#custom-docker-images

skorobkov:
[...]

PR #278, opened 2019-07-05
WIP: Add clang & ssl

dpc:
I"m trying to build my project using `trust` (and thus `cross`) and after much investigation: it seems like I need clang and ( maybe, but I'm not sure) ssl-dev in cross containers. It's really common requirements to need `clang` in Rust projects, unfortunately. I guess ssl is probably already installed anyway, but I guess it doesn't hurt to make it explicit.

Would a PR of this form be accepted? If so I can go over all other targets and make similiar modifications. If there's something to fix, please tell me now, before I copy it over 20 or so times. Thanks!

### DIFF
--- a/docker/aarch64-unknown-linux-musl/Dockerfile
+++ b/docker/aarch64-unknown-linux-musl/Dockerfile
@@ -15,7 +15,12 @@ RUN apt-get update && \
     autoconf \
     make \
     file \
-    binutils
+    binutils \
+    openssl \
+    libssl-dev \
+    llvm-dev \
+    libclang-dev \
+    clang
 
 COPY xargo.sh /
 RUN bash /xargo.sh


docker/aarch64-unknown-linux-musl/Dockerfile
```diff
@@ -15,7 +15,12 @@ RUN apt-get update && \
     autoconf \
     make \
     file \
-    binutils
+    binutils \
+    openssl \
```
briansmith: I don't think we should be adding openssl to images implicitly. See issue #229.

reitermarkus:
Closing this in favour of https://github.com/rust-embedded/cross/issues/229. I think only adding `clang` would be acceptable, however.

PR #286, opened 2019-08-24
Refactor Travis config.

reitermarkus:


.travis.yml
```diff
@@ -1,7 +1,7 @@
-dist: trusty
```
therealprof: My understanding here is that trusty was used on purpose due to the ancient libc it uses. Is this not required anymore?

reitermarkus: This doesn't matter, because we want to test the Docker image, not the host system. Currently Travis will not actually use the `x86_64-unknown-linux-gnu` image but will run directly on the host, we should force it to test the image even if the host is the same target.

reitermarkus:
bors try

reitermarkus:
Added.

PR #166, opened 2017-12-03
Add  qemu-system support for more targets

malbarbo:
Add qemu-system support for the following targets:

- aarch64-unknown-linux-gnu
- armv7-unknown-linux-gnueabihf
- i686-unknown-linux-gnu
- mips-unknown-linux-gnu
- mipsel-unknown-linux-gnu
- mips64el-unknown-linux-gnuabi64
- powerpc-unknown-linux-gnu
- powerpc64-unknown-linux-gnu
- powerpc64le-unknown-linux-gnu
- s390x-unknown-linux-gnu
- sparc64-unknown-linux-gnu
- x86_64-unknown-linux-gnu

Running `cross test` fails for `s390x-unknown-linux-gnu` and `sparc64-unknown-linux-gnu`, but `cargo run` works. This maybe be a bug in qemu or rustc.

Debian does not have a port for `mips64-unknown-linux-gnuabi64`. Support for arm targets can be add later, but I think a custom kernel will be need.

musl targets can get qemu-system support latter.

PR #302, opened 2019-09-04
Release 0.1.15

reitermarkus:


malbarbo:
@reitermarkus When the qemu-system supporting was added in https://github.com/rust-embedded/cross/pull/153 and https://github.com/rust-embedded/cross/pull/166 no documentation was provided... It would be great if some documentation could be add before a new release (I can't write it myself now, but maybe someone can).

By the way, thanks for moving the project forward, you are doing a great work!

reitermarkus:
> Actually I'd rather prefer the good old fashioned ASCII quotes...

I think typographic quotes should be preferred for prose.

> I can't write it myself now

I'm not using the `cross run` functionality myself so I can't either. And I don't think this should be blocking a new release, which is long overdue. Would be great if you could do it if you have time in the future.

Dylan-DPC:
> When the qemu-system supporting was added in #153 and #166 no documentation was provided..

@malbarbo Yep sorry about that. The prs were dragging on for too long so didn't want to drag it further

therealprof:
> > Actually I'd rather prefer the good old fashioned ASCII quotes...
 
> I think typographic quotes should be preferred for prose.

These are english text files, often processed on the command line. Typographic quotes are a dime a dozen and will cause random problems for people working with those documents. Until there's a very good reason to deviate I'd rather we stick to plain ASCII.

reitermarkus:
> These are english text files

And that's exactly the reason why we should use quotes appropriate for English text.

This is a Markdown file, there is no good reason why this file would ever be processed by a tool which doesn't support UTF-8 or HTML escape codes.

Dylan-DPC:
Let's not waste time bike-shedding such small things, it's a release note, we can always modify it later

Dylan-DPC:
bors: r+

therealprof:
> Let's not waste time bike-shedding such small things, it's a release note, we can always modify it later

No, we cannot. If it's released it will part of the release. Apart from that the point is not to include a change that was not there but to **NOT** make an incorrect change.

therealprof:
@Dylan-DPC Also I'd appreciate if you did not simply dismiss a members concern.

PR #303, opened 2019-09-06
Add support for "x86_64-pc-windows-msvc" on Azure Pipelines.

DiXN:


src/main.rs
```diff
@@ -62,7 +62,7 @@ impl Host {
         } else if *self == Host::X86_64UnknownLinuxGnu {
             target.map(|t| t.needs_docker()).unwrap_or(true)
         } else if *self == Host::X86_64PcWindowsMsvc {
-            target.map(|t| t.needs_docker()).unwrap_or(false)
+            target.map(|t| t.triple() == "x86_64-pc-windows-gnu" && t.needs_docker()).unwrap_or(false)
```
reitermarkus: This will prevent anything other than `x86_64-pc-windows-gnu` from running in Docker.

src/main.rs
```diff
@@ -62,7 +62,7 @@ impl Host {
         } else if *self == Host::X86_64UnknownLinuxGnu {
             target.map(|t| t.needs_docker()).unwrap_or(true)
         } else if *self == Host::X86_64PcWindowsMsvc {
-            target.map(|t| t.needs_docker()).unwrap_or(false)
+            target.map(|t| t.triple() != "x86_64-pc-windows-msvc" && t.needs_docker()).unwrap_or(false)
```
reitermarkus: You can use `self.triple()` here.

PR #304, opened 2019-09-06
Fix `versioned_image_name`.

reitermarkus:


burrbull:
can you show before/after example?

reitermarkus:
Before, the variable was missing from this script because I moved it into `build-docker-image.sh`, but it is needed both in `build-docker-image.sh` and in `docker_push.sh`.

Issue #313, opened 2019-09-08
Cross v0.1.15 Breaks f32/f64 `to_bits` on PowerPC


Alexhuszagh:
The following code worked for cross v0.1.14, so any issues were introduced with the latest release.

# Targets

- powerpc-unknown-linux-gnu
- powerpc64-unknown-linux-gnu
- powerpc64le-unknown-linux-gnu

# Failing Code

```rust
pub fn main() {
    let f: f32 = 1e-45;
    println!("to_bits={:#b}", f.to_bits());

    let p = &f as *const f32;
    let p = p as *const i32;
    unsafe {
        println!("transmuted as integer={:#b}", *p);
    }
}
```

# Results

```bash
to_bits=0b100
transmuted as integer=0b1
```

The value of `f.to_bits()` is `4`, rather than `1`, as expected, since `1e-45` is the smallest denormal 32-bit floating point number. The direct type-pun through a pointer alias gets the correct value of `1`. [to_bits](https://github.com/rust-lang/rust/blob/master/src/libcore/num/f64.rs#L429) directly calls `mem::transmute`, which should just pun the bytes to a new type.

# Explanation

The binary representation for the IEEE-754 single-precision floating point number 1e-45 is:

```
0 | 00000000 | 00000000000000000000001
```

The first group is the sign bit, the second group is the exponent (including the hidden bit), and the third group is the significant digits (mantissa). The endianness of floats should be the same as integers on the architecture. Therefore, `f.to_bits()` should return `1`, instead, for some unknown reason, it returns `4`.

# Minimal Steps to Reproduce

1. Create a repository (`cargo new --bin sample`).
2. Copy the failing code into `main.rs`.
3. Run the binary on a PowerPC architecture (`cross run --target powerpc-unknown-linux-gnu`).

reitermarkus:
Is this the same QEMU bug as in https://github.com/rust-embedded/cross/issues/314?

Alexhuszagh:
@reitermarkus I doubt it's the same Qemu bug, but it appears to be another bug in Qemu. I'm attempting to diagnose which versions are currently affected now.

# Function

```rust
pub fn to_bits(f: f32) -> u32 {
    f.to_bits()
}
````

The x86 assembly is simple, just a move instruction, as we'd expect:

```asm
example::to_bits:
        mov     eax, dword ptr [esp + 4]
        ret
```

The PowerPC assembly is as follows:

```asm
example::to_bits:
        stwu 1, -16(1)
[...]

PR #316, opened 2019-09-09
Patch for breaking 32-bit float conversions on ppc (#313).

Alexhuszagh:
Use qemu-3.0.1 for all powerpc architectures, and 4.1.0 for everything else.

Note: @reitermarkus, this requires bash for the scripting language, not shell, which the Dockerfiles currently use. I'm not sure if this is a deal-breaker.

Closes https://github.com/rust-embedded/cross/issues/313.

### DIFF
--- a/docker/qemu.sh
+++ b/docker/qemu.sh
@@ -1,7 +1,18 @@
+#!/usr/bin/env bash
+
 set -ex
 
 main() {
     local version=4.1.0
+
+    # Qemu versions 3.10.0 and above break 32-bit float conversions
+    # on powerpc, powerpc64, and powerpc64le. Last known working version
+    # is 3.0.1.
+    # Upstream Issue:
+    #   https://bugs.launchpad.net/qemu/+bug/1821444
+    if [[ $1 == ppc* ]]; then
+        version=3.0.1
+    fi
 
     local arch=$1 \
           os=$2 \


reitermarkus:
> this requires bash

Yeah, that's fine. Let's add the `#!/usr/bin/env bash` shebang line to make that clear.

docker/qemu.sh
```diff
@@ -1,7 +1,17 @@
...
+    # Qemu versions 3.10.0 and above break 32-bit float conversions
+    # on powerpc, powerpc64, and powerpc64le. Last known working version
+    # is 3.0.1.
+    # Upstream Issue:
+    #   https://bugs.launchpad.net/qemu/+bug/1821444
+    if [[ $1 == ppc* ]]
+    then
+        local version=3.0.1
+    else
+        local version=4.1.0
```
reitermarkus: Define this first as a default and only change it in the PPC case.

Alexhuszagh: Done, I've squashed it back into a single commit so I don't litter the commit history.

docker/qemu.sh
```diff
@@ -1,8 +1,20 @@
...
     local version=4.1.0
 
+    # Qemu versions 3.10.0 and above break 32-bit float conversions
+    # on powerpc, powerpc64, and powerpc64le. Last known working version
+    # is 3.0.1.
+    # Upstream Issue:
+    #   https://bugs.launchpad.net/qemu/+bug/1821444
+    if [[ $1 == ppc* ]]
+    then
+        local version=3.0.1
```
reitermarkus: No need for the `local` here.

docker/qemu.sh
```diff
@@ -1,8 +1,20 @@
...
 main() {
     local version=4.1.0
 
+    # Qemu versions 3.10.0 and above break 32-bit float conversions
+    # on powerpc, powerpc64, and powerpc64le. Last known working version
+    # is 3.0.1.
+    # Upstream Issue:
+    #   https://bugs.launchpad.net/qemu/+bug/1821444
+    if [[ $1 == ppc* ]]
+    then
```
reitermarkus: Move this to the same line as `if`.

PR #319, opened 2019-09-10
Fix cross when stdin is not a TTY

CryZe:
When stdin is not a TTY, cross fails because it still passes `-i` to `docker run`, which makes Docker fail like so:
```
the input device is not a TTY
```
Apparently GitHub Actions doesn't have a TTY as stdin, so this is the fix to make cross work with GitHub Actions.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -62,7 +62,6 @@ pub fn register(target: &Target, verbose: bool) -> Result<()> {
     docker_command("run")
         .arg("--privileged")
         .arg("--rm")
-        .arg("-i")
         .arg("ubuntu:16.04")
         .args(&["sh", "-c", cmd])
         .run(verbose)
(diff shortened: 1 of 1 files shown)


src/docker.rs
```diff
@@ -59,10 +59,18 @@ pub fn register(target: &Target, verbose: bool) -> Result<()> {
...
+
+    let mut docker = docker_command("run");
+
+    docker
         .arg("--privileged")
-        .arg("--rm")
-        .arg("-i")
+        .arg("--rm");
+
+    if atty::is(Stream::Stdin) {
```
reitermarkus: Let's remove the `-i` here completely.

src/docker.rs
```diff
@@ -165,8 +173,12 @@ pub fn run(target: &Target,
         docker.arg("-t");
     }
 
+    if atty::is(Stream::Stdin) {
```
reitermarkus: Move this into the `if` above.

src/docker.rs
```diff
@@ -163,10 +163,13 @@ pub fn run(target: &Target,
 
     if atty::is(Stream::Stdout) && atty::is(Stream::Stderr) {
         docker.arg("-t");
+        if atty::is(Stream::Stdin) {
+            docker.arg("-i");
```
reitermarkus: I only meant moving this line, so make the arg above `-it`.

reitermarkus: I guess we should replace `atty::is(Stream::Stdout) && atty::is(Stream::Stderr)` with just `atty::is(Stream::Stdin)`.

CryZe: Oh whoops. Sure.

CryZe: Not checking for stdout and stderr anymore doesn't cause anymore problems?

reitermarkus: Not sure, maybe we should swap the order instead.

If `stdin` is a TTY, use `-i`, and additionally if both `stdout` and `stderr` are a TTY, also use `-t`.

CryZe: I'll change it to that, as that sounds safer than not checking at all.

PR #335, opened 2019-10-06
Update emscripten to 1.38.46-upstream

CryZe:
Rust updated its emscripten version again to a version that uses the upstream LLVM wasm backend that the other Rust wasm targets are also using. cross' docker has been updated accordingly: https://github.com/rust-lang/rust/pull/63649/files#diff-3334ef4d3e22f40ded09f2045806021f

### DIFF
--- a/docker/emscripten.sh
+++ b/docker/emscripten.sh
@@ -19,24 +19,19 @@ main() {
     done
 
     cd /
-    curl -L https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-portable.tar.gz | \
-        tar -xz
-    cd /emsdk-portable
+    git clone https://github.com/emscripten-core/emsdk.git /emsdk-portable
 
     export HOME=/emsdk-portable/
 
-    ./emsdk update
-    ./emsdk install sdk-1.38.15-64bit
-    ./emsdk activate sdk-1.38.15-64bit
+    ./emsdk install 1.38.46-upstream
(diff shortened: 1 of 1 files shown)


reitermarkus:
Have you looked into the CI failure?

CryZe:
I don't think I have as I had very little time. But it seems like it can't find the build when trying to access it now.

reitermarkus:
Weird, I also can't access it anymore now.

reitermarkus:
bors try

CryZe:
The failures seem to be:

For asmjs: ` note: shared:ERROR: wasm2js does not support source maps yet (debug in wasm for now)`
For wasm32: `= note: shared:ERROR: BINARYEN_TRAP_MODE is not supported by the LLVM wasm backend`

We'll have to dig deeper what they mean.

reitermarkus:
> The docker images don't contain the Rust compiler right?

Yes, the compiler is mounted via a volume.

> So essentially only the tests would need to be updated to use the nightly compiler instead.

Would this not break using stable Rust then?

CryZe:
Yeah we should probably release this with 1.40

reitermarkus:
Emscripten has been updated in https://github.com/rust-embedded/cross/pull/361 but the C++ test is failing and has been disabled for now. `libstdc++` cannot be found, no idea what the problem is.

CryZe:
I'm also not planning on using emscripten, so if this PR is still necessary for anyone, feel free to create a new one.

PR #339, opened 2019-10-26
Fix Android’s CPU features

nagisa:
The Rust Android target uses CPU features such as +popcnt and +sse4.2 which
are not available on Penryn. List the used CPU exactly the same way as
it is listed in the Rust target [here](https://github.com/rust-lang/rust/blob/master/src/librustc_target/spec/x86_64_linux_android.rs#L7).

Please publish an updated docker image sooner rather than later.


Dylan-DPC:
Tests failing

therealprof:
> Tests failing

powerpc and sparc build failures? Looks totally unrelated.

reitermarkus:
PowerPC and Sparc usually need the kernel version to be updated.

reitermarkus:
Thanks!

Issue #298, opened 2019-09-03
Don't panic if docker version cannot be parsed


davidgraeff:
I'm using [podman](https://podman.io/) instead of Docker (podman has a docker cli interface).
But podman, because it works daemonless, correctly reports no "Server API Version":

```
Version:            1.5.1
RemoteAPI Version:  1
Go Version:         go1.12.7
OS/Arch:            linux/amd64
```

A less intrusive version check that just warns would work as well, wouldn't it?

See
https://github.com/rust-embedded/cross/blob/e1470782db43fc737cf9b7f14748331228b8229d/src/docker.rs#L26

Thanks

reitermarkus:
Can you submit a PR?

davidgraeff:
Sure

oblitum:
Would be nice to have official/direct support for podman, it's the simplest solution for running containers rootless and safely (without being part of high privileged user groups). I suppose that to use it now one has to provide a docker command that redirects to podman.

davidgraeff:
It's not that easy unfortunately. That's why I haven't submitted a patch yet.
Try it out on your machine with this wrapper script:

On my fedora system SELinux restricts the container to a degree that nothing works. Disabling SELinux fails with a user-id mapping.

oblitum:
It works on ArchLinux, but needs some fixes. First, your wrapper worked for the image to be pulled, but then what cross executes afterwards cause errors:

```
❯ cross build --release --locked --target armv7-unknown-linux-gnueabihf
Executing run --userns host --rm --user 1000:1000 -e XARGO_HOME=/xargo -e CARGO_HOME=/cargo -e CARGO_TARGET_DIR=/target -e USER=francisco -e CROSS_RUNNER= -v /home/francisco/.xargo:/xargo:Z -v /home/francisco/.cargo:/cargo:Z -v /cargo/bin -v /home/francisco/Projects/samples/rust:/project:Z,ro -v /home/francisco/.rustup/toolchains/stable-x86_64-unknown-linux-gnu:/rust:Z,ro -v /home/francisco/Projects/samples/rust/target:/target:Z -w /project -i -t rustembedded/cross:armv7-unknown-linux-gnueabihf-0.1.16 sh -c PATH=$PATH:/rust/bin "cargo" "build" "--release" "--locked" "--target" "armv7-unknown-linux-gnueabihf"
Trying to pull docker.io/rustembedded/cross:armv7-unknown-linux-gnueabihf-0.1.16...
[...]

PR #344, opened 2019-11-09
Adding support for Podman

ostrosco:
- Via the new container_engine parameter in Cross.toml, the user can now
  specify whether they want to use Podman or Docker.
- Several changes to src/docker.rs to pass in different options based on whether
  docker or podman is being used. This also mandates the use of `--userns host`
  which does require bumping the minimum Docker version.
- Updates to the code base based on rustfmt and clippy recommendations.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -13,16 +13,31 @@ use crate::id;
 
 const DOCKER_IMAGES: &[&str] = &include!(concat!(env!("OUT_DIR"), "/docker-images.rs"));
 const DOCKER: &str = "docker";
+const PODMAN: &str = "podman";
 
-pub fn docker_command(container_engine: &str, subcommand: &str) -> Command {
-    let mut docker = Command::new(container_engine);
-    docker.arg(subcommand);
-    docker.args(&["--userns", "host"]);
-    docker
+fn get_container_engine() -> Option<&'static str> {
+    if which::which(DOCKER).is_ok() {
+        Some(DOCKER)
+    } else if which::which(PODMAN).is_ok() {
+        Some(PODMAN)
+    } else {
+        None
+    }
+}
+
+pub fn docker_command(subcommand: &str) -> Result<Command> {
+    if let Some(ce) = get_container_engine() {
+        let mut command = Command::new(ce);
+        command.arg(subcommand);
+        command.args(&["--userns", "host"]);
+        Ok(command)
+    } else {
+        Err("no container engine found; install docker or podman".into())
+    }
 }
 
 /// Register binfmt interpreters
-pub fn register(target: &Target, toml: Option<&Toml>, verbose: bool) -> Result<()> {
+pub fn register(target: &Target, verbose: bool) -> Result<()> {
     let cmd = if target.is_windows() {
         // https://www.kernel.org/doc/html/latest/admin-guide/binfmt-misc.html
         "mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc && \
@@ -32,14 +47,7 @@ pub fn register(target: &Target, toml: Option<&Toml>, verbose: bool) -> Result<(
             binfmt-support qemu-user-static"
     };
 
-    let mut container_engine = DOCKER.to_string();
-    if let Some(toml) = toml {
-        if let Some(engine) = toml.container_engine(target)? {
-            container_engine = engine;
-        }
-    }
-
-    docker_command(&container_engine, "run")
+    docker_command("run")?
         .arg("--privileged")
         .arg("--rm")
         .arg("ubuntu:16.04")
@@ -80,8 +88,7 @@ pub fn run(target: &Target,
     // container doesn't have write access to the root of the Cargo project
     let cargo_toml = root.join("Cargo.toml");
 
-    let mut runner = None;
-    let mut container_engine = DOCKER.to_string();
+    let runner = None;
 
     Command::new("cargo").args(&["fetch",
                 "--manifest-path",
@@ -89,13 +96,7 @@ pub fn run(target: &Target,
         .run(verbose)
         .chain_err(|| "couldn't generate Cargo.lock")?;
 
-    if let Some(toml) = toml {
-        runner = toml.runner(target)?;
-        if let Some(engine) = toml.container_engine(target)? {
-            container_engine = engine;
-        }
-    }
-    let mut docker = docker_command(&container_engine, "run");
+    let mut docker = docker_command("run")?;
 
     if let Some(toml) = toml {
         for var in toml.env_passthrough(target)? {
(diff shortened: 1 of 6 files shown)


reitermarkus:
Please move the style changes to a separate PR. It's really hard to review the actual changes with them mixed in.

ostrosco:
Sorry about that. I've reverted everything back save for the changes to add podman support.

src/docker.rs
```diff
@@ -130,31 +143,18 @@ pub fn run(target: &Target,
...
-            docker.args(&["-e", var]);
-        }
+    docker
+        .args(&["-e", &format!("CROSS_RUNNER={}", runner.unwrap_or_else(|| String::new()))])
+        .args(&["-v", &format!("{}:/xargo:Z", xargo_dir.display())])
+        .args(&["-v", &format!("{}:/cargo:Z", cargo_dir.display())]);
 
-        runner = toml.runner(target)?;
+    // Prevent `bin` from being mounted inside the Docker container. This is
+    // not required for Podman.
```
reitermarkus: Why isn't this required?

ostrosco: I think it's less that it wasn't required and more that if it's there, podman will fail. This was discovered in #298 where with this setting, podman would fail with:

`Error: error checking path "/cargo/bin": stat /cargo/bin: no such file or directory`

The user found the work-around of not specifying it. I'll acknowledge that I'm not certain if there are any consequences to not adding it.

reitermarkus: This is specifically so that this directory isn't mounted on macOS, since the container would then try running macOS binaries.

ostrosco: Thanks for the clarification. Would an acceptable solution be to only add /cargo/bin to the volumes if /cargo/bin already exists on the host?

reitermarkus: The problem is *that* it already exists on the host. Maybe we could glob the directory and mount all subdirectories explicitly except `bin`.

reitermarkus: Oh right, then we have the problem that we can't mount files like `config`.

reitermarkus: Does `-v /dev/null:/cargo/bin` work?

ostrosco: Yep, that seems to do the trick. Should we make this change for both Docker and Podman, or do we want to use the old command for Docker and this new one for Podman?

reitermarkus: Let's use the same for both.

ostrosco: Scratch that. Mounting to /dev/null worked on the CLI, but in `cross` failed with it.

```
Error: cannot configure rootless cgroup using the cgroupfs manager
creating file '/home/ostrosco/.local/share/containers/storage/overlay/34c319fde31eb6cdef406999cddcbcc0fc8886d6b4a9c93a05a3c0f56d6d4936/merged/cargo/bin': Is a directory: OCI runtime error
```

I assume this is because /dev/null isn't a directory. I'm not sure if /var/empty is a better choice?

ostrosco: Also I did find that the lack of support for anonymous volumes was a bug in podman that was fixed in the latest release a few days ago: https://github.com/containers/libpod/issues/4276

Not sure if the best approach is to only support this version?

reitermarkus: If `-v /cargo/bin` works with the latest `podman`, we should require that version instead of the hacky `/dev/null` workaround.

ostrosco: Sounds good. Just finished testing on Podman 1.6.3 and everything's working as expected. I just pushed an update to always add /cargo/bin.

src/docker.rs
```diff
@@ -14,45 +12,19 @@ use crate::extensions::CommandExt;
...
-        Version::parse(&format!("{}.0", version_string.trim()))
-            .expect("Cannot parse Docker engine version")
-    };
-
-    /// Version requirements for user namespace.
-    ///
-    /// # Panics
-    /// Panics if the parsing fails
-    static ref USERNS_REQUIREMENT: VersionReq = {
-        VersionReq::parse(">= 1.24")
```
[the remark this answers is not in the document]
ostrosco: Just added it to the README.

src/docker.rs
```diff
@@ -14,45 +12,19 @@ use crate::extensions::CommandExt;
...
-pub fn docker_command(subcommand: &str) -> Command {
-    let mut docker = Command::new("docker");
+pub fn docker_command(container_engine: &str, subcommand: &str) -> Command {
+    let mut docker = Command::new(container_engine);
     docker.arg(subcommand);
-    if USERNS_REQUIREMENT.matches(&DOCKER_VERSION) {
-        docker.args(&["--userns", "host"]);
-    }
+
+    // We always add the `---userns host` flag for compatibility with Podman.
```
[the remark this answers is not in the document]
ostrosco: Removed.

src/main.rs
```diff
@@ -329,6 +329,27 @@ impl Toml {
         }
     }
 
+        /// Returns the `target.{}.container_engine` part of `Cross.toml`
```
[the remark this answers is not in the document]
ostrosco: All fixed.

src/main.rs
```diff
@@ -329,6 +329,27 @@ impl Toml {
...
+                .as_str()
+                .ok_or_else(|| format!("target.{}.container_engine must be a string", triple))?
+                .to_string();
+            Ok(Some(value))
+        } else {
+            Ok(None)
+        }
+    }
+
+
```
reitermarkus: One empty line too many here.

src/docker.rs
```diff
@@ -101,17 +79,50 @@ pub fn run(target: &Target,
     // We create/regenerate the lockfile on the host system because the Docker
     // container doesn't have write access to the root of the Cargo project
     let cargo_toml = root.join("Cargo.toml");
+
+    let mut runner = None;
+    let mut container_engine = DOCKER.to_string();
```
reitermarkus: I think we should use the `which` crate to automatically detect whether `docker` or `podman` is installed.

ostrosco: I'm fine with this, but I have a concern for users who may have both installed. What should `cross` default to in these cases?

reitermarkus: I'd say default to Docker since that's the current behaviour.

src/docker.rs
```diff
@@ -14,41 +12,28 @@ use crate::extensions::CommandExt;
...
-    /// Panics if the parsing fails
-    static ref USERNS_REQUIREMENT: VersionReq = {
-        VersionReq::parse(">= 1.24")
-            .expect("Unable to parse version requirements")
-    };
+const DOCKER: &str = "docker";
+const PODMAN: &str = "podman";
+
+fn get_container_engine() -> Option<&'static str> {
+    if which::which(DOCKER).is_ok() {
```
reitermarkus: I think this function can be simplified as `which::which(DOCKER).or_else(|| which::which(PODMAN))`.

ostrosco: This would change the return type from `Option<&'static str>` to `Result<std::path::PathBuf, which::error::Error>`. Do we want to return the path to the command or just the command itself?

reitermarkus: Returning the path is fine.

reitermarkus:
Thanks, @ostrosco, great work!

ostrosco:
Thanks and you're welcome!

PR #348, opened 2019-11-15
Allow writing `Cargo.toml` in project directory.

reitermarkus:
The project directory was previously deliberately mounted as read-only since “well behaved crates should only ever write to `$OUT_DIR` and never modify `$CARGO_MANIFEST_DIR`”.  That however, breaks the `--locked` flag and is not worth the additional complexity.

Fixes https://github.com/rust-embedded/cross/issues/346.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,10 +284,6 @@ $ QEMU_STRACE=1 cross run --target aarch64-unknown-linux-gnu
   work because `cross` use docker containers only mounts the Cargo project so
   the container doesn't have access to the rest of the filesystem.
 
-- `cross` will mount the Cargo project as READ ONLY. Thus, if any crate attempts
-  to modify its “source”, the build will fail. Well behaved crates should only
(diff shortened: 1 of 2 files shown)


reitermarkus:
bors try

Emilgardis:
Would it be possible to instead mount the Cargo.toml file in docker as rw as an alternative?

reitermarkus:
No, since it's a file. You can only mount directories as a volume.

reitermarkus:
> if you think it's the only way

I mean there could be another way, but I'm pretty sure it wouldn't be worth the additional complexity for virtually no gain. If this really was a big enough problem, ´cargo` itself should be responsible for preventing self-modifying crates.

Emilgardis:
> 
> 
> No, since it's a file. You can only mount directories as a volume.

I'm pretty certain it's possible in docker.

reitermarkus:
> I'm pretty certain it's possible in docker.

You're right, seems like it is indeed possible.

Still, this wouldn't work in the case when `Cargo.lock` doesn't exist yet.

PR #353, opened 2019-11-27
Bump Ubuntu 14.04 to 16.04.

reitermarkus:
Images started failing to build with Ubuntu 14.04.

This is a breaking change, but the next version will be 0.2.0 in any case.

This brings up the question whether we should just bump all images to use 18.04 so we can have a common base image for all of them. To keep supporting the oldest version is clearly not feasible in the long run since images will keep breaking.


reitermarkus:
bors try

Issue #370, opened 2020-02-14
Support specifying a rustc channel via +channel argument


anti-social:
`cargo` supports specifying channels via:
```
cargo +nightly bench
```

At the moment `cross` just runs cargo on the host system. As a workaround it is possible to specify channel via an environment variable:
```
RUSTUP_TOOLCHAIN=nightly cross bench
```

Think it would be nice to have this feature in order to be more `cargo` friendly.

Alexhuszagh:
This is currently supported and was added in #370.

PR #371, opened 2020-02-14
[wip] #370 Support for explicit channel argument

anti-social:
Need a little time to write some tests (never dealt with azure).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,17 +1,15 @@
 use std::{env, path::PathBuf};
 
-use rustc_version::Channel;
-
 use crate::Target;
 use crate::cargo::Subcommand;
 use crate::errors::Result;
-use crate::rustc::{ChannelExt, TargetList};
+use crate::rustc::TargetList;
 
 #[derive(Debug)]
 pub struct Args {
     pub all: Vec<String>,
     pub subcommand: Option<Subcommand>,
-    pub channel: Option<Channel>,
+    pub channel: Option<String>,
     pub target: Option<Target>,
     pub target_dir: Option<PathBuf>,
 }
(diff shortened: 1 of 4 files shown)


src/rustc.rs
```diff
@@ -38,6 +38,33 @@ impl VersionMetaExt for VersionMeta {
     }
 }
 
+pub(crate) trait ChannelExt {
```
reitermarkus: Maybe this could be upstreamed?

reitermarkus:
CI should be working again after you rebase.

src/rustc.rs
anti-social: Not sure about that. `rustc_version` parses `stable` channel differently: https://github.com/Kimundi/rustc-version-rs/blob/1d692a965f4e48a8cb72e82cda953107c0d22f47/src/lib.rs#L185

reitermarkus: In that case I don't see a benefit in using `rustc_version::Channel` over just passing the string through to `cargo` in the container. Let's make `cargo` error on a wrong channel instead of duplicating that logic here.

src/cli.rs
```diff
@@ -2,16 +2,20 @@ use std::{env, path::PathBuf};
...
 pub struct Args {
     pub all: Vec<String>,
     pub subcommand: Option<Subcommand>,
+    pub channel: Option<String>,
     pub target: Option<Target>,
     pub target_dir: Option<PathBuf>,
 }
 
-pub fn parse(target_list: &TargetList) -> Args {
+pub fn parse(target_list: &TargetList) -> Result<Args> {
```
reitermarkus: I don't think the `Result` is needed here anymore.

reitermarkus:
Thanks!

PR #377, opened 2020-02-20
Avoid pulling if image already exists locally.

reitermarkus:
This avoids overwriting the local image when making changes to images locally.


reitermarkus:
bors retry

reitermarkus:
bors retry

PR #387, opened 2020-03-08
Docker in docker

urso:
When using docker-in-docker, for example in dev or CI containers, one
normally gives access to the host docker environment to the parent container.
If the container spins up another child container, and wishes
to share a directory, one needs to compute the correct paths from
view of the host system.

### How it Works
If the environment variable `CROSS_DOCKER_IN_DOCKER=true`, we call 
`docker inspect $HOSTNAME`, to learn about details of the current container.
The `Mounts` field contains information about all mounts from the host
into the current container. The `GraphDriver` field gives us information
about the storage driver and the location of the contains root directory on the host system. Based on these information we compute
a table of all container mounts (`source->destination`).

When starting a child container we adapt all paths to be mounted, that have
a prefix in the table, to start with source. For example when mounting a
dev containers `/usr/local/cargo` directory, we will actually mount
`/var/lib/docker/overlay2/<parent container id>/merged/usr/local/cargo`.

If the project itself is mounted into the parent container, we will not
use the overlay directory, but find the project path on the host system.

### Limitations 

Finding the mount point for the containers root directory is only
implemented for the overlayfs2 storage driver.

In order to access the parent containers rust setup, the child container
mounts the parents overlayfs. The parent must not be stopped before the
child container, as the overlayfs can not be unmounted correctly by
docker if the child container still accesses it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,49 @@ RUN dpkg --add-architecture arm64 && \
 $ docker build -t my/image:tag path/to/where/the/Dockerfile/resides
 ```
 
+### Docker in Docker
+
+When running `cross` from inside a docker container, `cross` needs access to
+the hosts docker daemon itself. This is normally achieved by mounting the
+docker daemons socket `/var/run/docker.sock`. For example:
+
+```
+$ docker run -v /var/run/docker.sock:/var/run/docker.sock -v .:/project \
+  -w /project my/development-image:tag cross build --target mips64-unknown-linux-gnuabi64
+```
+
+The image running `cross` requires the rust development tools to be installed.
+
+With this setup `cross` must find and mount the correct host paths into the
+container used for cross compilation. This includes the original project directory as
+well as the root path of the parent container to give access to the rust build
+tools.
+
+To inform `cross` that it is running inside a container set `CROSS_DOCKER_IN_DOCKER=true`.
+
+A development or CI container can be created like this:
+
+```
+FROM rust:1
+
+# set CROSS_DOCKER_IN_DOCKER to inform `cross` that it is executed from within a container
+ENV CROSS_DOCKER_IN_DOCKER=true
+
+# install `cross`
+RUN cargo install cross
+
+...
+
+```
+
+**Limitations**: Finding the mount point for the containers root directory is
+currently only available for the overlayfs2 storage driver. In order to access
+the parent containers rust setup, the child container mounts the parents
+overlayfs. The parent must not be stopped before the child container, as the
+overlayfs can not be unmounted correctly by Docker if the child container still
+accesses it.
+
+
 ### Passing environment variables into the build environment
 
 By default, `cross` does not pass any environment variables into the build
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,14 +1,16 @@
+use std::str::FromStr;
 use std::{env, path::PathBuf};
 
-use crate::Target;
 use crate::cargo::Subcommand;
 use crate::rustc::TargetList;
+use crate::Target;
 
 pub struct Args {
     pub all: Vec<String>,
     pub subcommand: Option<Subcommand>,
     pub target: Option<Target>,
     pub target_dir: Option<PathBuf>,
+    pub docker_in_docker: bool,
 }
 
 pub fn parse(target_list: &TargetList) -> Args {
@@ -27,7 +29,10 @@ pub fn parse(target_list: &TargetList) -> Args {
                     all.push(t);
                 }
             } else if arg.starts_with("--target=") {
-                target = arg.splitn(2, '=').nth(1).map(|s| Target::from(&*s, target_list));
+                target = arg
+                    .splitn(2, '=')
+                    .nth(1)
+                    .map(|s| Target::from(&*s, target_list));
                 all.push(arg);
             } else if arg == "--target-dir" {
                 all.push(arg);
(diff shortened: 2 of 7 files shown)


src/cli.rs
```diff
@@ -50,10 +52,15 @@ pub fn parse(target_list: &TargetList) -> Args {
         }
     }
 
+    let dind = env::var("CROSS_DOCKER_IN_DOCKER")
```
reitermarkus: Call this variable `docker_in_docker`.

reitermarkus: Also, the environment variable should be documented somewhere.

reitermarkus:
Please also add some tests for the parse functions.

src/cli.rs
urso: I renamed `dind` to `docker_in_docker` and did add a `Docker in Docker` section to the readme.

urso:
> Please also add some tests for the parse functions.

I didn't really find any unit tests. It seams that all tests are driven by `ci/test.sh`. Plus I'm not really familiar with your CI setup. Any recommendations on testing?

reitermarkus:
Add them as you would in any other Rust project, so they run when you run `cargo test`. Then we can add a `cargo test` before the `cargo install` in `ci/test.sh`.

urso:
Added some tests + added `cargo test` the azure pipeline instead of test.sh.

I don't think there is need to run the unit tests with each target in the pipeline. Would it make sense to enable/use stages to the azure pipeline (needs to be explicitely enabled, as its a feature in preview)?

src/cli.rs
```diff
@@ -41,19 +43,51 @@ pub fn parse(target_list: &TargetList) -> Args {
...
     Args {
         all,
         subcommand: sc,
         target,
         target_dir,
+        docker_in_docker,
+    }
+}
+
+fn parse_bool_env(val: Result<String, env::VarError>) -> bool {
```
reitermarkus: I don't think this needs to be a function, given that it is only used once.

urso: True. Given that there is no dedicated CI test for docker-in-docker, I at least wanted to test that the environment variable is parsed correctly.

TotalKrill:
Is there any progress on this? I need it for our CI setup

src/cli.rs
urso: @reitermarkus Would you prefer me to remove the test again, or keep it as is?

urso:
@TotalKrill I currently use `cross` like this in my development container: https://gist.github.com/urso/033a80a10a9b9656548134adeb373e8d#file-dockerfile-L46

README.md
```diff
@@ -123,6 +123,49 @@ RUN dpkg --add-architecture arm64 && \
...
+
+...
+
+```
+
+**Limitations**: Finding the mount point for the containers root directory is
+currently only available for the overlayfs2 storage driver. In order to access
+the parent containers rust setup, the child container mounts the parents
+overlayfs. The parent must not be stopped before the child container, as the
+overlayfs can not be unmounted correctly by docker if the child container still
```
reitermarkus, line 165: Docker

src/cli.rs
reitermarkus: Let's remove it. It's mainly testing `FromStr for bool`, which should be tested plenty already.

urso: Done.

README.md
urso: Done

reitermarkus:
Thanks, @urso! Could you please squash your last two commits?

urso:
Commits have been squashed.

reitermarkus:
Thanks!

src/docker.rs
```diff
@@ -179,3 +194,232 @@ pub fn image(toml: Option<&Toml>, target: &Target) -> Result<String> {
...
+fn dockerinfo_parse_user_mounts(info: &serde_json::Value) -> Vec<MountDetail> {
+    info.pointer("/0/Mounts")
+        .and_then(|v| v.as_array())
+        .map(|v| {
+            let make_path = |v: &serde_json::Value| PathBuf::from(&v.as_str().unwrap());
+            let mut mounts = vec![];
+            for details in v {
+                let source = make_path(&details["Source"]);
+                let destination = make_path(&details["Destination"]);
+                if source != destination {
```
romainreignier, line 260: @urso I am not sure to understand why this check is needed?
Actually, I have tried to run latest cross from github with docker-in-docker and if I mount my rust project as `-v $PWD:$PWD -w $PWD`, `source == destination` so it is not mounted in the second stage docker and `/project` is empty.
I might open an issue to discuss it further if you want.
Thank you.

urso: This is just a 'translation' table replacing the prefix that matches 'destination' with the prefix that matches 'source'. If `source == destination`, then this will be a noop.

If you open a shell in your development/CI container you can check the environment via `$ echo $CROSS_DOCKER_IN_DOCKER; docker inspect $HOSTNAME`. Running `cross <command> -v` will also print the docker commands as they will be executed. Very helpful for debugging.

romainreignier: Thanks, I did some debugging yesterday and by removing this check [in my fork](https://github.com/romainreignier/cross/commit/52757396d812291b18cfece817fb8895da3dc934) allowed me to use cross in a docker ran by Jenkins.

The issue here is that Jenkins mount the current workspace with the same path in Docker with
```
-v /path/to/my/workspace:/path/to/my/workspace
```

So `source == destination` but it is not a noop, it allows to mount files from host to the second docker stage.

I have managed to reproduce with:
```bash
# On host, create a file in current directory
touch testfile
# start a first docker stage
docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $PWD:$PWD -w $PWD debian:buster bash
# in the first docker stage, install docker
apt update && apt install -y --no-install-recommends docker.io
# check that the test file is there
ls
# it is here, OK
# get the merged path
merge_path=$(docker inspect `hostname` --format '{{.GraphDriver.Data.MergedDir}}')
# start the second docker stage
docker run -it --rm -v $merge_path/$PWD:/project debian:buster bash
# in the second docker stage, check that the test file is here
ls /project
# there is no file, NOK
# exit the second docker
exit
# new start with classic mount
docker run -it --rm -v $PWD:/project debian:buster bash
ls /project
# the testfile is here, OK
```

urso: Oh, I see. If `source == destination` for mounted dirs, then the replacement will swap out the root path `/` with the containers merge path. Thanks to default mounting rules the `/project` will be empty (mounts are not passed through by default). This is why we need the path translation rules in the first place.

You are right, we should remove the if condition and always add the mount infos. I'm no maintainer, nor regular contributor to `cross`, but would be nice if you can create a PR removing this check.

romainreignier: Ok, thanks for having a look to it. I will create a PR.

PR #396, opened 2020-03-12
Fix CI.

reitermarkus:
Consolidates https://github.com/rust-embedded/cross/pull/395 and https://github.com/rust-embedded/cross/pull/394.


reitermarkus:
bors try

PR #398, opened 2020-03-17
armv5te-unknown-linux-gnueabi target

Darneas:


Darneas:
As far as I can see, the automated tests pass. How can I help to make this pull request happen?

README.md
```diff
@@ -205,6 +205,7 @@ terminate.
 | `arm-unknown-linux-gnueabihf`        | 2.27   | 7.3.0   | ✓   | 4.1.0 |   ✓    |
 | `arm-unknown-linux-musleabi`         | 1.1.20 | 6.3.0   |     | 4.1.0 |   ✓    |
 | `arm-unknown-linux-musleabihf`       | 1.1.20 | 6.3.0   |     | 4.1.0 |   ✓    |
+| `armv5te-unknown-linux-gnueabi`      | ???    | ???     | ??? | ???   |   ?    |
```
ryankurte, line 208: is there a reason these are currently unspecified?

ryankurte:
thanks for the reminder, sometimes we need a push to get things back on the stack. apart from the comment above, lgtm

README.md
Darneas: Is there a straightforward way to print these versions in the container? I have a local one and could add these infos.

PR #409, opened 2020-04-16
Allow to mount host path in docker-in-docker mode

romainreignier:
As discussed here: https://github.com/rust-embedded/cross/pull/387#discussion_r408257447
The check is not needed to allow to mount paths that are mounted in the first docker with the same path.


reitermarkus:
bors try

reitermarkus:
Thanks, @romainreignier!

PR #424, opened 2020-05-13
Prevent pkg-config from being removed in image

### LINKED ISSUE #422: pkg-config purged from arm-unknown-linux-gnueabihf
`pkg-config` is getting removed in the `apt-get purge`, even though it was explicitly installed in [`common.sh`](https://github.com/rust-embedded/cross/blob/3e60c8cd5bfe40a1baefcdca1926fc0c94550ba0/docker/common.sh#L28):

https://dev.azure.com/rust-embedded/cross/_build/results?buildId=576&view=logs&j=ed13e0de-1834-5225-9caa-5c6f5a5a1fa1&t=91dd8e19-28bc-5135-e5ed-520472032e91&l=2665

But on `armv7-unknown-linux-gnueabihf` for instance, this is not happening:

https://dev.azure.com/rust-embedded/cross/_build/results?buildId=576&view=logs&j=36c4a3c9-c17a-548c-f0e4-0d2902437e43&t=26e6497a-1b48-5743-221f-9d1b01950b2c&l=4078

And testing locally:

```shell
$ docker run rustembedded/cross:arm-unknown-linux-gnueabihf which pkg-config || echo "not found"
not found

$ docker run rustembedded/cross:armv7-unknown-linux-gnueabihf which pkg-config
/usr/bin/pkg-config
```

jmagnuson:
Currently, installing the toolchain for `arm-unknown-linux-gnueabihf`
uses `xz-utils` which gets pruned. This also removes `pkg-config`, so
remove the prune altogether to keep both packages. `curl` is also kept
as that was originally installed in `common.sh` as well.

Fixes #422

### DIFF
--- a/docker/Dockerfile.arm-unknown-linux-gnueabihf
+++ b/docker/Dockerfile.arm-unknown-linux-gnueabihf
@@ -10,11 +10,9 @@ COPY xargo.sh /
 RUN /xargo.sh
 
 RUN mkdir /usr/arm-linux-gnueabihf && \
-    apt-get install --assume-yes --no-install-recommends curl xz-utils && \
     cd /usr/arm-linux-gnueabihf && \
     curl -L https://developer.arm.com/-/media/Files/downloads/gnu-a/8.3-2019.03/binrel/gcc-arm-8.3-2019.03-x86_64-arm-linux-gnueabihf.tar.xz | \
-    tar --strip-components 1 -xJ && \
-    apt-get purge --auto-remove -y curl xz-utils
+    tar --strip-components 1 -xJ
 
 ENV PATH /usr/arm-linux-gnueabihf/bin:$PATH
 


docker/Dockerfile.arm-unknown-linux-gnueabihf
```diff
@@ -13,8 +13,8 @@ RUN mkdir /usr/arm-linux-gnueabihf && \
     apt-get install --assume-yes --no-install-recommends curl xz-utils && \
     cd /usr/arm-linux-gnueabihf && \
     curl -L https://developer.arm.com/-/media/Files/downloads/gnu-a/8.3-2019.03/binrel/gcc-arm-8.3-2019.03-x86_64-arm-linux-gnueabihf.tar.xz | \
-    tar --strip-components 1 -xJ && \
-    apt-get purge --auto-remove -y curl xz-utils
+    tar --strip-components 1 -xJ
+    # packages not removed since they were originally installed in common.sh
```
reitermarkus, line 17: Maybe just remove the `apt-get install` line completely.

jmagnuson: Done in amended commit 843ad77bdb7b49cca78bd07da94ff053ed76fdf8.

reitermarkus:
Thanks, @jmagnuson!

PR #443, opened 2020-07-02
Allow specifying volumes in `Cross.toml`.

reitermarkus:


src/docker.rs
```diff
@@ -152,10 +170,10 @@ pub fn run(target: &Target,
         .args(&["-v", &format!("{}:/cargo:Z", cargo_dir.display())])
         // Prevent `bin` from being mounted inside the Docker container.
         .args(&["-v", "/cargo/bin"])
-        .args(&["-v", &format!("{}:/project:Z", mount_root.display())])
+        .args(&["-v", &format!("{}:/{}:Z", mount_root.display(), mount_root.display())])
```
jonas-schievink, line 173: This change will make it impossible to use a rustc wrapper with Cargo, which is the only way to pass custom flags due to https://github.com/rust-embedded/cross/issues/463, and the wrapper path needs to be absolute.

Setting it to `/project/my-wrapper.sh` works just fine, but if the mount point is changed, there is no easy way to pass the absolute path to the wrapper.

jonas-schievink: Hmm, actually I already have to do something different for cross builds vs. native builds, so maybe this won't be that big of a deal? It's certainly a breaking change though.

PR #445, opened 2020-07-16
Add cross subcommands shortcuts

sigmaSd:
Hi, this pr adds sub-commands shortcuts `['b','c','r', 't']` to the cli.
The main reason is currently the error when using them is misleading.
Example:
this command:
`cross b --target arm-linux-androideabi`
outputs this error: 
`error: linking with `cc` failed: exit code: 1` 
which took me a while to figure out that cross is not recognizing the sub-command and is falling-back to cargo.
Also they are useful to have.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -38,12 +38,12 @@ impl Subcommand {
 impl<'a> From<&'a str> for Subcommand {
     fn from(s: &str) -> Subcommand {
         match s {
-            "build" => Subcommand::Build,
-            "check" => Subcommand::Check,
+            "b" | "build" => Subcommand::Build,
+            "c" | "check" => Subcommand::Check,
             "doc" => Subcommand::Doc,
-            "run" => Subcommand::Run,
+            "r" | "run" => Subcommand::Run,
             "rustc" => Subcommand::Rustc,
-            "test" => Subcommand::Test,
+            "t" | "test" => Subcommand::Test,
             "bench" => Subcommand::Bench,
             "deb" => Subcommand::Deb,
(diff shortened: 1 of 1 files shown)


Emilgardis:
Thank you for your contribution!

This seems like it should be the default behaviour, but should cross maybe support all aliases, including userdefined aliases?

In cargo, the [b,c,r,t](https://github.com/rust-lang/cargo/blob/b92636f24f5297c935f9d040c97b03d1b4e0e72b/src/bin/cargo/main.rs#L63) aliases are shadowed by user-defined aliases.

sigmaSd:
I agree it makes sense. The issue here is to handle aliases 100% there are a lot of things to account for https://doc.rust-lang.org/cargo/reference/config.html, 
1- handling config file location
2- handling aliases environment variable
3- parsing the file
And all of these have a lot of edge cases. As far as I see so far cross tried to avoid this complexity and delegate most of this to cargo 
The good thing is -from the docs- `Aliases are not allowed to redefine existing built-in commands.` which mean that misleading errors when building or running should not happen

Emilgardis:
bors retry

domenicquirl:
Hi! I just spent a good while tearing my hair out over this, because I was trying to run `cross b` on a system where the corresponding `--target` was also installed and the command cause a regular build to run and complete without error (context: cross-compiling from a linux system to `armv7`, having installed the toolchain and a matching linker on the source machine to test the build locally). Got it figured out, but landed here in the process and I think this landing would help avoid such situations for others the future.

reitermarkus (approved):

PR #446, opened 2020-07-25
`docker-compose`-style `context` and `dockerfile` config

alsuren:
I was following along with https://www.collabora.com/news-and-blog/blog/2020/06/23/cross-building-rust-gstreamer-plugins-for-the-raspberry-pi/ and found the custom docker image workflow quite awkward (see also https://github.com/rust-embedded/cross/issues/149 )

In this PR, I add support for `context` and `dockerfile` options, based on the `build` config options that you can supply when using `docker-compose` (see https://docs.docker.com/compose/compose-file/#build for details) 

I hacked something up at https://github.com/lcsfelix/reading-xiaomi-temp/pull/1/files which uses this patch to cross-compile to a raspberry pi.

I would be very interested in getting feedback from @gdesmott on this because I have never used podman, and don't know whether the same tricks can be used.

# Outstanding Questions:

- [ ] kill off `context` option?

It might be possible to get away without supporting the `context` option, if you construct a dockerfile like this:

```
FROM japaric/aarch64-unknown-linux-gnu:v0.1.4 as dirtyhack

FROM debian:buster

COPY --from=dirtyhack \
    /common.sh /cmake.sh /xargo.sh /qemu.sh /dropbear.sh /linux-image.sh /linux-runner \
    /
RUN /common.sh && \
    && /cmake.sh \
    && /xargo.sh \
    && apt-get install -y --no-install-recommends \
        g++-arm-linux-gnueabihf \
        libc6-dev-armhf-cross \
    && /qemu.sh arm softmmu \
    && /dropbear.sh \
    && /linux-image.sh armv7


RUN dpkg --add-architecture armhf && \
    apt-get update && \
    apt-get install -y libgstreamer1.0-dev:armhf libgstreamer-plugins-base1.0-dev:armhf libssl-dev:armhf

ENV CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER=arm-linux-gnueabihf-gcc \
    CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_RUNNER="/linux-runner armv7" \
    CC_armv7_unknown_linux_gnueabihf=arm-linux-gnueabihf-gcc \
    CXX_armv7_unknown_linux_gnueabihf=arm-linux-gnueabihf-g++ \
    QEMU_LD_PREFIX=/usr/arm-linux-gnueabihf \
    RUST_TEST_THREADS=1 \
    PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabihf/pkgconfig
```
(disclaimer: I haven't tried this yet. I will do so when I'm back in wifi range)

This involves downloading and throwing away an entire cross-comilation docker image in order to copy the build context files out. We could provide a docker image that just contains the things from the build context, to help people make their custom dockerfiles in a slightly more efficient manner. Is this something that we want to do?

- [ ] support other docker-compose `build` options?

`docker-compose` specifies a bunch of other options in the `build` section. Should we think about supporting those as well, or is that too much scope creep?

- [ ] toml namespaces?

should we add `[target.armv7-unknown-linux-gnueabihf.docker.build]` section to our toml, or is it okay to keep everything flat under the `[target.armv7-unknown-linux-gnueabihf]` namespace?

- [ ] code structure, verbosity, and running `docker build` multiple times?

In order to get the sha id of the docker image, I run `docker build --quiet` and strip off the trailing newline. Anyone who is used to building docker images will be expecting to see a bunch of output when `docker build` is running. By specifying --quiet, we suppress this output, so people will just observe the `cross` command hanging. If they specify --verbose then they will see the output they're expecting.

I'm actually not that fussed about calling `docker build` lots of times, because it's pretty cheap when everything is cached (I think it just checksums all of the files that are COPYed from the build context and then noops). If anyone else has a better way to do this, please shout.

Also, even if you are happy with how I'm interacting with docker, please suggest refactors to my code. I feel like there are a bunch of layering violations, but my brain is not working well enough to unpick them at the moment.


src/docker.rs
```diff
@@ -183,15 +185,36 @@ pub fn run(target: &Target,
...
+        docker.args(&["-f", dockerfile]);
+    }
+    if verbose {
+        // Running `docker build` twice in quick succession is cheap
+        // (~0.1s extra) because of docker's layer cache.
+        // Here, we run it once without --quiet, to get the verbose output,
+        // then run it again with --quiet to get the image sha.
+        docker.run(verbose)?
+    }
+    let mut image_id = docker.arg("--quiet").run_and_get_stdout(verbose)?;
```
rvolgers, line 207: You should add the `--iidfile` option to the first `docker build` to write the image id to a temp file, instead of running it again and trying to parse the human-readable output.

alsuren: That seems to work. Thanks. Pushed.

src/docker.rs
```diff
@@ -183,15 +185,42 @@ pub fn run(target: &Target,
...
+        docker.args(&["-f", dockerfile]);
+    }
+    if verbose {
+        docker.run(verbose)?;
+    } else {
+        docker.run_and_get_stdout(verbose)?;
+    }
+    let mut image_id = std::fs::read_to_string(&image_id_filename)?;
+    image_id.retain(|c| c != '\n');
+    std::fs::remove_file(&image_id_filename).unwrap_or_else(
```
alsuren, line 211: I kind-of want to do this within a ScopeGuard, or import and use https://docs.rs/tempfile/3.1.0/tempfile/struct.NamedTempFile.html . What is `cross`'s policy on adding new dependencies?

docker/Dockerfile.armv7-unknown-linux-gnueabihf
```diff
@@ -1,28 +1,31 @@
+# This was built from ./Dockerfile.context
+FROM alsuren/cross-context:latest as context
+
 FROM ubuntu:16.04
```
alsuren, line 4: This idea turns out to be really powerful when creating custom images, because it allows you to provide `context` as an empty directory and `dockerfile` as something that uses this pattern to fetch build scripts etc.

In https://github.com/lcsfelix/reading-xiaomi-temp/pull/1/files I was previously using `context = "../cross/docker"` which is probably not something you want to check into your git repo, because it requires all developers to keep a copy of the `cross` repo in a specific location. 

TODO: 
- [ ] upload this `cross-context` docker image somewhere more official-sounding, and make sure that it gets uploaded and tagged with each release.
- [ ] revert the changes to this file, and produce a dedicated "example custom build image" dockerfile that uses this pattern.

alsuren:
@rvolgers I would be interested in getting experience reports. If you have time, could you install from my branch and tell me how you get on.

    cargo install --git=https://github.com/alsuren/cross --branch=docker-build-context

docker/Dockerfile.context
```diff
@@ -0,0 +1,49 @@
+# This dockerfile creates a thin docker image that can be used for making custom
+# cross Dockerfiles that live in other git repositories.
+# It has been uploaded to `alsuren/cross-context` so it can be used like so:
+
+## Fetch this image and use it as `context`:
+# FROM alsuren/cross-context:latest as context
```
alsuren, line 6: - [ ] What's the process for getting this uploaded somewhere more official?

Emilgardis: Our CI has access to dockerhub, see

https://github.com/rust-embedded/cross/blob/2af06e2a871f4cf467eb241a40d6ebef0cc43a8f/azure-pipelines.yml#L123

For example

alsuren: Hrm. The more I think about it, maybe this Dockerfile.context thing should be punted to another PR.

- It isn't strictly necessary for my use-case (cross-compiling something that links against dbus). You can get away with something like:
```
FROM rustembedded/cross:armv7-unknown-linux-gnueabihf-0.2.1 as context

RUN dpkg --add-architecture armhf && \
    apt-get update && \
    apt-get install -y libssl-dev:armhf libdbus-1-dev:armhf

ENV PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabihf/pkgconfig
```
- adding Dockerfile.context means that all of your build-scripts are suddenly exposed like public API, and are easy to break accidentally.
- Testing the context image is fiddly and intrusive, and requires more thought.

alsuren:
Okay. I have backed out all of the questionable Dockerfile.context bits. I think I'm happy with this.

I've also been using it for a while. The thing that's most questionable is the fact that we fall back to building on the host if we fail to build/find a docker image. This means that if you mess up the Dockerfile syntax and don't pass --verbose then you get output like this (and it's easy to miss the first line and wonder why everything is broken further down).

```
Warning: `"/usr/local/bin/docker" "build" "./docker" "--iidfile" "/var/folders/kp/k8982sg57h7cq12ylw459f180000gn/T/50928.iid" "-f" "./docker/Dockerfile.dbus.armv7-unknown-linux-gnueabihf"` failed with exit code: Some(1) Falling back to `cargo` on the host.
   Compiling libc v0.2.81
   Compiling syn v1.0.54
   Compiling cfg-if v0.1.10
   Compiling log v0.4.11
...
```

 I think that this should be fixed by adding a flag to forbid falling back to building on the host. Do we want this in this PR or a follow-up?

Emilgardis (changes requested):

src/docker.rs
```diff
@@ -193,15 +195,67 @@ pub fn run(target: &Target,
...
+        |e| println!(
```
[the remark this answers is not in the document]
alsuren: Good point. Fixed.

I also pushed da74ad3, because I realised that I had introduced some dead code as part of an earlier change.

Emilgardis (approved):
Looks good, going to wait with merge myself as I want to try it out myself, if any other code owner reviews feel free to merge

Issue #35, opened 2017-01-04
Recommended way of adding additional libraries to link against


shaunmulligan:
@japaric thanks for this awesome project!! 
I was just wondering how some one should go about adding additional libraries that need to be linked against. Some projects like https://github.com/plietar/librespot require additional dependencies such as `portaudio19-dev` and maybe some others. I was just wondering what is the appropriate method for adding these into cross so that the project builds correctly.

japaric:
Simplest way: build the library as a *static* library in a build script and link to that.

If you want to link to a shared library and you are using the `pkg-config` crate, then you basically *can't* cross compile because that crate doesn't support cross compilation scenarios (AFAIK).

If you want to link to a cross compiled shared library provided by ubuntu, e.g. `libfoo-dev:i386`, or against a shared library that you intent to build manually, you'll have to override the docker and the only way to do that is to create a *local* image named, for example, `japaric/i686-unknown-linux-gnu:v0.1.3` because that's what `cross` expects it to be called.

I suggest you base that local image on the ones `cross` uses by default. So you could use something like this for the Dockerfile:

``` dockerfile
FROM japaric/i686-unknown-linux-gnu:v0.1.3

# Or build it yourself
RUN dpkg --add-architecture i386 && \
    apt-get update && \
    apt-get install --no-install-recommends -y libfoo-dev:i386
```

iav:
This method, and code from main page, not work.
Because `Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty/universe/binary-armhf/Packages`, same for 
```
FROM rustembedded/cross:aarch64-unknown-linux-gnu-0.1.16

RUN dpkg --add-architecture arm64 && \
    apt-get update && \
    apt-get install libfoo:arm64
```
` Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty/main/binary-arm64/Packages  404  Not Found`
Because `trusty` have only amd64 and i386 architectures in repos.

gdesmott:
I'd suggest to close this as a dup of #149 as that's basically the same issue and this ticket contains extra information about how to make this work.

rvolgers:
If I understand correctly, this is precisely what was fixed in https://github.com/rust-embedded/cross/commit/3b07a08aa8739a9f22450ceecc56531d108661d5

Alexhuszagh:
We've documented this now extensively on the wiki, including the following examples:
[...]

PR #454, opened 2020-08-20
Bump README custom-container cross version to incorporate fixes

theunkn0wn1:
The custom dockerfile example in the README targets cross version 0.1.16, [and is known to not actually work as-is](https://github.com/rust-embedded/cross/commit/3b07a08aa8739a9f22450ceecc56531d108661d5).
The fix for this particular issue was released with cross 0.2, thus this PR bumps the cross version.

This became an issue in my use case as I wanted to cross-compile to a raspberry Pi, and needed to install some external libraries not included in the base images, and ran into issues such as #149 and #35 while attempting to follow the recommendation in the readme.

Bumping the cross version to 0.2.x resolved the above issues for me, and this PR is to help other users that may find themselves running into the same issues when following the readme closely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ This way you won't have to figure out how to install a cross C toolchain in your
 custom image. Example below:
 
 ``` Dockerfile
-FROM rustembedded/cross:aarch64-unknown-linux-gnu-0.1.16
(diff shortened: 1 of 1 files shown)


therealprof (approved):
LGTM, thanks.

bors r+

PR #460, opened 2020-09-16
Add --docker-image argument overriding Cross.toml

willglynn:
I have a situation where one project needs to be built in multiple environments for the same target triple. I can easily model the environments as separate Docker images, but `Cross.toml` only lets me specify one `image` per target. If I want to build each version automatically, I would have to automatically edit `Cross.toml` to do so.

This PR adds a `--docker-image` argument, allowing me to write a script like:

```sh
cross build --target=foo --docker-image=my/image:first-environment --release
package -o first_environment.pkg target/foo/app

cross build --target=foo --docker-image=my/image:second-environment --release
package -o second_environment.pkg target/foo/app
```

This PR is longer than strictly necessary since I accepted `rustfmt`'s output for any line I touched. I can provide a minimal patch instead if you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ the default one. Normal Docker behavior applies, so:
 
 - If only `tag` is omitted, then Docker will use the `latest` tag.
 
+You may also pass `--docker-image=my/image:tag` specify a Docker image on a
+per-command basis.
+
 It's recommended to base your custom image on the default Docker image that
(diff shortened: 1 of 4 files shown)


reitermarkus:
I think this should use an environment variable instead, similar to how `cargo` uses `CARGO_TARGET_<triple>_RUNNER`, i.e. `CROSS_TARGET_<triple>_DOCKER_IMAGE`.

PR #462, opened 2020-09-20
add cross metadata

Emilgardis:
this will enable running `cargo metadata` in the container instead of on the host.


reitermarkus (approved):

reitermarkus:
Thanks!

PR #471, opened 2020-10-18
Fix i686-pc-windows-gnu

malbarbo:
The build command in mingw.sh script was failing but the script continued to run and returned 0. The result was that the mingw toolchain with sjlj was being used in the image, not the one with dwarf expectation we are trying to build!

I set the option -e and fixed the script.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,9 +22,9 @@ jobs:
         aarch64-unknown-linux-gnu:       { TARGET: aarch64-unknown-linux-gnu,       CPP: 1, DYLIB: 1, STD: 1, RUN: 1, RUNNERS: qemu-user qemu-system }
         arm-unknown-linux-gnueabi:       { TARGET: arm-unknown-linux-gnueabi,       CPP: 1, DYLIB: 1, STD: 1, RUN: 1 }
         arm-unknown-linux-gnueabihf:     { TARGET: arm-unknown-linux-gnueabihf,     CPP: 1, DYLIB: 1, STD: 1, RUN: 1 }
-        armv7-unknown-linux-gnueabihf:   { TARGET: armv7-unknown-linux-gnueabihf,   CPP: 1, DYLIB: 1, STD: 1, RUN: 1, RUNNERS: qemu-user qemu-system                                                                     }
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -22,9 +22,9 @@ jobs:
         aarch64-unknown-linux-gnu:       { TARGET: aarch64-unknown-linux-gnu,       CPP: 1, DYLIB: 1, STD: 1, RUN: 1, RUNNERS: qemu-user qemu-system }
         arm-unknown-linux-gnueabi:       { TARGET: arm-unknown-linux-gnueabi,       CPP: 1, DYLIB: 1, STD: 1, RUN: 1 }
         arm-unknown-linux-gnueabihf:     { TARGET: arm-unknown-linux-gnueabihf,     CPP: 1, DYLIB: 1, STD: 1, RUN: 1 }
-        armv7-unknown-linux-gnueabihf:   { TARGET: armv7-unknown-linux-gnueabihf,   CPP: 1, DYLIB: 1, STD: 1, RUN: 1, RUNNERS: qemu-user qemu-system                                                                     }
```
malbarbo, line 25: Some cosmetic changes

docker/mingw.sh
```diff
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 
-set -x
+set -xe
```
reitermarkus, line 3: Make this `set -euo pipefail` like in all other scripts.

reitermarkus (changes requested):

docker/mingw.sh
malbarbo: Fixed. Thanks.

malbarbo:
Rebased to fix the solaris build.

docker/mingw.sh
```diff
@@ -95,7 +97,7 @@ EOF
 -targets := i686-w64-mingw32 x86_64-w64-mingw32
 +targets := i686-w64-mingw32
 -threads := posix win32
-+threads := posix
++threads := posix win32
```
reitermarkus, line 100: Seems like this patch line isn't needed anymore since `-` and `+` are the same?

malbarbo: My bad. Fixed now.

reitermarkus (approved):

reitermarkus:
Thanks!

PR #472, opened 2020-10-19
Add a way to choose Podman instead of Docker

oblique:
By default, `cross` tries to use Docker or Podman, with the former to have priority. In you want to use other container engine or explicitly choose one you can set the name of binary (or the path) in `CROSS_CONTAINER_ENGINE` environment variable.

For example in case you want use Podman, you can set `CROSS_CONTAINER_ENGINE=podman`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,15 @@ overlayfs can not be unmounted correctly by Docker if the child container still
 accesses it.
 
 
+### Explicitly choose the container engine
+
+By default, `cross` tries to use [Docker] or [Podman], in that order.
+If you want to choose a container engine explicitly, you can set the
+binary name (or path) using the `CROSS_CONTAINER_ENGINE`
+environment variable.
+
+For example in case you want use [Podman], you can set `CROSS_CONTAINER_ENGINE=podman`.
+
 ### Passing environment variables into the build environment
 
 By default, `cross` does not pass any environment variables into the build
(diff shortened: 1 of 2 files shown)


src/docker.rs
```diff
@@ -17,7 +18,17 @@ const DOCKER: &str = "docker";
 const PODMAN: &str = "podman";
 
 fn get_container_engine() -> Result<std::path::PathBuf> {
-    which::which(DOCKER).or_else(|_| which::which(PODMAN)).map_err(|e| e.into())
+    let use_podman = env::var("CROSS_USE_PODMAN")
```
reitermarkus, line 21: Maybe the environment variable should rather be called `CROSS_CONTAINER_ENGINE` and contain the (path to the) command, i.e. `CROSS_CONTAINER_ENGINE=podman`?

oblique: Good point. I'm changing it

oblique: ready

README.md
```diff
@@ -166,6 +166,15 @@ overlayfs can not be unmounted correctly by Docker if the child container still
 accesses it.
 
 
+### Explicitly choose the container engine
+
+By default, `cross` tries to use [Docker] or [Podman], with the former to have
+priority. In you want to use other container engine or explicitly choose one, you
+can set the name of binary (or the path) in `CROSS_CONTAINER_ENGINE` environment
+variable.
```
reitermarkus, line 174:
```suggestion
By default, `cross` tries to use [Docker] or [Podman], in that order.
If you want to choose a container engine explicitly, you can set the
binary name (or path) using the `CROSS_CONTAINER_ENGINE`
environment variable.
```

reitermarkus (approved):

reitermarkus:
Thanks!

PR #481, opened 2020-11-05
Update the ubuntu version for x86_64-unknown-linux-gnu

mathiaspeters-sig:


docker/linux-image.sh
```diff
@@ -83,7 +83,12 @@ main() {
             ;;
         x86_64)
             arch=amd64
-            kernel="${kversion}-amd64"
+            #kernel="${kversion}-amd64"
```
mathiaspeters-sig, line 86: While building an image with ubuntu 20.04 it fails to find some packages. Adding this allows it to find them, and the kernel bump was necessary for the image to build locally. Also, as far as I can see there is only one docker image using amd64 that is using this script so these changes will have no effect on other images

Alexhuszagh:
I'm closing this in favor of #591, for a few reasons:
1. This image, and many more, are all handled by #591, which will be merged in 0.3.0.
2. This uses an outdated kernel version.
3. It uses the old `x86_64-unknown-linux-gnu` image which is CentOS-based and not Ubuntu-based.
4. It doesn't have the `libcrypt1` requirement for building almost all newer Linux images.
5. It does not have some painfully-discovered workarounds required for Ubuntu 20.04 support.

For example, some of the versions in the Debian repositories (not unexpectedly) are older than the versions installed in Ubuntu, preventing us from downloading them and causing linking issues later on in full-system emulation, a patch which is available [here](https://github.com/cross-rs/cross/blob/6615f31bcf1722b4071e77f8587300efdf25f05c/docker/linux-image.sh#L237-L266).

PR #483, opened 2020-11-08
Update armv7-unknown-linux-gnueabihf base image

w4tson:
I was attempting to cross compile gtk app (using druid) and needed to create custom cross image with the relevant libs installed. The version of the libs needed required required a higher version of ubuntu, hence this PR. Thanks

### DIFF
--- a/docker/Dockerfile.armv7-unknown-linux-gnueabihf
+++ b/docker/Dockerfile.armv7-unknown-linux-gnueabihf
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 COPY common.sh /
 RUN /common.sh


ryankurte:
hey thanks for the PR! for `glibc` compatibility related reasons [1](https://github.com/rust-embedded/cross/pull/164#issuecomment-352188362) [2](https://github.com/rust-embedded/cross/issues/85) the default images usually have to remain _old_ :-/ you may find it useful to specify a [custom image](https://github.com/rust-embedded/cross#custom-docker-images) with the updated base and any other libraries you need.

w4tson:
Ah thanks, I figured there might be a catch. This is an ace project. The custom route works well enough

PR #438, opened 2020-06-29
Support sub project build without parent Cargo.toml.

avkviring:

Issue #388, opened 2020-03-10
Building with local dependencies


Piripant:
It seems that when building with a local dependency cross fails to load the source for that local dependency. If this is my Cargo.toml:
```toml
[dependencies]
nasatopo = { path = "../nasatopo" }
```
The output of `cross build` is:
```
error: failed to load source for a dependency on `nasatopo`

Caused by:
  Unable to update /nasatopo

Caused by:
  failed to read `/nasatopo/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```
Which is technically correct as the is not /nasatopo, but it the Cargo.toml I specified a local path, not a global one, as cross seems to understand.
Furthermore if I try to replace the relative path with a global one, the same error still happens, this time pointing to a path which is actually valid and takes to the location of the nasatopo source folder.

I think this has to do with how a docker container interfaces with the file-system, but I am not familiar and cannot tell.

velvia:
This is hitting me as well.   I even tried symlink but it doesn't work either.

zekronium:
Cross only mounts the cargo project.

I had this issue as well and resolved it by moving the dependency into the project dir.

The path then has to be changed to
`nasatopo = { path = "nasatopo" }`

reitermarkus:
Yeah, we would need to parse the `Cargo.toml` and find all `path` dependencies, then mount them inside the container.

kornelski:
Could you mount root of the workspace?

I'm trying to build one of crates in a workspace, but running into the same problem - that crate gets moved out of its workspace.

bplower:
I'm bumping into this as well, but I was hoping to solve it using the volume mount feature. I tried adding a `Cross.toml` file with the following content:
```toml
[build.env]
volumes = [
    "MY_DEP",
]
```

In my Cargo.toml I have:

```toml
[dependencies]
my_dep = { path = "../my-dep" }
```
However when I run cross with that environment variable set, it fails with the same error reported by @Piripant. Verbose output shows that the volume mount isn't being added:

```console
$ MY_DEP=../my-dep cross build --target x86_64-unknown-linux-gnu --verbose
+ "rustc" "--print" "sysroot"
+ "rustup" "toolchain" "list"
+ "rustup" "target" "list" "--toolchain" "stable-x86_64-unknown-linux-gnu"
+ "rustup" "component" "list" "--toolchain" "stable-x86_64-unknown-linux-gnu"
[...]

Issue #277, opened 2019-07-02
Supporting relative path in Cargo.toml depends


TotalKrill:
How would one do this, currently I am building the entire workspace, which makes it work, but then other tools doesn't work since they do not support workspaces it seems

Alexhuszagh:
A simple way to reproduce this:

```toml
submod = { path = "../submod, version = "1.0" }
```

Relative paths work fine if they're a child of the directory, any ".." in the path causes issues.

Dylan-DPC:
Thanks for opening the issue (and sorry about the delays) 

From the readme: 
> path dependencies (in Cargo.toml) that point outside the Cargo project won't work because cross use docker containers only mounts the Cargo project so the container doesn't have access to the rest of the filesystem.

So right now it wouldn't be possible to make this change without making a major change in how `cross` works which is beyond our scope at the moment.

Alexhuszagh:
This was fully fixed in #684, #713, and #937.

Alexhuszagh:
This was fully fixed in #684.

Issue #139, opened 2017-10-17
Dependencies referenced by path.


jD91mZM2:
Example:
```
crate = { path = "../crate" }
```

This fails with
```
error: failed to load source for a dependency on `crate`

Caused by:
  Unable to update file:///crate

Caused by:
  failed to read `/crate/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

japaric:
This is expected and documented in the [README](https://github.com/japaric/cross#caveats--gotchas).

jD91mZM2:
Sorry, my bad

gnzlbg:
If I have the following structure:

* `root_crate/`
  * `src/`
  * `examples/`
    * `example_crate/`
        * `src/`

where `example_crate` uses `path = ../../` to refer to the root crate, and one executes from the `root_crate`: `cross test -p examples/example_crate` this should work though as long as `example_crate` is part of the `root_crate`'s workspace. 

However, if one excludes `example_crate` from the `root_crate`'s workspace, then doing `cross test --manifest-path examples/example_crate` will fail, because that will try to write a `Cargo.lock` into the `examples/example_crate` directory, which is read only.

Alexhuszagh:
This was fully implemented in #684.

PR #494, opened 2020-12-04
feat: Parse Cargo's `--manifest-path` option to determine mounted docker root

wngr:
This commits adds support for parsing the `--manifest-path` option to cross. So
far, the `Cargo.toml` manifest of the crate (or its Cargo workspace) to compile
has been assumed to be in the current working directory. This means, that
relative crate dependencies were not supported, because those paths were not
mapped into the docker container.

Take the following example structure, where `my-bin` depends on `my-lib`:
```
.
├── my-bin
│   ├── Cargo.toml
│   └── src
│       └── main.rs
└── my-lib
    ├── Cargo.toml
    └── src
        └── lib.rs
```

This commits enables such scenarios, by running cross from `.` like so:
`cross build --manifest-path=my-lib/Cargo.toml --target x86_64-pc-windows-gnu`,
as `.` is mapped as the container's root, and the options passed through to
Cargo.

Related #388 #139 #277 #78 #438

Co-authored-by: Kviring Alexey <alex@kviring.com>

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -64,26 +64,35 @@ impl Root {
 }
 
 /// Cargo project root
-pub fn root() -> Result<Option<Root>> {
-    let cd = env::current_dir().wrap_err("couldn't get current directory")?;
-
-    let mut dir = &*cd;
-    loop {
-        let toml = dir.join("Cargo.toml");
-
-        if fs::metadata(&toml).is_ok() {
-            return Ok(Some(Root {
-                path: dir.to_owned(),
-            }));
+pub fn root(manifest_path: Option<PathBuf>) -> Result<Option<Root>> {
+    if let Some(manifest_path) = manifest_path {
+        if !manifest_path.is_file() {
+            eyre::bail!("No manifest found at \"{}\"", manifest_path.display());
         }
+        return Ok(Some(Root {
+            path: manifest_path
+                .parent()
+                .expect("File must have a parent")
+                .to_owned(),
+        }));
+    } else {
+        let cd = env::current_dir().wrap_err("couldn't get current dir")?;
+        let mut dir = &*cd;
+        loop {
+            let toml = dir.join("Cargo.toml");
 
-        match dir.parent() {
-            Some(p) => dir = p,
-            None => break,
+            if fs::metadata(&toml).is_ok() {
+                return Ok(Some(Root {
+                    path: dir.to_owned(),
+                }));
+            }
+            if let Some(p) = dir.parent() {
+                dir = p;
+            } else {
+                break Ok(None);
+            }
         }
     }
-
-    Ok(None)
 }
 
 /// Pass-through mode
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -13,11 +13,13 @@ pub struct Args {
     pub target: Option<Target>,
     pub target_dir: Option<PathBuf>,
     pub docker_in_docker: bool,
+    pub manifest_path: Option<PathBuf>,
 }
 
 pub fn parse(target_list: &TargetList) -> Args {
     let mut channel = None;
     let mut target = None;
+    let mut manifest_path: Option<PathBuf> = None;
     let mut target_dir = None;
     let mut sc = None;
     let mut all: Vec<String> = Vec::new();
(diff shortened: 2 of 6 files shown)


wngr:
Would appreciate some feedback on the PR -- is something missing? Would you consider merging this feature upstream?

src/cli.rs
```diff
@@ -13,19 +13,34 @@ pub struct Args {
     pub target: Option<Target>,
     pub target_dir: Option<PathBuf>,
     pub docker_in_docker: bool,
+    pub project_dir: Option<PathBuf>,
 }
 
 pub fn parse(target_list: &TargetList) -> Args {
     let mut channel = None;
     let mut target = None;
+    let mut project_dir: Option<PathBuf> = None;
```
reitermarkus, line 22: Can you make this be the `manifest_path` instead? That way we could use the same logic whether `--manifest-path` is specified or not. I. e. automatically find the `Cargo.toml` when running `cross` in a subdirectory as well.

wngr: Not sure I understand. Do you mean just renaming the variable?
`project_dir` is not necessarily `manifest_path`. That's the point of this PR to decouple the mounted root path from the location of the `Cargo.toml` file. So I'd say it's better to have a distinctive name for it.

reitermarkus: > `project_dir` is not necessarily `manifest_path`

From looking at the code, `project_dir` is `current_dir`/`manifest_path`, i.e. the absolute path to the `Cargo.toml`, isn't it?

wngr: Right, I was confused.

src/cargo.rs
```diff
@@ -73,8 +73,11 @@ impl Root {
 }
 
 /// Cargo project root
-pub fn root() -> Result<Option<Root>> {
-    let cd = env::current_dir().chain_err(|| "couldn't get current directory")?;
+pub fn root(manifest_path: Option<PathBuf>) -> Result<Option<Root>> {
+    let cd = match manifest_path {
+        Some(dir) => dir,
```
reitermarkus, line 78: `manifest_path` is not a directory but is already a path to a `Cargo.toml`.

wngr: Thanks. I'd appreciate another look now.

UebelAndre:
This would be a great change to be able to get in! 😃

wcampbell0x2a:
Any updates on this? I would use this with `actions-rs/cargo@v1`, so a release of this would be awesome.

TotalKrill:
We are currently using this rebased upon master with no issues building a huge project with microservices and interdependecies. 

It has allowed us to migrate from copy pasting Cargo.toml workspace files depending on which software to build for which target to have the files include the correct libraries, to being able to build them much simpler. Please merge

src/cargo.rs
```diff
@@ -73,24 +74,35 @@ impl Root {
...
-    let cd = env::current_dir().chain_err(|| "couldn't get current directory")?;
-
-    let mut dir = &*cd;
-    loop {
-        let toml = dir.join("Cargo.toml");
-
-        if fs::metadata(&toml).is_ok() {
-            return Ok(Some(Root { path: dir.to_owned() }));
+pub fn root(manifest_path: Option<PathBuf>) -> Result<Option<Root>> {
+    if let Some(cargo_path) = manifest_path {
```
reitermarkus, line 78:
```suggestion
    if let Some(manifest_path) = manifest_path {
```

reitermarkus (changes requested):
Please add a changelog entry.

Emilgardis:
Please rebase instead of adding merge commits, if you need guidance how just ask :D

src/cargo.rs
```diff
@@ -64,26 +64,35 @@ impl Root {
...
-        let toml = dir.join("Cargo.toml");
-
-        if fs::metadata(&toml).is_ok() {
-            return Ok(Some(Root {
-                path: dir.to_owned(),
-            }));
+pub fn root(manifest_path: Option<PathBuf>) -> Result<Option<Root>> {
+    if let Some(manifest_path) = manifest_path {
+        if !manifest_path.is_file() {
+            eyre::bail!("No manifest found at \"{}\"", manifest_path.display());
```
Emilgardis, line 70: we should be consistent about how we present paths

we currently have three ways of doing it

`format!("{}", path)` ``format!("`{}`", path)`` and `format!(r#""{}""#, path)`

don't know which one is best and it's not a blocker for this pr

Emilgardis (approved):

Emilgardis:
I think this is the wrong approach now, however it can be fixed without explicit support if we parse cargo metadata for the target (and other) folders. See #684 

Am I wrong in this thinking?

otavio:
Cloud this be closed in favour of #684 ?

Emilgardis:
implemented/included in #684, thanks!

datdenkikniet:
Could this PR be reopened? The solution implemented in #684 does not solve the problem immediately described by/solved by this PR.

While the effect may be similar, it does not solve the problem if one is not using a workspace. The reason we can't/wish not to use a workspace is that compiling for multiple targets within one workspace can create conflicts with `libc` versions, as well as with other dependencies compiled against other system libraries.

Emilgardis:
Have you tried the changes? It implements this pr, 09e5d0eeaa15ddf05930ba6aab73ef589ef65a48 in #684 is this pr.

datdenkikniet:
Aho, apologies. Missed that bit :) Thanks for the response!

Emilgardis:
No worries!

I'm curious about the libc clashes you mentioned, is it something like https://github.com/cross-rs/cross/issues/724 maybe?

PR #464, opened 2020-10-07
aarch64-unknown-linux-musl: do not set RUSTFLAGS

jonas-schievink:
https://github.com/rust-lang/rust/issues/46651 has been fixed, and this helps with https://github.com/rust-embedded/cross/issues/463

PR #497, opened 2020-12-08
don't set RUSTFLAGS in aarch64-musl image

japaric:
instead use a linker wrapper to work around issue rust-lang/rust#46651
not setting RUSTFLAGS inside the container lets end users pass rustc flags via `.cargo/config`
this is an alternative to #464 that doesn't require bumping the MSRV of the `aarch64-unknown-linux-musl` target (that is the aarch64-musl image will continue to work with Rust 1.42 where bug rust-lang/rust#46651 is still presen )


japaric:
Note that the reported CI failures are unrelated to this PR. CI is broken on the main branch; PR #498 fixes that.

reitermarkus (approved):

PR #498, opened 2020-12-09
bump linux-image version to fix CI

japaric:
the Dockerfiles try to use linux-image-4.19.0-10 which no longer exists in the debian package repository; this is making CI fail: https://dev.azure.com/rust-embedded/cross/_build/results?buildId=772&view=logs&j=e6deedce-f53c-5dcc-524e-14349f140ffd&t=9e60ab89-47e0-5dc9-e657-a2fbb695cbf9&l=6521
this PR updates the package version (revision?) to linux-image-4.19.0-**11**

### DIFF
--- a/docker/android-system.sh
+++ b/docker/android-system.sh
@@ -54,43 +54,36 @@ EOF
     curl --retry 3 -sSfL https://storage.googleapis.com/git-repo-downloads/repo -O
     chmod +x repo
 
-    # the `repo` tool requires python3 so change the default python interpreter
-    ln -sf python3 /usr/bin/python
-
     # this is the minimum set of modules that are need to build bionic
     # this was created by trial and error
-    ./repo init -u https://android.googlesource.com/platform/manifest -b android-5.0.0_r1
-    ./repo sync -c bionic
-    ./repo sync -c build
-    ./repo sync -c external/compiler-rt
-    ./repo sync -c external/jemalloc
-    ./repo sync -c external/libcxx
-    ./repo sync -c external/libcxxabi
-    ./repo sync -c external/libselinux
-    ./repo sync -c external/mksh
-    ./repo sync -c external/openssl
-    ./repo sync -c external/stlport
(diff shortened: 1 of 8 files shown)


japaric:
> [rust-embedded.cross (aarch64-linux-android) Failing after 5m](https://dev.azure.com/rust-embedded/cross/_build/results?buildId=773&view=logs&j=aff0125d-cb57-5513-a71a-8603628a1277&t=45ca1fc7-55f1-5b44-8f11-091bc0bc54ce&l=2643)

the android builds are failing because the `repo` tool used in the `android-system.sh` script no longer supports python 2 which is what the Dockerfile is using. `repo` supports python 3.6+ and ubuntu 16.04 ships with python 3.5 so trying to use the `python3` package doesn't work either. I tried to use an older snapshot of the `repo` tool (2.7-2.10) but those still don't work with python2 because the `repo` tool appears to execute some third party python code that's not python2 compatible.

japaric:
ok, moved the android images to Ubuntu 20.04; that should fix the issues with the `repo` tool

basing those images on Ubuntu 20.04 (*) shouldn't raise the minimum supported android version because we are building an android SDK from scratch and android doesn't depend on glibc

(*) this is more of an issue with the linux-gnu targets which depend on glibc; there an old ubuntu image is used to keep the supported glibc version (of Rust programs produced by `cross`) as low as possible

docker/android-system.sh
```diff
@@ -53,6 +54,9 @@ EOF
     curl --retry 3 -sSfL https://storage.googleapis.com/git-repo-downloads/repo -O
     chmod +x repo
 
+    # the `repo` tool requires python3 so change the default python interpreter
```
reitermarkus, line 57: Can we call it with `python3 ./repo` instead?

japaric: that works and it's better! updated the PR

reitermarkus (approved):

PR #500, opened 2020-12-11
use runner setting specified in Cross.toml

### LINKED ISSUE #499: Cross.toml: runner setting is ignored
``` toml
# Cross.toml
[target.aarch64-unknown-linux-gnu]
runner = ["qemu-system"]
```

The above setting is ignored by `cross`.

As you can see in https://github.com/rust-embedded/cross/blob/8df1f8cd2fe6d09cf312fd4c2cd498499ebc423d/src/docker.rs#L106 and https://github.com/rust-embedded/cross/blob/8df1f8cd2fe6d09cf312fd4c2cd498499ebc423d/src/docker.rs#L176

The `CROSS_RUNNER` variable is always set to `""` in the docker container.

Apparently, the issue was (unintentionally) introduced in #344.

japaric:
fixes #499

### DIFF
--- a/docker/linux-image.sh
+++ b/docker/linux-image.sh
@@ -51,7 +51,7 @@ main() {
             echo "APT::Get::AllowUnauthenticated true;" | tee -a /etc/apt/apt.conf.d/10-nocheckvalid
 
             dropbear="dropbear"
-            deps=(libcrypt1:$arch)
+            deps=(libcrypt1:"${arch}")
             ;;
         powerpc64)
             # there is no stable port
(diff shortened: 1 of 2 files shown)


reitermarkus (dismissed):

japaric:
it seems that the debian sysroot that's used by the `qemu-system` mode is missing a shared library. This is the output of `cross test` with CROSS_DEBUG=1:
``` console
dropbear: error while loading shared libraries: libtomcrypt.so.1: cannot open shared object file: No such file or directory
[    2.593663] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
[    2.593663]
[    2.594109] CPU: 2 PID: 1 Comm: dropbear Not tainted 4.19.0-11-arm64 #1 Debian 4.19.146-1
[    2.594255] Hardware name: linux,dummy-virt (DT)
[    2.595546] Call trace:
```

japaric:
> it seems that the debian sysroot that's used by the qemu-system mode is missing a shared library.

last two commits should fix this
(ah, I forgot to run shellcheck locally)

reitermarkus (approved):

PR #506, opened 2021-01-05
Update emscripten version

josephlr:
Emscripten and rustc have to use compatible versions of LLVM to avoid
linker errors. With the update to LLVM 11, emscripten also requires an
update, see https://github.com/rust-lang/rust/pull/75716.

Also, now the emscripten is updated, the fix for
https://github.com/kripken/emscripten/issues/4542 is now available, which
means we can delete the node-wasm workaround.

Signed-off-by: Joe Richey <joerichey@google.com>


reitermarkus:
Closing since the version is updated. The images are currently disabled on CI, however.

PR #516, opened 2021-02-01
Add support for building ARM64 iOS crates on Linux

aspenluxxxy:
Adds support for `cross build --target aarch64-apple-ios` on Linux hosts.

Notes:

 * the [Apple LLVM](https://github.com/apple/llvm-project) is built in the dockerfile
 * there is no support for GCC / GNU utilities, everything is LLVM.
 * the [cctools-port](https://github.com/tpoechtrager/cctools-port) wrapper is used
 * some of the files downloaded should _really_ be mirrored - jsdelivr was the only CDN with support for Github releases, and it has a 50MB limit (quite a bit below the size of the SDK files)


Moxinilian:
Hello! I just opened #579 without seeing your PR existed. Any reason why it was closed?

Absolucy:
@Moxinilian because chances are it was never going to be merged anyways

Moxinilian:
Besides the obvious issue with the file delivery you mentioned, your PR seems to be good to me. Maybe if we show there is interest in this, we could have it move forward.

Moxinilian:
Additionally, the iPhoneOS SDK should be fetched directly from Apple, as I am not sure of the legality of the GitHub repository this depends upon.

I think it would be possible to extract it from an Xcode copy fetched from Apple using an user-provided Apple developer account. Obviously, as users of the feature must comply with Apple’s terms of service, this whole feature is only to be used on Linux installed on Apple-branded hardware.

PR #524, opened 2021-03-03
docker: Add Nix Store volume support

otavio:
When running inside NixOS or using Nix packaging we need to add the Nix
Store to the running container so it can load the needed binaries.

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #552 - Added CHANGELOG.md automation
 - #543 - Added environment variables to control the UID and GID in the container
 - #534 - fix image builds with update of dependencies
+- #524 - docker: Add Nix Store volume support
 - #502 - fix ci: bump openssl version in freebsd
 - #501 - x86_64-linux: lower glibc version requirement to 2.17 (compatible with centos 7)
 - #500 - use runner setting specified in Cross.toml
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -82,6 +82,7 @@ pub fn run(
     let xargo_dir = env::var_os("XARGO_HOME")
         .map(PathBuf::from)
         .unwrap_or_else(|| home_dir.join(".xargo"));
+    let nix_store_dir = env::var_os("NIX_STORE").map(PathBuf::from);
     let target_dir = target_dir.clone().unwrap_or_else(|| root.join("target"));
 
     // create the directories we are going to mount before we mount them,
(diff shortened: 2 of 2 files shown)


otavio:
The error on the pipeline does not seem related. Any idea what might have caused it?

otavio:
> The error on the pipeline does not seem related. Any idea what might have caused it?

This is indeed not related. See:
```
E: Unable to locate package linux-image-4.19.0-11-arm64:arm64
E: Couldn't find any package by glob 'linux-image-4.19.0-11-arm64'
E: Couldn't find any package by regex 'linux-image-4.19.0-11-arm64'
```

otavio:
Any news on this?

reitermarkus:
What is the use case for this? It seems like this should already be possible using `env.volumes`.

otavio:
It fails to run automatically; so we would need to pass it to every time we call cross when using NixOS.

reitermarkus:
> It fails to run automatically

What exactly fails to run? I'm not really familiar with NixOS, can you explain in more detail which packages need to be available/mounted in the container and why?

otavio:
I am unsure if you know what NixOS is. So please take a look at [How Nix works](https://nixos.org/guides/how-nix-works.html) document. It explains how the Nix store works and more.

In summary, without the patch, `cross` fails to run because _it cannot find the binaries it runs inside the container_. It took me a while to understand the root cause of the error but now it works just fine. This is the same patch I [apply](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/rust/cargo-cross/default.nix#L21-L26) in the official Nix package.

reitermarkus:
> It explains how the Nix store works and more.

Okay, I understand now what the Nix store does.

> it cannot find the binaries it runs inside the container

I'm still not sure what specifically the problem is though. Let me know if the following assumption is correct.

Let's say we are talking about the `cargo` binary:

- We mount the `rustup` toolchain into the container.
- The toolchain's `./bin/cargo` is not a binary but a symlink into the Nix store. 
- Calling `cargo` fails since the Nix store is not mounted.

otavio:
Exactly, your understanding is correct. Doing the mount inside the container allows it to run successfully.

reitermarkus:
Just one more question. Does Nix use some kind of customized version of `rustup` or how does symlinking the toolchain bineries into Nix store happen? I'd assume `rustup` just installs a complete pre-built toolchain from a tarball. Or is this a custom toolchain installed by Nix and not `rustup`?

otavio:
Yes, it does. See [rustup](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/rust/rustup/default.nix).

reitermarkus (approved):

otavio:
Should I do anything on my side on this one?

Emilgardis:
fix the fmt error and it should be good to go :)

https://github.com/cross-rs/cross/runs/5578287402?check_suite_focus=true#step:4:10

bors d+

Emilgardis:
yeah, thought so. Should maybe fix that in upstream bors

Emilgardis (approved):
bors r+

Issue #147, opened 2017-11-08
get configuration from the environment


japaric:
Today Cross can be configured via the Cross.toml file. However, we would also like to be able to specify this configuration via environment variables.

### What to do

- Each configuration entry allowed in Cross.toml should be overrided by a corresponding environment variable, if set. If the configuration entry is named `build.env.passthrough` then its environment variable counterpart will be named `CROSS_BUILD_ENV_PASSTHROUGH`. If the configuration entry takes a list of strings as value then the env variable value will be parsed as a list of *space* separated strings.

### References

- The Cross.toml parsing logic is in `src/main.rs`, in the `impl Toml` block.
- Cargo has a similar behavior. You'll probably be able to borrow some logic from there.

Required for japaric/trust#73

olitha:
Hi Japaric,

I would like to set images based on target devices that have the same architecture but should be built in different containers. This could be done with setting an env variable to select the build image.

So I am interested in this feature and I would like to give it a try to implement it.

I had a look in the cargo source and it looks for me as there the env vars are read when they are used. My idea instead is to use a `struct Config` that holds both the env vars and the `Toml`. It acts then as a facade and checks for vars first, then the toml.
Signatures of functions then change because I pass the Config.
The only limit I see with this, is that no env vars could be set during the build process that could change the build in later steps, like in a build.rs. But this is anyway not relevant for cross probably.

Do you see any downside with this?

olitha:
Right now for the xargo flag the `[build]]xargo` value overrides the values in target. So there is no way to specify a default.

I wonder if it would be more intuitive to use the most special value, meaning:
```
[build]
xargo = true
[target.aarch64-unknown-linux-gnu]
xargo = false
```
instead of currently using `xargo=true` always, it would use `xargo=true` for all targets that are not further specified, but `xargo=false` for target aarch64-unknown-linux-gnu

Alexhuszagh:
Closed in #525.

PR #525, opened 2021-03-05
Allow config settings with environment variables

olitha:
#147 

Config is the facade and gets first from env then from toml.
All references to toml change to config.

What are your thoughts?

---

Closes https://github.com/rust-embedded/cross/pull/460.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -85,6 +85,27 @@ impl Environment {
 
         (passthrough_build, passthrough_target)
     }
+    
+    fn volumes(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
+        let target_envs = self.get_target_var(target, "ENV_VOLUMES");
+        let mut volumes_target = None;
+        if let Some(envs) = target_envs {
+            volumes_target = Some(vec![]);
+            volumes_target.as_mut().map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
+        }
+        let build_envs = self.get_build_var("ENV_VOLUMES");
+        let mut volumes_build = None;
+        if let Some(envs) = build_envs {
+            volumes_build = Some(vec![]);
+            volumes_build.as_mut().map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
+
+        }
+        // validate no =
+
+        (volumes_build, volumes_target)
+    }
+
+    
 }
 
 #[derive(Debug)]
@@ -148,8 +169,33 @@ impl Config {
         }
         println!("{:?}", collect);
         Ok(collect)
-
+        
     }
+    
+    pub fn env_volumes(&self, target: &Target) -> Result<Vec<String>> {
+        let mut collect = vec![];
+        let (build_env, target_env) = self.env.volumes(target);
+        if let Some(mut vars) = build_env {
+            collect.extend(vars.drain(..));
+
+        } else {
+            if let Some(ref toml) = self.toml {
+                collect.extend(toml.env_volumes_build()?.drain(..).map(|v| v.to_string()));
+            }
+        }
+        if let Some(mut vars) = target_env {
+            collect.extend(vars.drain(..));
+
+        } else {
+            if let Some(ref toml) = self.toml {
+                collect.extend(toml.env_volumes_target(target)?.drain(..).map(|v| v.to_string()));
+            }
+        }
+        println!("env volumes {:?}", collect);
+        Ok(collect)
+    
+    }
+    
 
     
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,21 +1,37 @@
 use crate::{Result, Target, Toml};
 
-use std::env::var;
+use std::{collections::HashMap, env::var};
+#[cfg(not(test))]
 #[derive(Debug)]
 struct Environment(&'static str);
+#[cfg(test)]
+#[derive(Debug)]
+struct Environment(&'static str, HashMap<&'static str, &'static str>);
+
 
 impl Environment {
-    
-    fn build_var_name(&self, path: &str) -> String {
-        format!(
-            "{}_{}",
-            self.0,
-            path.to_ascii_uppercase().replace("-", "_")
-        )
+
+    #[cfg(not(test))]
+    fn new()-> Self {
+        Environment("CROSS")
+    }
+    #[cfg(test)]
+    fn new()-> Self {
+        Environment("CROSS", HashMap::new())
     }
 
-    fn get_var(name: &str) -> Option<String> {
+
+    fn build_var_name(&self, name: &str) -> String {
+        format!("{}_{}", self.0, name.to_ascii_uppercase().replace("-", "_"))
+    }
+
+    #[cfg(not(test))]
+    fn get_var(&self, name: &str) -> Option<String> {
         var(name).ok().and_then(|v| Some(v))
+    }
+    #[cfg(test)]
+    fn get_var(&self, name: &str) -> Option<String> {
+        self.1.get(name).and_then(|v| Some(v.to_string()))
     }
     fn target_path(target: &Target, key: &str) -> String {
         format!("TARGET_{}_{}", target.triple(), key)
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -0,0 +1,383 @@
...
+        if let Some(envs) = target_envs {
+            passthrough_target = Some(vec![]);
+            passthrough_target
+                .as_mut()
+                .map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
+        }
+        let build_envs = self.get_build_var("env_passthrough");
+        let mut passthrough_build = None;
+        if let Some(envs) = build_envs {
+            passthrough_build = Some(vec![]);
```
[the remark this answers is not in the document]
reitermarkus: The same applies to all other occurences of this pattern with `Some(vec![])`.

src/config.rs
```diff
@@ -0,0 +1,383 @@
...
+                .map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
+        }
+        let build_envs = self.get_build_var("ENV_VOLUMES");
+        let mut volumes_build = None;
+        if let Some(envs) = build_envs {
+            volumes_build = Some(vec![]);
+            volumes_build
+                .as_mut()
+                .map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
+        }
```
reitermarkus, line 120: This pattern already occurs 4 times, this should be refactored into a function.

src/config.rs
```diff
@@ -0,0 +1,383 @@
...
+        let build_envs = self.get_build_var("ENV_VOLUMES");
+        let mut volumes_build = None;
+        if let Some(envs) = build_envs {
+            volumes_build = Some(vec![]);
+            volumes_build
+                .as_mut()
+                .map(|vec| vec.extend(envs.split_whitespace().map(|v| v.to_string())));
```
reitermarkus, line 119: `split_whitespace` will break volume paths containing spaces. I'd suggest using `:` as separator, same as e.g. `PATH`.

src/config.rs
```diff
@@ -0,0 +1,383 @@
...
+        };
+        Ok(None)
+    }
+    pub fn image(&self, target: &Target) -> Result<Option<String>> {
+        let env_value = self.env.image(target);
+        if let Some(env_value) = env_value {
+            return Ok(Some(env_value));
+        }
+        self.toml.as_ref().map_or(Ok(None), |t| t.image(target))
+    }
```
reitermarkus, line 175: Add whitespace between functions.

```suggestion
    }
    
```

reitermarkus (changes requested):
Looks good, except for some things the can be simplified/refactored.

src/config.rs
olitha: In my understanding toml and env should provide the same API to the user. The volume values only describe the environment variable names under which the path that is bound to docker is found. This would be the same regardless if config is done with toml or env. 
This lookup is done in docker.rs:
https://github.com/rust-embedded/cross/blob/4f213ba07b214d36225be63e5b72a6346937a699/src/docker.rs#L133-L137
(changes then from toml to config)
https://github.com/olitha/cross/blob/423dea64dbf046f642274173138d392c91a0d16f/src/docker.rs#L131-L136
So I think volumes can be split by whitespaces as well.

src/config.rs
olitha: Thanks! I refactored

src/config.rs
olitha: I reformatted

src/config.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    fn build_var_name(&self, name: &str) -> String {
+        format!("{}_{}", self.0, name.to_ascii_uppercase().replace("-", "_"))
+    }
+
+    #[cfg(not(test))]
+    /// get value from process env
+    fn get_var(&self, name: &str) -> Option<String> {
+        var(name).ok().and_then(|v| Some(v))
```
reitermarkus, line 26: `and_then` is unneeded.

src/config.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    fn runner(&self, target: &Target) -> Option<String> {
+        self.get_target_var(target, "RUNNER")
+    }
+
+    fn passthrough(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
+        let passthrough_target = self
+            .get_target_var(target, "env_passthrough")
+            .and_then(|ref s| Some(split_to_cloned_by_ws(s)));
```
reitermarkus, line 93: Use `map` for these to avoid the need for `Some`  the closure.

src/config.rs
reitermarkus: Yes, you're right.

src/config.rs
olitha: Thanks. used .ok()

src/config.rs
olitha: changed it thanks :)

src/rustc.rs
```diff
@@ -8,10 +8,11 @@ use crate::errors::*;
 use crate::extensions::CommandExt;
 
 pub struct TargetList {
-    triples: Vec<String>,
+    pub triples: Vec<String>,
 }
 
 impl TargetList {
+    // TODO remove pub
```
reitermarkus, line 15: Anything blocking this?

src/main.rs
```diff
@@ -347,77 +385,124 @@ impl Toml {
...
+    
     /// Returns the list of volumes to pass through for `target`,
     /// including volumes specified under `build` and under `target`.
-    pub fn env_volumes(&self, target: &Target) -> Result<Vec<&str>> {
+    fn env_volumes(&self, target: &Target) -> Result<Vec<&str>> {
         let mut bwl = self.build_env("volumes")?;
         let mut twl = self.target_env(target, "volumes")?;
         bwl.extend(twl.drain(..));
-
+
```
reitermarkus, line 453: Superfluous whitespace here.

src/main.rs
```diff
@@ -432,10 +517,12 @@ fn toml(root: &Root) -> Result<Option<Toml>> {
...
-                return Err(format!("couldn't parse {} as TOML table", path.display()).into())
+                return Err(format!("couldn't parse {} as TOML table", path.display()).into());
             },
         }))
     } else {
         Ok(None)
     }
 }
+
+
```
reitermarkus, line 528: Superfluous newline here.

src/rustc.rs
olitha: made it pub for using it in config tests

olitha:
I changed the Environment struct so that it can be constructed with an optional Map. I did not like the cfg(test) inside the "not test" code.
Makes this also sense for the Configuration? It uses an Environment (that must be created for tests to be independent of env variables during test). The env for config could also be made optional.

src/config.rs
```diff
@@ -0,0 +1,360 @@
+use crate::{Result, Target, Toml};
+
+use std::collections::HashMap;
+use std::env::var;
```
reitermarkus, line 4:
```suggestion
use std::env;
```

Calling it as `env::var` is a bit more readable I think.

src/config.rs
```diff
@@ -0,0 +1,360 @@
...
+        self.get_target_var(target, "IMAGE")
+    }
+
+    fn runner(&self, target: &Target) -> Option<String> {
+        self.get_target_var(target, "RUNNER")
+    }
+
+    fn passthrough(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
+        let passthrough_target = self
+            .get_target_var(target, "env_passthrough")
```
reitermarkus, line 79:
```suggestion
            .get_target_var(target, "ENV_PASSTHROUGH")
```

src/config.rs
```diff
@@ -0,0 +1,360 @@
...
+    fn target_path(target: &Target, key: &str) -> String {
+        format!("TARGET_{}_{}", target.triple(), key)
+    }
+
+    fn build_path(key: &str) -> String {
+        format!("BUILD_{}", key)
+    }
+
+    fn get_build_var(&self, key: &str) -> Option<String> {
+        self.get_var(&self.build_var_name(&Environment::build_path(key)))
```
reitermarkus, line 33:
```suggestion
        self.get_var(&self.build_var_name(Self::build_path(key)))
```

src/config.rs
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    fn get_target_var(&self, target: &Target, key: &str) -> Option<String> {
+        self.get_var(&self.build_var_name(&Environment::target_path(target, key)))
```
reitermarkus, line 37:
```suggestion
        self.get_var(&self.build_var_name(&Self::target_path(target, key)))
```

src/config.rs
```diff
@@ -0,0 +1,360 @@
...
+        self.get_target_var(target, "RUNNER")
+    }
+
+    fn passthrough(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
+        let passthrough_target = self
+            .get_target_var(target, "env_passthrough")
+            .map(|ref s| split_to_cloned_by_ws(s));
+
+        let passthrough_build = self
+            .get_build_var("env_passthrough")
```
reitermarkus, line 83:
```suggestion
            .get_build_var("ENV_PASSTHROUGH")
```

src/config.rs
```diff
@@ -0,0 +1,360 @@
...
+            .map(|ref s| split_to_cloned_by_ws(s));
+
+        (passthrough_build, passthrough_target)
+    }
+
+    fn volumes(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
```
reitermarkus, line 89: This and `passthrough` share the exact same logic.

olitha:
@reitermarkus Hi could you have a look? I would like to bring this to a an end

reitermarkus:
bors try

reitermarkus (approved):

PR #538, opened 2021-03-20
Issue 530: Fix container engine detection when podman-docker installed

### LINKED ISSUE #530: Failure when podman-docker is installed
If one has `podman-docker` installed (a wrapper for podman that emulates docker's CLI, needed for things like docker-compose to be compatible), cross fails:
```
cross build --release --target=aarch64-unknown-linux-gnu

Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
error: failed to open: /target/release/.cargo-lock

Caused by:
  Permission denied (os error 13)
```

It looks like cross is trying to use docker, but since my user doesn't have access to "docker" (the emulated interface) it fails. 

Without `podman-docker`, everything works perfectly fine using rootless podman containers.

Cross should probably either default to trying podman first and then falling back to docker, or have some sort of check so this doesn't happen.

alexhunt7:
When using `podman` with `podman-docker` installed, `cross` will fail to build with a permission error.
```
cross build --release --target=aarch64-unknown-linux-gnu

Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
error: failed to open: /target/release/.cargo-lock

Caused by:
  Permission denied (os error 13)
```

If using `cross` from master, this can be worked around by specifying `CROSS_CONTAINER_ENGINE=podman`, but this does not work for the latest released version (v0.2.1).

Even though this has a workaround, this should be fixed such that we do not need to specify `CROSS_CONTAINER_ENGINE`. Simply defaulting to `podman` first fixes it.

Resolves #530


reitermarkus:
If both `podman` and `docker` are installed, `docker` should be the default. Is there a way to detect if `docker` is `podman-docker`?

alexhunt7:
> If both `podman` and `docker` are installed, `docker` should be the default.

Why? If anything, I'd assume the opposite, since `docker` requires a daemon and root privileges to run, while `podman` can be run by normal users. Most distributions ship with `podman` in their default repositories, but not `docker`. Many of these users will have `podman-docker` installed, which causes `cross` to hit the permissions error if `cross` treats it like `docker`.

I assume there are very few users who would have both `docker` and `podman` installed. Even if we do hit one, defaulting to using `podman` causes no harm, while defaulting to `docker` breaks things.

> Is there a way to detect if `docker` is `podman-docker`?

If `podman-docker` is installed, `/usr/bin/docker` will be a script that simply passes all the arguments to `podman`. There is also a symlink from /run/docker.sock to /run/podman/podman.sock . Searching for the word `podman` in `/usr/bin/docker` is probably the simplest way.

```
[alex:~] 17s $ dnf repoquery -l podman-docker | grep -v /usr/share/man
Last metadata expiration check: 0:00:43 ago on Tue 30 Mar 2021 10:33:43 PM PDT.
/usr/bin/docker
/usr/lib/tmpfiles.d/podman-docker.conf
/usr/bin/docker
/usr/lib/tmpfiles.d/podman-docker.conf
```
```
[alex:~] $ cat /usr/bin/docker
#!/usr/bin/sh
[ -f /etc/containers/nodocker ] || \
echo "Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg." >&2
exec /usr/bin/podman "$@"
```
```
[alex:~] $ cat /usr/lib/tmpfiles.d/podman-docker.conf
L+  /run/docker.sock   -    -    -     -   /run/podman/podman.sock
```
```
[alex:~] $ ll /run/docker.sock
lrwxrwxrwx. 1 root root 23 Mar 30 22:29 /run/docker.sock -> /run/podman/podman.sock
```

From within a container, there are several (potentially unreliable) ways https://github.com/containers/podman/issues/3586 .

Elinvynia:
What is needed to get this merged?

Currently this issue is preventing me from using cross at all and I would like to help.

Emilgardis:
I think this change make sense, I don't use podman but the reasoning to use it when both docker and podman is installed makes sense to me.

Right now we're locked by CI (#609), and we're in the process of a migration, see rust-embedded/wg#590

Alexhuszagh (commented):
I've fixed the merge issues, and I'd personally approve of these changes since podman defaults to less permissions and doesn't require a daemon. I'll let another maintainer be the final judge, however, since although podman aims to be entirely compatible with docker, this is a change in the default cross behavior.

Emilgardis (approved):

Emilgardis:
@reitermarkus you previosuly mentioned this shouldn't be done. has your position changed?

reitermarkus:
I think it makes sense to be the default on Linux. I think on macOS and Windows `podman` is only usable as a remote client so it should not be the default for that reason.

Alexhuszagh:
> I think it makes sense to be the default on Linux. I think on macOS and Windows `podman` is only usable as a remote client so it should not be the default for that reason.

I've updated it to do this by default. I've used the check to default to podman first if the OS is Linux, rather than use podman on anything other than Windows or macOS, since podman can't run natively on anything other than Linux anyway, even if it gets support on the BSDs or similar (also, I doubt Docker will ever support FreeBSD).

Alexhuszagh:
Uh I tried quashing multiple commits into one, but I think I failed spectacularly, which was not what I intended based on the [documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally).
[@cross/issue-530](https://github.com/cross-rs/cross/tree/issue-530)

PR #540, opened 2021-03-21
Allow mount_path to be specified when defining volumes.

TarekkMA:
The previous implementation of this feature, would result in the `host_path` and the `mount_path` to be the same with not option to make a custom mount path.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -134,8 +134,16 @@ pub fn run(target: &Target,
             validate_env_var(var)?;
 
             if let Ok(val) = env::var(var) {
-                let host_path = Path::new(&val).canonicalize()?;
-                let mount_path = &host_path;
+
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
Ah I can see why this would be needed. Say I need to include data from `$HOME/.config/mydata`, and since the home directory will be different paths for the host and mount, then you want to be able to mount this to the same logical point in the container? The logic has changed significantly in our code, so this will likely need a fresh PR, but I'm not opposed to this on principal.

Alexhuszagh:
Actually closing since you can do this via `DOCKER_OPTS`, and for very esoteric behavior like this that would require much more than "zero" setup, since it is almost certainly host-specific, you can always provide it via another config option, and isn't a valid path which is what we planned to support.

PR #544, opened 2021-03-31
Remove extra slash in project mount 

TarekkMA:
Old code was adding extra slash
```
 "-v" "/Users/runner/work/my-rust-project://Users/runner/work/my-rust-project:Z"
```


Emilgardis:
I remember there being a specific reason for this. Ill try and dig it up unless someone else knows why.

PR #549, opened 2021-04-08
don't set `SQLITE3_LIB_DIR` in FreeBSD images

### LINKED ISSUE #520: Compiling `diesel-migrations` for FreeBSD fails.
Compiling `diesel-migrations` with `cross build --target=x86_64-unknown-freebsd` results in the following error:

```   Compiling diesel_migrations v1.4.0
error: libthr.so.3: cannot open shared object file: No such file or directory
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/diesel_migrations-1.4.0/src/lib.rs:82:1
   |
82 | extern crate migrations_macros;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: could not compile `diesel_migrations`
```

`Cargo.toml`:
```[package]
name = "ohgodwhy"
version = "0.1.0"
authors = ["gahag <gabriel.s.b@live.com>"]
edition = "2018"

[dependencies]
diesel = { version = "1.4", features = ["sqlite"] }
diesel_migrations = { version = "1.4", features = ["sqlite"] }
```
`main.rs`:
```
fn main() {
    println!("Hello, world!");
}
```

Is this an issue with the cross' FreeBSD setup, or with diesel-migrations?

tesaguri:
The environment variable has broken procedural macro crates which depend on `libsqlite3-sys`, such as `migrations_macros`.

Even without the variable, `libsqlite3-sys` finds the correct path anyway.

This fixes #520, as shown in the CI at [tesaguri/cross-freebsd-diesel_migrations-test] repository.

~Additionally, this PR updates `freebsd{,-extras}.sh` to use OpenSSL from the base system, because the URLs `https://pkg.freebsd.org/FreeBSD:12:{amd64,i686}/quarterly/All/openssl-1.1.1j,1.txz` returns `404` now.~

[tesaguri/cross-freebsd-diesel_migrations-test]: https://github.com/tesaguri/cross-freebsd-diesel_migrations-test

### DIFF
--- a/docker/Dockerfile.i686-unknown-freebsd
+++ b/docker/Dockerfile.i686-unknown-freebsd
@@ -18,5 +18,4 @@ RUN /freebsd-extras.sh i686
 ENV CARGO_TARGET_I686_UNKNOWN_FREEBSD_LINKER=i686-unknown-freebsd12-gcc \
     CC_i686_unknown_freebsd=i686-unknown-freebsd12-gcc \
     CXX_i686_unknown_freebsd=i686-unknown-freebsd12-g++ \
-    I686_UNKNOWN_FREEBSD_OPENSSL_DIR=/usr/local/i686-unknown-freebsd12/  \
-    SQLITE3_LIB_DIR=/usr/local/x86_64-unknown-freebsd12/lib/
+    I686_UNKNOWN_FREEBSD_OPENSSL_DIR=/usr/local/i686-unknown-freebsd12/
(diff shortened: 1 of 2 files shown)


reitermarkus (dismissed):

tesaguri:
Rebased onto `master` and removed commit 9a2b65ca9b9836b171a30f3dace4e692c84573d2 since it somehow caused `curl` to fail at certificate verification and the commit seems to be no longer necessary anyway.

Still, test on `x86_64-pc-windows-msvc` is failing:

<https://dev.azure.com/rust-embedded/cross/_build/results?buildId=945&view=logs&j=a3d15252-dc4a-56ed-f6b6-444e7d9eafda&t=9d375de8-baa3-57ad-6628-9eaf4ea568de&l=82>

But I don't think it's related to this change. It seems that `libc`, starting from v0.2.103, has began to depend on trait implementations for arbitrary length arrays (in rust-lang/libc#2409), which was stabilized in Rust 1.47.0 ([release note]).

[release note]: https://github.com/rust-lang/rust/blob/1.47.0/RELEASES.md#libraries

tesaguri:
Rebased onto master. The MSRV issue seems to have already been fixed by the upstream.

reitermarkus (approved):

PR #552, opened 2021-04-20
Add changelog.yml GHA workflow

therealprof:
Adds a GHA check to ensure every PR is accompanied by a edit of CHANGELOG.md to list the changes so we don't run into the usual issue of having to come up with all entries when someone requests a release.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,5 +1,5 @@
 on:
-  pull_request_target:
+  pull_request:
 
 name: Changelog check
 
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  pull_request_target:
```
therealprof, line 2:
```suggestion
  pull_request:
```

adamgreig (approved):
thanks!

bors merge

PR #554, opened 2021-05-02
Add `--force-non-host` flag.

reitermarkus:
Closes https://github.com/rust-embedded/cross/issues/536.

### DIFF
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -130,10 +130,31 @@ cross will not attempt to configure the toolchain further so that it can run you
     })
 }
 
+fn version(msg_info: &mut MessageInfo) -> Result<Version> {
+    let out = rustup_command(msg_info, false)
+        .arg("--version")
+        .run_and_get_stdout(msg_info)?;
+
+    match out
+        .lines()
+        .next()
+        .and_then(|line| line.split_whitespace().nth(1))
+    {
+        Some(version) => {
+            semver::Version::parse(version).wrap_err_with(|| "failed to parse rustup version")
+        }
+        None => eyre::bail!("failed to get rustup version"),
+    }
+}
+
 pub fn install_toolchain(toolchain: &QualifiedToolchain, msg_info: &mut MessageInfo) -> Result<()> {
+    let mut command = rustup_command(msg_info, false);
     let toolchain = toolchain.to_string();
-    rustup_command(msg_info, false)
-        .args(&["toolchain", "add", &toolchain, "--profile", "minimal"])
+    command.args(&["toolchain", "add", &toolchain, "--profile", "minimal"]);
+    if version(msg_info)? >= semver::Version::new(1, 25, 0) {
+        command.arg("--force-non-host");
+    }
+    command
         .run(msg_info, false)
         .wrap_err_with(|| format!("couldn't install toolchain `{toolchain}`"))
 }
--- a/.changes/554.json
+++ b/.changes/554.json
@@ -1,0 +1,5 @@
+{
+    "type": "fixed",
+    "description": "add the `--force-non-host` flag for newer rustup versions",
+    "issues": [536]
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,7 @@ dependencies = [
  "owo-colors",
  "regex",
  "rustc_version",
+ "semver",
  "serde",
  "serde_ignored",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ directories = "4.0.1"
 walkdir = { version = "2", optional = true }
 tempfile = "3.3.0"
 owo-colors = { version = "3.4.0", features = ["supports-colors"] }
+semver = "1"
 
 [target.'cfg(not(windows))'.dependencies]
 nix = { version = "0.24", default-features = false, features = ["user"] }


Dylan-DPC (approved):

Dylan-DPC:
bors: r+

reitermarkus:
bors retry

reitermarkus:
@rbtcollins, `rustup toolchain install` doesn't seem to have this option and `cross` doesn't call `rustup update` anywhere.

Emilgardis:
just a small nitpick, should this be backported into https://github.com/rust-embedded/cross/tree/v0.2.1 ?

rbtcollins:
Hmmmm, will look into that. Internally the two are basically aliases.

Emilgardis:
What's the status on this? @rbtcollins

rbtcollins:
We added the flag to rustup, and its in a deprecation period where it isn't required. I encourage merging the use of the flag, so that we can move forward with making it required.

There is a UI glitch associated with it triggering unnecessarily in some cases on Windows, but we should get that fixed for the next release I hope

Emilgardis:
bors retry

adamgreig:
It looks like this is failing on MacOS and Windows builds with:

```
+ cross check --target x86_64-apple-darwin
error: Found argument '--force-non-host' which wasn't expected, or isn't valid in this context
	Did you mean --force?

USAGE:
    rustup toolchain install <toolchain>... --force

For more information try --help
error: couldn't install toolchain `1.42.0-x86_64-unknown-linux-gnu`
caused by: `"rustup" "toolchain" "install" "--force-non-host" "1.42.0-x86_64-unknown-linux-gnu"` failed with exit code: Some(1)
note: run with `RUST_BACKTRACE=1` for a backtrace
```

i.e., related to adding `--force-non-host`, any ideas?

reitermarkus:
> We added the flag to rustup, and its in a deprecation period where it isn't required.

@rbtcollins, `rustup toolchain install` still doesn't accept this flag.

Emilgardis:
https://github.com/rust-lang/rustup/pull/2968 has been merged, not sure when next release is

otavio:
As rustup has support for it, are this going in 0.2.2?

Emilgardis:
> As rustup has support for it, are this going in 0.2.2?

this issue is blocked on new rustup release, it can't be included without upstream support

Alexhuszagh:
> As rustup has support for it, are this going in 0.2.2?

We need to wait for a release that supports it, because `1.24.3` doesn't, which I believe is the latest? And we check currently for `1.24.0`.

```bash
$ rustup toolchain install 1.58.0 --force-non-host
error: Found argument '--force-non-host' which wasn't expected, or isn't valid in this context
        Did you mean --force?

USAGE:
    rustup toolchain install <toolchain>... --force

For more information try --help
$ rustup -V
rustup 1.24.3 (ce5817a94 2021-05-31)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.61.0 (fe5b13d68 2022-05-18)
```

So we need to wait for a later release, but yes, if this is done prior to our `0.22` release, I think we should add this.

otavio:
We could check for 1.24.4 as it has been merged already.

Alexhuszagh (approved):
Ok it seems like rustup doesn't have different branches, so it's safe to assume this will be in 1.24.4. @Emilgardis look good?

Alexhuszagh:
I accidentally rebased it off the wrong branch, I've since fixed it. @reitermarkus can you squash the commits down to a single one and we can merge it?

Emilgardis:
rustup 1.25.0 has been released, containing the fix

https://blog.rust-lang.org/2022/07/11/Rustup-1.25.0.html

Alexhuszagh:
We should probably merge this quickly then.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #885 - handle symlinks when using remote docker.
 - #868 - ignore the `CARGO` environment variable.
 - #867 - fixed parsing of `build.env.passthrough` config values.
+- #554 - Add `--force-non-host` flag.
```
Emilgardis, line 47: wrong entry, should be  a .changes file

src/rustup.rs
```diff
@@ -104,9 +104,29 @@ pub fn available_targets(toolchain: &str, msg_info: &mut MessageInfo) -> Result<
...
+
 pub fn install_toolchain(toolchain: &str, msg_info: &mut MessageInfo) -> Result<()> {
-    rustup_command(msg_info, false)
-        .args(&["toolchain", "add", toolchain, "--profile", "minimal"])
+    let mut command = rustup_command(msg_info, false);
+    command.args(&["toolchain", "add", toolchain, "--profile", "minimal"]);
+    if version(msg_info)? >= semver::Version::new(1, 25, 0) {
+        command.arg("--force-non-host");
+    }
+    command
```
Emilgardis, line 129: I don't think we should do this, but rather always try `--force-non-host`, and if that fails then try without

Emilgardis (changes requested):

CHANGELOG.md
Emilgardis: with the Github cli it's easy

`gh pr checkout 554`, it sets up the remote and branch (with correct upstream ref)

 
also possible to do with dev.github.com on the pr, just add `dev.*` to the url

Alexhuszagh: I've tried that, it keeps pushing to the wrong remote.

Alexhuszagh: Wait this worked for the first time ever.

src/rustup.rs
```diff
@@ -104,10 +104,18 @@ pub fn available_targets(toolchain: &str, msg_info: &mut MessageInfo) -> Result<
...
+
 pub fn install_toolchain(toolchain: &str, msg_info: &mut MessageInfo) -> Result<()> {
-    rustup_command(msg_info, false)
-        .args(&["toolchain", "add", toolchain, "--profile", "minimal"])
+    install_toolchain_command(toolchain, msg_info)
+        .arg("--force-non-host")
         .run(msg_info, false)
+        .or_else(|_| install_toolchain_command(toolchain, msg_info).run(msg_info, false))
         .wrap_err_with(|| format!("couldn't install toolchain `{toolchain}`"))
 }
```
Emilgardis, line 120: now that I think about this some more, maybe it's not a good idea to have it this simplistic, two rustup calls is probably the best solution. I should've requested/implemented `--force-non-host` as a env var as well :/

Emilgardis: The reason I think this is not good is because there'll be some strange behaviour if both calls are invalid/rustup is not properly configured.

Emilgardis (changes requested):

src/rustup.rs
Alexhuszagh: Is there a better solution though? I think the version is better since we capture `stdout` and error if it fails, and then can check if the version is correct. We get the appropriate error message. The downside is we have 2 rustup calls always.

Alexhuszagh: Just a ping on this: what's the best design here?

Emilgardis: lets keep it as is, but don't throw away the first error

src/rustup.rs
```diff
@@ -130,11 +130,26 @@ cross will not attempt to configure the toolchain further so that it can run you
     })
 }
 
-pub fn install_toolchain(toolchain: &QualifiedToolchain, msg_info: &mut MessageInfo) -> Result<()> {
+pub fn install_toolchain_command(
+    toolchain: &QualifiedToolchain,
```
Emilgardis, line 134: Move this function to inside `install_toolchain`

src/rustup.rs
```diff
@@ -130,11 +130,26 @@ cross will not attempt to configure the toolchain further so that it can run you
...
+}
+
+pub fn install_toolchain(toolchain: &QualifiedToolchain, msg_info: &mut MessageInfo) -> Result<()> {
+    install_toolchain_command(toolchain, msg_info)
+        .arg("--force-non-host")
         .run(msg_info, false)
+        .or_else(|e| {
+            install_toolchain_command(toolchain, msg_info)
+                .run(msg_info, false)
+                .map_err(|_ignore| e)
```
Emilgardis, line 151: This now throws away this error instead.

Emilgardis (changes requested):
Move install

src/rustup.rs
Alexhuszagh: Should we wrap the error then? I felt it would produce redundant error messages.

Emilgardis: how about

```
.or_else(|e| {
            install_toolchain_command(toolchain, msg_info)
                .run(msg_info, false)
		.wrap_err("installing toolchain with and without `--force-non-host` failed")
```

Alexhuszagh: For better error handling, I've reverted to using version checking.

Emilgardis (approved):
bors r+

PR #563, opened 2021-06-26
Crates.io quality of life improvement

Nicoretti:
Hi @all,

thx for the great tool, realy appricate your efforts!
Just referenced the REAMDE.md within the Cargo.toml meta data, so people on crates.io
have a better experience with it.

Best
Nico

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ name = "cross"
 repository = "https://github.com/rust-embedded/cross"
 version = "0.2.1"
 edition = "2018"
+readme = "README.md"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for this PR! I was a bit confused why the crates.io page doesn't already show the README, but it turns out our last release was before Cargo added a fault value for this field. Since 2020-08-27, the `readme` field [now has a default value of `README.md`](https://github.com/rust-lang/cargo/pull/8277) if unspecified, but our last release was 2020-06-20, just a month beforehand.

So, I believe any future release should already include the README without needing to set this field.

PR #567, opened 2021-07-05
Add support for native AArch64 hosts

davidcole1340:
Currently, running `cross` on an AArch64 host requires you to run x86 Docker images. This PR adds support for running Docker images that have been built with an AArch64 architecture.

I haven't finished going through the dockerfiles to check they work with the Ubuntu AArch64 base image however thought I'd get the PR out there to get feedback :). I have got the dockerfile for `arm-unknown-linux-gnueabihf` working by building the GCC toolchain while building the container (the old toolchain targeted ARMv7 anyway?) which was based on the dockerfile from [raspi-toolchain](https://github.com/Pro/raspi-toolchain).


aig787:
@davidcole1340 I'd really love this feature merged as well, I'm happy to pitch in if you need an extra set of hands

davidcole1340:
@aig787 Hi Aaron, unfourtunately I don't have the bandwidth to put any time into this PR for the next month but feel free to fork off my previous work if you'd like. If not, I'd love an extra set of hands on it and I can have a deeper look later in Feb. Most of the work required is making the Docker images compile on AArch64 hosts.

Main problems I came up against:
- A lot of packages are only available in the later Ubuntu releases. This means the glibc version needs to be bumped (which I think is what we are trying to avoid, but best to clarify with the team on Matrix).
- Android NDK was flat out not supported when I last checked.
- Emscripten was unsupported or required some work-arounds.
- Among others...

Not a contribution.

otavio:
Are you, @aig787, still volunteering to adopt this PR?

Emilgardis:
I'm working on this, ill create a issue for proper assignment. 

Closing this issue as it's stale, thanks for your initial work @davidcole1340 !

PR #573, opened 2021-07-22
Repro builds: Deterministically iterate through docker/ folder

kpcyrd:
One of the Arch Linux rebuilders has trouble reproducing the [cross binary](https://wolfpit.net/rebuild/api/v0/builds/24560/diffoscope) we distribute in Arch Linux. We assume this is because it runs a different filesystem than the official build server, causing `fs::read_dir` to yield the directory content in a different order.

This patch explicitly sorts the directory content of docker/ in build.rs before constructing a list that gets embedded in the binary.

Full writeup can be found over here: https://twitter.com/sn0int/status/1418254640324521997

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
 use std::env;
 use std::error::Error;
-use std::fs::{read_dir, File};
-use std::io::Write;
+use std::fs::File;
+use std::io::{self, Write};
 use std::path::PathBuf;
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


adamgreig:
bors try

adamgreig (approved):
Thank you, and nice write-up too!

bors merge

kpcyrd:
Thanks! Are there any plans for a new release? I've noticed there hasn't been one since about a year. :)

adamgreig:
Yea, we'd like a new release (https://github.com/rust-embedded/cross/issues/542) but currently have a couple last blockers (and should ideally get a few more pending PRs merged...). The main issue is we're desperately low on maintainer time, since most of the rust-embedded team nominally responsible don't use cross or know much about it. I've just posted #574 with details of that.

PR #582, opened 2021-08-28
freebsd: Add libprocstat.so

svenstaro:
If we don't add this library, we'll eventually run into linking problems due to libprocstat.so not being found.


svenstaro:
I have no idea why this fails to build.

Emilgardis:
Its unrelated, probably https://github.com/nix-rust/nix/issues/1491

Need to bump nix version

Emilgardis:
bors try

svenstaro:
So bors says it works but the CI still fails?

Emilgardis:
We merged a fix for the CI in #583, but github actions doesn't run on the (would be) merge commit, it runs on the prs ref.
If you want to fix it, you could rebase your branch, but there is no need, as bors runs on the merged ref.

I don't really feel comfortable merging this as I'm not familiar with FreeBSD. But it doesn't seem like it'd be bad to have.

Emilgardis:
@reitermarkus r=me if you think this is ok, i suspect it is

svenstaro:
Rebased and added CHANGELOG entry.

> I don't really feel comfortable merging this as I'm not familiar with FreeBSD. But it doesn't seem like it'd be bad to have.

Since this really just adds a library I think the possible bad impact is fairly limited. Without this, my fairly boring Rust program didn't compile and with it, it does. :)

Emilgardis (approved):
Thanks!

svenstaro:
Thank _you_ for taking care of cross. I think cross fills an important role in the Rust ecosystem and it needs love. :)

svenstaro:
Any chance for a good ol' mergerino?

Emilgardis:
ill merge it, don't see that should be an issue.

Thank you!

bors r+

svenstaro:
Huh how did this fail?

Emilgardis:
https://dev.azure.com/rust-embedded/cross/_build/results?buildId=930&view=logs&j=6e847cca-24ce-5b2d-e476-7fadc4226834&t=4f4f83a6-27d6-5138-5def-6e63a5c30d1a&l=16603

404, could be temporary or it needs to be changed

svenstaro:
It needs to be updated to https://pkg.freebsd.org/FreeBSD:12:i386/quarterly/All/openssl-1.1.1l,1.txz. I'm submitting another PR I guess.

svenstaro:
Alright @Emilgardis I posted a fix as #588.

PR #583, opened 2021-08-28
bump nix

Emilgardis:


Emilgardis:
CI should probably be ran with `--locked` deps, but this should fix previously seen errors

adamgreig (approved):
bors merge

PR #586, opened 2021-09-04
Remove hard-coded old Qemu version for PPC targets.

### LINKED ISSUE #585: Remove Hard-Coding of Qemu Versions on PPC
Qemu versions 4.2.0 have [fixed](https://bugs.launchpad.net/qemu/+bug/1821444) floating-point math bugs that were introduced in version 3.1.0. This means the following work-around is no longer necessary, as long as we use Qemu 4.2.0+:
https://github.com/rust-embedded/cross/blob/44f31b887c866637910baa6e7424d3ff5afdda2d/docker/qemu.sh#L72-L79

Since the local version we use is already 4.2+, this fix can be applied immediately.

Alexhuszagh:
Fixes #585. The relevant upstream issues that have made this possible are [1821444](https://bugs.launchpad.net/qemu/+bug/1821444) ("qemu-ppc (user) incorrectly translates float32 arithmetics"), so this initial [workaround](https://github.com/rust-embedded/cross/pull/316)  is no longer required.

### DIFF
--- a/docker/qemu.sh
+++ b/docker/qemu.sh
@@ -69,15 +69,6 @@ build_static_pixman() {
 main() {
     local version=4.2.0
 
-    # Qemu versions 3.1.0 and above break 32-bit float conversions
-    # on powerpc, powerpc64, and powerpc64le. Last known working version
-    # is 3.0.1.
-    # Upstream Issue:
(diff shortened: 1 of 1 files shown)


reitermarkus:
Included in https://github.com/cross-rs/cross/pull/609.

PR #593, opened 2021-09-06
Ensure build steps use non-interactive package installs.

### LINKED ISSUE #589: Ensure `DEBIAN_FRONTEND=noninteractive` is set during image builds
For apt installations,  some dependencies may default to interactive user input, such as tzdata. This can be especially an issue when upgrading dependencies, since existing build configurations may break due to new dependencies.

The fix, therefore, is in all all Dockerfiles to specify the following. Using `ARG` rather than `ENV` is preferably since it is only used during the build process, and is not present after the Docker image has been completed.

```dockerfile
ARG DEBIAN_FRONTEND=noninteractive
RUN apt-get install --assume-yes --no-install-recommends \
    g++-arm-linux-gnueabi \
    libc6-dev-armel-cross && \
    /qemu.sh arm
```

And in all scripts:

```bash
export DEBIAN_FRONTEND=noninteractive
```

This has been useful when upgrading dependencies, since of the newer versions do actually have issues with interactive console input.

Alexhuszagh:
Fixes #589. Some libraries, such as `tzdata`, can ask for user input
during installation, which is problematic when building a Docker image,
since it prevents automatation and stdin i typically not attached. This
has also been an issue upgrading Docker images to the latest Ubuntu
versions.

### DIFF
--- a/docker/Dockerfile.aarch64-linux-android
+++ b/docker/Dockerfile.aarch64-linux-android
@@ -1,4 +1,5 @@
 FROM ubuntu:20.04
+ARG DEBIAN_FRONTEND=noninteractive
 
 COPY common.sh lib.sh /
 RUN /common.sh
--- a/docker/Dockerfile.aarch64-unknown-linux-gnu
+++ b/docker/Dockerfile.aarch64-unknown-linux-gnu
@@ -1,4 +1,5 @@
 FROM ubuntu:16.04
+ARG DEBIAN_FRONTEND=noninteractive
 
 COPY common.sh lib.sh /
 RUN /common.sh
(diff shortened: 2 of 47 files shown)


reitermarkus:
Does it make sense to set this in `lib.sh` instead?

Alexhuszagh:
> Does it make sense to set this in `lib.sh` instead?

That should work, I'll redo this PR then. Seems like a much "drier" approach.

Emilgardis (approved):
bors r+

Emilgardis:
oh yeah, need someone else to r+ #690

Emilgardis:
bors cancel

#695

Issue #598, opened 2021-10-08
aarch64 builds using nightly fail with missing `getauxval` in `init_have_lse_atomics` in `cpu_model.c` in `compiler_builtins`


tonyg:
## [Updated] Workaround

See [this comment](#issuecomment-942138394).

## To reproduce

1. `mkdir foo`
2. `cd foo`
3. `cargo init`
4. `cross build --target=aarch64-unknown-linux-musl --release --all-targets` -- note that this *succeeds*
5. `echo 'nightly' > rust-toolchain`
6. `cross build --target=aarch64-unknown-linux-musl --release --all-targets` -- note that this *fails*

## Actual outcome

Step 6 fails with

```
$ cross build --target=aarch64-unknown-linux-musl --release --all-targets
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.57.0-nightly (485ced56b 2021-10-07)

info: checking for self-updates
   Compiling foo v0.1.0 (/project)
error: linking with `aarch64-linux-musl-gcc` failed: exit status: 1
  |
[...]

PR #599, opened 2021-10-13
Build binary crate in CI test with nightly as well as stable

tonyg:
Hopefully this isn't a completely silly idea :) but I wonder if (in connection with #598) building with nightly as part of CI makes sense? This PR makes an extremely small change to `test.sh` which should (?) build the binary `hello` crate both with stable and nightly toolchains.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -88,6 +88,8 @@ EOF
         cargo init --bin --name hello .
         retry cargo fetch
         cross build --target "${TARGET}"
+        echo nightly > rust-toolchain
+        cross build --target "${TARGET}"
         popd
 
         rm -rf "${td}"


tonyg:
Promisingly, ci is failing on aarch64-unknown-linux-musl as expected per #598 ...

reitermarkus:
It seems arbitrary to only test nightly in that specific spot. Generally I see testing nightly as only making our CI more error prone without having an easy way to fix it when it happens.

When something breaks on nightly, it will either need to be fixed upstream in Rust or in the Docker image by someone who is using that target and can actually debug the problem effectively.

otavio:
I don't see much value in adding it. @reitermarkus made raised very good concerns about it and I agree fully with him. For me, this could be closed.

Alexhuszagh:
Also in agreement with closing this.

PR #600, opened 2021-10-13
Allow target to be specified with an environmental variable

rtsuk:
My TextMate-based development environment can't easily pass the
target on the command line.


Alexhuszagh:
This is now implemented in a different manner, using `CARGO_BUILD_TARGET`, in #804.

PR #601, opened 2021-10-15
Warn user if there's a file cross.toml with wrong capitalization

denisfalqueto:
I've lost some good days struggling with errors because I've created a file named cross.toml and it wouldn't affect the build, because the name was not capitalized correctly. So, I think it should be cool if cross could warn the user if there's a file with a capitalization problem present on the project root.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -538,6 +538,10 @@ fn toml(root: &Root) -> Result<Option<Toml>> {
             },
         }))
     } else {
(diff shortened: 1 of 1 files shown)


otavio (approved):

otavio:
@maintainers, this seems very simple and useful.

Emilgardis (approved):
thank you!

bors r+

Issue #368, opened 2020-02-06
Ability to set default build target


Progdrasil:
It would be useful for those who develop for one specific target to be able to define the default build target. In much the same way [cargo does]. This would reduce the following example command:
```console
cross build --target armv7-unknown-linux-gnueabihf 
```
To the equivalent `Cross.toml` and command
```toml
[build]
target = "armv7-unknown-linux-gnueabihf"
```
```console
cross build
```

[cargo does]: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget

bausano:
This would be much appreciated. `Cargo.toml` uses this format for setting default, maybe the same could be used by cross:

```
[target]
default = "armv7-unknown-linux-gnueabihf"
```

PR #609, opened 2021-11-18
switch to GHCR and use more github actions

Emilgardis:
addresses #607 #571

## what this does

- [x] change default registry to look for cross images to `ghcr.io`
- [x] move from azure CI for docker images to GHA
- [x] add clippy (that does not fail job, can be made to do this later) and rustfmt
- [x] bump QEMU to 5.1
- [ ] Add (better) layer caching

## Some nice things to have/include
- [x] Cache docker images, possibly using buildx (probably out of scope of this pr)
- [x] Use actions/cache
- [ ] make configuration easier to maintain
- [ ] possibly replace docker setup with actions

PR #624, opened 2022-01-08
Allow setting default build target

zappolowski:
This implements #368 

This feature allows to specify the default target to build either in `Cross.toml`

```toml
[build]
target = "aarch64-unknown-linux-gnu"
```
or by setting `CROSS_BUILD_TARGET` to avoid the need to always specify `--target`.

I've opted for `build.target` as this resembles the configuration of [cargo](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #624 - Add `build.default-target`
 - #670 - Use serde for deserialization of Cross.toml
 - Change rust edition to 2021 and bump MSRV for the cross binary to 1.58.1
 - #654 - Use color-eyre for error reporting
--- a/docs/cross_toml.md
+++ b/docs/cross_toml.md
@@ -6,6 +6,7 @@ The `build` key allows you to set global variables, e.g.:
 ```toml
 [build]
 xargo = true
+default-target = "x86_64-unknown-linux-gnu"
 ```
 
 # `build.env`
--- a/src/config.rs
+++ b/src/config.rs
@@ -78,6 +78,10 @@ impl Environment {
         self.get_values_for("ENV_VOLUMES", target)
     }
 
+    fn target(&self) -> Option<String> {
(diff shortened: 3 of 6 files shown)


Emilgardis (dismissed):
Looks good! Cant merge right now due to CI being incomplete

zappolowski:
> Cant merge right now due to CI being incomplete

Is some action required on my site to complete the CI?

Emilgardis:
No, see #609 <3

reitermarkus:
Does this need to be in `Cross.toml`? Can't this be in `.cargo/config.toml` and we automatically detect it?

zappolowski:
> Does this need to be in Cross.toml? Can't this be in .cargo/config.toml and we automatically detect it?

I've opted for using `Cross.toml` in the first place as it's a configuration made for `cross` and shouldn't affect `cargo`. Another issue was/is that according to [the documentation](https://doc.rust-lang.org/cargo/reference/config.html) the actual value can be defined in several places and duplicating the resolution implemented by `cargo` (and keeping it up to date) seemed unnecessary complex.

My proposal would be to postpone this feature until [cargo config](https://github.com/rust-lang/cargo/issues/9301) is stabilized as this should make it possible to just parse the output of `cargo config get --format toml build.target` and use this without the hassle mentioned above.

reitermarkus:
> My proposal would be to postpone this feature until https://github.com/rust-lang/cargo/issues/9301 is stabilized

Sounds fair.

Ideally I would like to get rid of `Cross.toml` eventually and use `package.metadata.cross` in `Cargo.toml` to store `cross` configuration.

otavio (approved):

otavio:
I don't see why this should be postponed as this is very useful and a valid use case. Once the cargo feature is merged this can be reworked in a new release.

Emilgardis:
can you please rebase this again so we can get this merged? There's been some changes with #670

Emilgardis:
I've renamed the option to `default-target` to better convey what it does

Emilgardis (approved):
bors r+

thanks!

Issue #565, opened 2021-06-29
Add pre-build hook


andrewbanchich:
I'd like to use an official image like `centos/devtoolset-7-toolchain-centos7`, which I can do by adding to my `Cross.toml`. However, the image doesn't come with `openssl` installed. It seems like overkill to have to create and maintain a Dockerfile, then add an extra build script to my CI just to get `openssl`.

Instead, would it be possible to do something like this?
```toml
[target.x86_64-unknown-linux-gnu]
image = "centos/devtoolset-7-toolchain-centos7"
pre-build = "yum install openssl-devel"
```

Emilgardis:
This has been implemented!

```toml
[target.x86_64-unknown-linux-gnu]
image = "centos/devtoolset-7-toolchain-centos7"
pre-build = ["yum install openssl-devel"]
```

should just work now with latest cross :)

PR #635, opened 2022-02-07
Add pre-build hooks

aig787:
This PR addresses #565 and hopefully a number of the open issues about the lack of openssl in the builders. It does add a dependency on [gosu](https://github.com/tianon/gosu) which handles lowering the privileges down to user level after the pre-build hooks are run. For now I've only added it to the x86_64-unknown-linux-gnu target for testing pending feedback on it.

I was able to build Cargo successfully with all features with
```toml
image = "rustembedded/cross:x86_64-unknown-linux-gnu-0.2.1"
pre-build = """
yum update -y
yum install -y openssl-devel
"""
```

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -146,20 +146,6 @@ pub fn run(
 
     docker.arg("--rm");
 
-    // We need to specify the user for Docker, but not for Podman.
-    if let Ok(ce) = get_container_engine() {
-        if ce.ends_with(DOCKER) {
-            docker.args(&[
-                "--user",
-                &format!(
-                    "{}:{}",
-                    env::var("CROSS_CONTAINER_UID").unwrap_or(id::user().to_string()),
-                    env::var("CROSS_CONTAINER_GID").unwrap_or(id::group().to_string()),
-                ),
-            ]);
-        }
-    }
-
     docker
         .args(&["-e", "XARGO_HOME=/xargo"])
         .args(&["-e", "CARGO_HOME=/cargo"])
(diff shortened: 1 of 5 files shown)


Emilgardis:
Awesome!

This is really a workaround but it's quite neat.
Why do we need to lower privileges?

aig787:
I'm not sure we necessarily need to - just seemed good practice to me especially since when hooks aren't provided the containers are run with the `--user` argument. Lowering privileges was just an attempt at getting us into the same state for the build whether they were provided or not.

src/docker.rs
```diff
@@ -206,9 +192,47 @@ pub fn run(
...
+            Some((
+                env::var("CROSS_CONTAINER_UID").unwrap_or(id::user().to_string()),
+                env::var("CROSS_CONTAINER_GID").unwrap_or(id::group().to_string()),
+            ))
+        } else {
+            None
+        }
+    } else {
+        None
+    };
```
otavio, line 207: Please move the code back to the original place as it is a non-required code change.

otavio (changes requested):
I just small code comment, but I'd like to say I like the idea, and it does avoid the need for many custom images.

Emilgardis:
marking as draft as only one target is implemented right now

aig787:
> marking as draft as only one target is implemented right now

I'm happy to get moving on the other targets if we're happy with the implementation. I'm also happy to revert the privilege de escalation first, just need to know what the preference is from your side.

Emilgardis:
> > marking as draft as only one target is implemented right now
> 
> I'm happy to get moving on the other targets if we're happy with the implementation. I'm also happy to revert the privilege de escalation first, just need to know what the preference is from your side.

I think the gid/uid setting is good, there is a reason we allow setting it.

Emilgardis:
see my comment on https://github.com/cross-rs/cross/issues/643#issuecomment-1086898664 regarding the "issue" that the image is now centos.

Emilgardis:
Been thinking about this some more now, would it make sense to actually not do this, and just use the technique in #678 and hook into a dockerfile passed to stdin to `docker build --build-arg cmd=blabla [additional flags] -`

```
FROM {target}
ARG CMD=exit 0
RUN $CMD
```

This way we can even support custom images

patrickelectric:
This is a great feature!

PR #636, opened 2022-02-08
Update FreeBSD quarterly package versions

mirekdlugosz:
FreeBSD updated their packages and current master fails to build docker image - as downloading packages fail.

Merging this PR does not necessitate new container image publication, it just fixes image building on master.


Emilgardis:
fixed in #609! Thanks for the pr anyway!

PR #613, opened 2021-11-23
freebsd: include libkvm in build image

wg:
Cross builds targeting FreeBSD currently fail for projects depending on version 0.2.105+ of the libc crate due to a new dependency on libkvm, introduced here: https://github.com/rust-lang/libc/commit/9ab890d42b8435517369320ae6a259386049917a#diff-75cac48df6376b0d5783bc764c591d1a896a92429653beda3f0b3e93aa8c339fR290

The compilation error looks like:

```
error: linking with `x86_64-unknown-freebsd12-gcc` failed: exit status: 1
...
  = note: /usr/local/lib/gcc/x86_64-unknown-freebsd12/6.4.0/../../../../x86_64-unknown-freebsd12/bin/ld: cannot find -lkvm
          collect2: error: ld returned 1 exit status
```

This PR adds the static and shared libkvm binaries to the build image to resolve this issue.

PR #637, opened 2022-02-08
freebsd: Add libkvm.so

mirekdlugosz:
I have noticed that [Starship](https://github.com/starship/starship) [fails to create FreeBSD build](https://github.com/starship/starship/runs/5094984722?check_suite_focus=true). Command `cross build --release --locked --target x86_64-unknown-freebsd` fails with error:

```
/usr/local/lib/gcc/x86_64-unknown-freebsd12/6.4.0/../../../../x86_64-unknown-freebsd12/bin/ld: cannot find -lkvm
```

While trying to debug the issue, I have noticed that file `libkvm.so.7` exists in temporary directory [during container build](https://github.com/cross-rs/cross/blob/master/docker/freebsd.sh), but it's not copied over to destination directory and doesn't end up in final image.

This file is only 70 kB, so I thought we might just add it to image, which should allow all projects depending on libkvm to build FreeBSD binaries without any extra effort.

But if you don't want to add more libraries to container image, please let me know - I'll work with starship people on finding an approach that works for them.

### DIFF
--- a/docker/freebsd.sh
+++ b/docker/freebsd.sh
@@ -74,6 +74,7 @@ main() {
     cp "${td}/freebsd/lib/libthr.so.3" "${destdir}/lib"
     cp "${td}/freebsd/lib/libutil.so.9" "${destdir}/lib"
     cp "${td}/freebsd/lib/libssp.so.0" "${destdir}/lib"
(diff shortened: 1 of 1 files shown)


Emilgardis:
Thanks! But see https://github.com/cross-rs/cross/pull/613 which is included in #609

PR #654, opened 2022-03-16
switch to color-eyre for error reporting

Emilgardis:


src/config.rs
```diff
@@ -265,16 +265,10 @@ mod tests {
...
-                true
-            );
-            assert_eq!(
-                target.as_ref().unwrap().contains(&"PASS2".to_string()),
-                true
-            );
+            assert!(build.as_ref().unwrap().contains(&"TEST1".to_string()));
+            assert!(build.as_ref().unwrap().contains(&"TEST2".to_string()));
+            assert!(target.as_ref().unwrap().contains(&"PASS1".to_string()),);
+            assert!(target.as_ref().unwrap().contains(&"PASS2".to_string()),);
```
otavio, line 271: I fixed those in another PR #655.

otavio (changes requested):

Emilgardis:
can't think of any more suggestion notes to add right now, more can be added later

Cargo.toml
```diff
@@ -12,7 +12,8 @@ edition = "2018"
 
 [dependencies]
 atty = "0.2"
-error-chain = "0.12"
+color-eyre = "0.6.1"
+eyre = "0.6.7"
```
otavio, line 16:
```suggestion
color-eyre = "0.6"
eyre = "0.6"
```

otavio (changes requested):
Last minor thing ;-)

Cargo.toml
Emilgardis: I don't like being non-specific about minimal versions, but we're a bin target and people should install with `--locked`, so I guess this is fine

reitermarkus (approved):

Cargo.toml
```diff
@@ -12,22 +12,22 @@ edition = "2018"
...
 toml = "0.5"
 which = { version = "4", default_features = false }
-shell-escape = "0.1.4"
-serde_json = "1.0.48"
+shell-escape = "0.1"
+serde_json = "1"
 
 [target.'cfg(not(windows))'.dependencies]
 nix = "0.23"
+libc = "0.2.104"
```
otavio, line 27: Why not using `0.2` here as well?

Emilgardis: I think it's worth using it here, since libc has been `0.2` for a while. I don't have a strong preference though (except for what I previously stated ;) )

Emilgardis:
bors r=reitermarkus

Emilgardis:
bors r=reitermarkus

Cargo.toml
otavio: I think it is not a matter of taste but constancy. Or we use it for all or for none :-)

Emilgardis: alright fine, I'll cave in 😁

otavio (approved):

reitermarkus (approved):

Emilgardis:
> > hmm, bors shouldn't care about comments afaik
> 
> I think it's because you pushed new commits which dismissed my approval, and bors requires one approval.

ah, they were done exactly the same minute, but no dismissal message was shown

Emilgardis:
I also have this which I want to add when published: https://github.com/yaahc/color-eyre/pull/106

PR #656, opened 2022-03-17
remove extra slash

otavio:
This is a rebase of #544.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -194,7 +194,7 @@ pub fn run(
         .args(&["-v", "/cargo/bin"])
         .args(&[
             "-v",
-            &format!("{}:/{}:Z", mount_root.display(), mount_root.display()),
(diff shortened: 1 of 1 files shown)


reitermarkus (approved):

Emilgardis:
[This](https://github.com/cross-rs/cross/pull/544#issuecomment-811796867) comment was not a concern then? I was imagining mount paths specified as relative paths would affect this

otavio:
I don't see why we'd need this. It seems valid to expect users to provide full paths. We can handle relative paths internally if need but seems too magical for me.

reitermarkus:
> I remember there being a specific reason for this. Ill try and dig it up unless someone else knows why.

I couldn't find any specific reason for having the slash there looking through `git blame`. If there is indeed a problem with relative paths we can canonicalize them beforehand, if that isn't already the case.

PR #663, opened 2022-03-17
change edition to 2021

Emilgardis:
also removes author field as it is not displayed on crates.io anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Change rust edition to 2021 and bump MSRV for the cross binary to 1.58.1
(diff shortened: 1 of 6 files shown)


otavio (commented):
I did the same here but didn't send it as it doesn't use any 2021 specific feature or addition. A side effect is raising the MSRV. Is it a problem?

Emilgardis:
> I did the same here but didn't send it as it doesn't use any 2021 specific feature or addition. A side effect is raising the MSRV. Is it a problem?

msrv is already 1.58, 2021 edition is 1.56.0

otavio (approved):

Emilgardis:
with color-eyre added, I want to transition all possible instances of `format!("...{}...", thing)` to `format!("...{thing}...")`

reitermarkus (approved):

otavio (approved):

PR #665, opened 2022-03-17
make env.volumes only change package path when used

Emilgardis:
This fixes windows usage.

It also tries to fix windows builds when using `env.volumes` by making windows compatible
paths for `env.volumes`

not using env.volumes will now also mount the project in `/project` again

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -89,7 +89,17 @@ pub fn run(
     let cargo_dir = mount_finder.find_mount_path(cargo_dir);
     let xargo_dir = mount_finder.find_mount_path(xargo_dir);
     let target_dir = mount_finder.find_mount_path(target_dir);
-    let mount_root = mount_finder.find_mount_path(root);
+    let host_root = mount_finder.find_mount_path(root);
+    let mount_root: PathBuf;
+    #[cfg(target_os = "windows")]
+    {
+        // On Windows, we can not mount the directory name directly. Instead, we use wslpath to convert the path to a linux compatible path.
+        mount_root = wslpath(&host_root, verbose)?;
+    }
+    #[cfg(not(target_os = "windows"))]
+    {
+        mount_root = host_root.clone();
+    }
     let sysroot = mount_finder.find_mount_path(sysroot);
 
     let mut cmd = if uses_xargo {
@@ -123,18 +133,35 @@ pub fn run(
         // flag forwards the value from the parent shell
         docker.args(&["-e", var]);
     }
-
+    let mut env_volumes = false;
     for ref var in config.env_volumes(target)? {
         validate_env_var(var)?;
 
         if let Ok(val) = env::var(var) {
-            let host_path = Path::new(&val).canonicalize()?;
-            let mount_path = &host_path;
+            let host_path: PathBuf;
+            let mount_path: PathBuf;
+
+            #[cfg(target_os = "windows")]
+            {
+                // Docker does not support UNC paths, this will try to not use UNC paths
+                host_path = dunce::canonicalize(&val)
+                    .wrap_err_with(|| format!("when canonicalizing path `{val}`"))?;
+                // On Windows, we can not mount the directory name directly. Instead, we use wslpath to convert the path to a linux compatible path.
+                mount_path = wslpath(&host_path, verbose)?;
+            }
+            #[cfg(not(target_os = "windows"))]
+            {
+                host_path = Path::new(&val)
+                    .canonicalize()
+                    .wrap_err_with(|| format!("when canonicalizing path `{val}`"))?;
+                mount_path = host_path.clone();
+            }
             docker.args(&[
                 "-v",
                 &format!("{}:{}", host_path.display(), mount_path.display()),
             ]);
             docker.args(&["-e", &format!("{}={}", var, mount_path.display())]);
+            env_volumes = true;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/docker.rs
```diff
@@ -236,6 +257,31 @@ pub fn image(config: &Config, target: &Target) -> Result<String> {
...
+        .arg("-a")
+        .arg(path)
+        .run_and_get_stdout(verbose)
+        .wrap_err_with(|| {
+            format!(
+                "could not get linux compatible path for `{}`",
+                path.display()
+            )
+        })
+        .suggestion("is WSL installed on the host?")
```
Emilgardis, line 281: redundant

src/docker.rs
```diff
@@ -236,6 +257,31 @@ pub fn image(config: &Config, target: &Target) -> Result<String> {
...
 }
 
+#[cfg(target_os = "windows")]
+fn wslpath(path: &Path, verbose: bool) -> Result<PathBuf> {
+    let wslpath = if let Ok(wslpath) = which::which("wsl.exe") {
+        wslpath
+    } else {
+        return Err(eyre::eyre!("could not find wsl.exe")
+            .warning("usage of `env.volumes` requires WSL on Windows"));
+    };
```
Emilgardis, line 267: can be combined into one statement

src/docker.rs
```diff
@@ -186,14 +197,24 @@ pub fn run(
...
+    } else {
+        docker.args(&["-v", &format!("{}:/project:Z", mount_root.display())]);
+    }
+    docker
         .args(&["-v", &format!("{}:/rust:Z,ro", sysroot.display())])
-        .args(&["-v", &format!("{}:/target:Z", target_dir.display())])
-        .args(&["-w", &mount_root.display().to_string()]);
+        .args(&["-v", &format!("{}:/target:Z", target_dir.display())]);
+
+    if env_volumes {
```
Emilgardis, line 213: doing multiple ifs to preserve order

src/docker.rs
```diff
@@ -186,14 +197,24 @@ pub fn run(
         .args(&["-v", &format!("{}:/xargo:Z", xargo_dir.display())])
         .args(&["-v", &format!("{}:/cargo:Z", cargo_dir.display())])
         // Prevent `bin` from being mounted inside the Docker container.
-        .args(&["-v", "/cargo/bin"])
-        .args(&[
+        .args(&["-v", "/cargo/bin"]);
+    if env_volumes {
+        docker.args(&[
             "-v",
             &format!("{}:{}:Z", mount_root.display(), mount_root.display()),
```
Emilgardis, line 204: this needs a host_root and mount_root as well

Emilgardis:
I have not tested this with env.volumes yet

Emilgardis:
bors try --target x86

Emilgardis:
This should be feature complete now 🎉

src/docker.rs
```diff
@@ -233,6 +270,28 @@ pub fn image(config: &Config, target: &Target) -> Result<String> {
...
+        .arg(path)
+        .run_and_get_stdout(verbose)
+        .wrap_err_with(|| {
+            format!(
+                "could not get linux compatible path for `{}`",
+                path.display()
+            )
+        })
+        .map(|s| s.trim().into())
+}
```
Emilgardis, line 293: it would be possible to fallback to a manual implementation of "normalizing". the logic is quite simple but I don't feel like it's necessary to do.

Emilgardis: I take that back, I tried to do it correctly but handling this is quite annoying :)

Emilgardis: maybe

```rust
fn wslpath2(path: impl AsRef<Path>) -> Result<String> {
    fn get_path_prefix_drive(path: &Path) -> Option<&str> {
        match path.components().next().unwrap() {
            std::path::Component::Prefix(prefix_component) => match prefix_component.kind() {
                std::path::Prefix::VerbatimDisk(_) => Some(
                    prefix_component
                        .as_os_str()
                        .to_str()
                        .expect("windows drive letters should be ascii A-Z")
                        .strip_prefix(r"\\?\")?
                        .strip_suffix(":")?,
                ),
                std::path::Prefix::Disk(_) => Some(
                    prefix_component
                        .as_os_str()
                        .to_str()
                        .expect("windows drive letters should be ascii A-Z")
                        .strip_suffix(":")?,
                ),
                _ => None,
            },
            _ => None,
        }
    }
    let path = path.as_ref();
    let components = path.components().skip(2);
    let mut path_c = String::from("/mnt/");

    path_c
        .push_str(&get_path_prefix_drive(path).ok_or_else(|| eyre::eyre!("no drive letter found"))?.to_lowercase());
    let mut sep = true;
    for comp in components {
        if sep {
            path_c.push('/');
        } else {
            sep = true;
        }
        match comp {
            std::path::Component::Normal(p) => path_c.push_str(
                p.to_str()
                    .ok_or_else(|| eyre::eyre!("path needs to be utf8"))?,
            ),
            std::path::Component::ParentDir => path_c.push_str(".."),
            std::path::Component::CurDir => path_c.push('.'),
            _ => sep = false,
        }
    }
    Ok(path_c)
}
```

Emilgardis:
has anyone been able to test this and give this a r+ @cross-rs/maintainers <3

Emilgardis:
ping @cross-rs/maintainers 

It would be good to have this included for 0.3.0

reitermarkus:
I'm not using Windows so I cannot test this, but it looks good so far.

> not using env.volumes will now also mount the project in `/project` again

Does this have any advantage or is this needed to fix Windows support?

reitermarkus (approved):

Emilgardis:
> > not using env.volumes will now also mount the project in `/project` again
> 
> Does this have any advantage or is this needed to fix Windows support?

No advantage, except for not leaking paths in the general case, like what was done before.

bors r=reitermarkus

PR #667, opened 2022-03-18
Add dockerfile for cross itself

ByteDream:
This adds a dockerfile which has cross and docker already installed and can be directly used to cross compile and test. It mostly adopts a comment in [#260](https://github.com/cross-rs/cross/issues/260#issuecomment-678829526). 

It uses docker in docker (#387) so to work correctly it must be run with the docker socket mounted as volume.
```
$ docker run -v /var/run/docker.sock:/var/run/docker.sock cross
root@1e27ac2ed941:/# git clone https://github.com/cross-rs/cross
root@1e27ac2ed941:/# cd cross
root@1e27ac2ed941:/cross# cross build --target aarch64-unknown-linux-gnu
```

Although the [README already describes](https://github.com/cross-rs/cross#docker-in-docker) how to achieve this, a provided Dockerfile that can be downloaded from Docker Hub in the best case would be much easier than writing your own. Especially when using a CI pipeline inside Docker, where creating a custom Dockerfile is not that easy, this could be a big help.
The directory of the Dockerfile may not be the best choice since only compile targets are in this directory but it was the best fit.


docker/Dockerfile.cross
```diff
@@ -0,0 +1,7 @@
+FROM rust:latest
+
+ENV CROSS_DOCKER_IN_DOCKER=true
+
+RUN wget -O- https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz | tar -C /usr/bin/ --strip-components=1 -xzf - docker/docker
```
otavio, line 5: We need to check signature or checksum here so we don't rely on the file being unmodified in future.

otavio (changes requested):

docker/Dockerfile.cross
Emilgardis: I don't think this is necessary.

otavio: It is not rare for tarball files to change behind the scene, so ensuring it is valid is a good security matter.

ByteDream:
The checksum check is now implemented (md5) and it only checks the `docker` binary inside of the tarball since this is the only file which is needed.

otavio (dismissed):

docker/Dockerfile.cross
```diff
@@ -0,0 +1,10 @@
+FROM rust:latest
+
+ENV CROSS_DOCKER_IN_DOCKER=true
+
+ARG checksum="d9d9b964d428733e98c2579b91772af4  /usr/bin/docker"
+
+RUN wget -O- https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz | tar -C /usr/bin/ --strip-components=1 -xzf - docker/docker
+RUN echo $checksum | md5sum -c
+
+RUN cargo install cross
```
reitermarkus, line 10: This should install the current version in this repository.

docker/Dockerfile.cross
reitermarkus: Is there a Docker image we can copy the binary from instead using a multi-stage build?

ByteDream: There is a [docker in docker image](https://hub.docker.com/_/docker,) but it has some other binaries which are not necessary and rust has to be installed manually which takes more steps than this multi-stage build.
The docker in docker image receives it binaries from `download.docker.com` ([layer 6](https://hub.docker.com/layers/docker/library/docker/latest/images/sha256-3e1787157b0cd05a6c56f1d7bc916aa486f53f40f0d1538e34b917bdf947f94d)) just like here, so there is no big difference when using the docker in docker image (besides the unwanted binaries which it has).

docker/Dockerfile.cross
```diff
@@ -7,4 +7,4 @@ ARG checksum="d9d9b964d428733e98c2579b91772af4  /usr/bin/docker"
 RUN wget -O- https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz | tar -C /usr/bin/ --strip-components=1 -xzf - docker/docker
 RUN echo $checksum | md5sum -c
 
-RUN cargo install cross
+RUN cargo install --git https://github.com/cross-rs/cross
```
Emilgardis, line 10: This is still wrong imo, I'd expect this to be built in our CI and published as `ghcr.io/cross-rs/cross-in-docker`.

Emilgardis (changes requested):

docker/Dockerfile.cross
reitermarkus: Yes, this needs to be something like

```
COPY . .
RUN cargo install --path . --locked
```

otavio:
@ByteDream could you rebase and add a changelog entry? So we could trigger CI and see if it'll succeed.

docker/Dockerfile.cross-in-docker
```diff
@@ -0,0 +1,10 @@
+FROM rust:latest
+
+ENV CROSS_DOCKER_IN_DOCKER=true
+
+ARG checksum="d9d9b964d428733e98c2579b91772af4  /usr/bin/docker"
+
+RUN wget -O- https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz | tar -C /usr/bin/ --strip-components=1 -xzf - docker/docker
+RUN echo $checksum | md5sum -c
+
+RUN cargo install --git https://github.com/cross-rs/cross
```
Emilgardis, line 10: This is still wrong, it should copy from the checked out code, if this is kept the CI will work, but it'll check the most recent release, not the proposed changes

Emilgardis (changes requested):

docker/Dockerfile.cross-in-docker
otavio: Sure, it should use the checkout for building, thanks @Emilgardis for raising this up.

.github/workflows/ci.yml
```diff
@@ -170,6 +170,8 @@ jobs:
             - { target: thumbv7em-none-eabi,              os: ubuntu-latest,                    std: 1 }
             - { target: thumbv7em-none-eabihf,            os: ubuntu-latest,                    std: 1 }
             - { target: thumbv7m-none-eabi,               os: ubuntu-latest,                    std: 1 }
+            
+            - { target: cross                             os: ubuntu-latest }
```
Emilgardis, line 174: This won't work either sadly, it'll have to be it's own task/action for the image to be built and published

Emilgardis (changes requested):
The image needs to be added

Alexhuszagh:
Closed since #878 as merged.

PR #668, opened 2022-03-19
cleanup `action.yml`s

Emilgardis:


Emilgardis:
i don't understand why the changelog check is not respecting the label

https://github.com/cross-rs/cross/runs/5611031251?check_suite_focus=true#step:3:4

reitermarkus:
> i don't understand why the changelog check is not respecting the label

Because you added the label after the check ran. I actually created an action specifically for this use case, which reruns a workflow when specific labels are applied/removed: https://github.com/reitermarkus/rerun-workflow

Also, the workflow itself must get the label from the API instead of from the event, since at the time of the event, the label wasn't there, even on a rerun.

Emilgardis:
I added the label at the same time as creating the pr from the web interface. You'd think github would handle that scenario well for actions

Issue #664, opened 2022-03-17
Create documentation for Cross.toml


Emilgardis:
### Discussed in https://github.com/cross-rs/cross/discussions/532

We should add some kind of specification, this should be helped with #657

Alexhuszagh:
`Cross.toml` is now extensively documented on the [wiki](https://github.com/cross-rs/cross/wiki/Configuration#config-file), which includes a link to a sample configuration [file](https://github.com/cross-rs/wiki_assets/blob/main/Configuration/Cross.toml) where every variable is explained in depth.

Issue #657, opened 2022-03-17
allow `Cargo.toml` pkg metadata as config source


Emilgardis:
Ideally I would like to get rid of `Cross.toml` eventually and use `package.metadata.cross` in `Cargo.toml` to store `cross` configuration.

_Originally posted by @reitermarkus in https://github.com/cross-rs/cross/issues/624#issuecomment-1065815961_

Emilgardis:
This would be a good opportunity to switch to a serde deserialize implementation of getting configs

reitermarkus:
To get the config, we can use https://docs.rs/cargo/0.60.0/cargo/util/config/struct.Config.html.

We'll just need to parse some more arguments to get all information for calling https://docs.rs/cargo/0.60.0/cargo/util/config/struct.Config.html#method.configure.

Emilgardis:
the cargo lib is quite heavy, not sure it's worth it

mntns:
I just submitted a PR for using `serde` to deserialize the `Cross.toml` and would be happy to integrate the `Cargo.toml` as config source if it gets merged.

Emilgardis:
Should it be valid to have configuration in both `Cross.toml` and `Cargo.toml`? I think not.

If we do allow it, should we panic when a value exists twice?

mntns:
I just submitted a PR for that feature. For now, a warning will be shown if there are a `Cross.toml` and configuration values in the `Cargo.toml` at the same time, making the two configuration sources mutually exclusive.

Should the `Cargo.toml` method be the primary form of configuration? That could then be reflected more explicitly in the documentation.

Alexhuszagh:
> Should the `Cargo.toml` method be the primary form of configuration? That could then be reflected more explicitly in the documentation.

I think so, our goal was to phase out `Cross.toml` I believe (which wouldn't be anytime soon), so first having `Cargo.toml` be the primary form of configiration, then warning about `Cross.toml`if it exists a few releases later would be good.

PR #670, opened 2022-03-22
Implement `Cross.toml` deserialization using `serde`

mntns:
This PR implements the deserialization of the `Cross.toml` config using `serde`. It can therefore be seen as some preliminary work on the issues #657, #664, and is related to #532 on #624.

I added some basic documentation as module-level documentation, although I'm not sure whether this is the best place for that. Ideally it should be documentated in the `README.md` at some point. Furthermore, `CrossToml` lives in its own module, but a more suitable place might be the `config` module.

### DIFF
--- a/src/cross_toml.rs
+++ b/src/cross_toml.rs
@@ -102,6 +102,7 @@ impl CrossToml {
     }
 }
 
+#[cfg(test)]
 mod tests {
     use super::*;
 
--- a/src/cross_toml.rs
+++ b/src/cross_toml.rs
@@ -1,38 +1,4 @@
-//! Implements the parsing of `Cross.toml`
-//!
-//! `Cross.toml` can contain the following elements:
-//!
-//! # `build`
-//! The `build` key allows you to set global variables, e.g.:
-//!
-//! ```toml
-//! [build]
-//! xargo = true
-//! ```
-//!
-//! # `build.env`
-//! With the `build.env` key you can globally set volumes that should be mounted
-//! in the Docker container or environment variables that should be passed through.
-//! For example:
-//!
-//! ```toml
-//! [build.env]
-//! volumes = ["vol1", "vol2"]
-//! passthrough = ["IMPORTANT_ENV_VARIABLES"]
-//! ```
-//!
-//! # `target.TARGET`
-//! The `target` key allows you to specify parameters for specific compilation targets.
-//!
-//! ```toml
-//! [target.aarch64-unknown-linux-gnu]
-//! volumes = ["vol1", "vol2"]
-//! passthrough = ["VAR1", "VAR2"]
-//! xargo = false
-//! image = "test-image"
-//! runner = "custom-runner"
-//! ```
-//! 
+#![doc = include_str!("../docs/cross_toml.md")]
 
 use std::collections::HashMap;
 use serde::Deserialize;
@@ -165,7 +131,7 @@ mod tests {
             targets: None,
             build: Some(CrossBuildConfig {
                 env: Some(CrossBuildEnvConfig {
-                    volumes: Some(vec!["vol1".to_string(), "vol2".to_string()]),
+                    volumes: Some(vec!["VOL1_ARG".to_string(), "VOL2_ARG".to_string()]),
                     passthrough: Some(vec!["VAR1".to_string(), "VAR2".to_string()])
                 }),
                 xargo: Some(true),
@@ -178,7 +144,7 @@ mod tests {
           xargo = true
  
           [build.env]
-          volumes = ["vol1", "vol2"]
+          volumes = ["VOL1_ARG", "VOL2_ARG"]
           passthrough = ["VAR1", "VAR2"]
         "#;
         let parsed_cfg = CrossToml::from_str(test_str)?;
@@ -195,7 +161,7 @@ mod tests {
             Target::BuiltIn { triple: "aarch64-unknown-linux-gnu".to_string() },
             CrossTargetConfig {
                 passthrough: Some(vec!["VAR1".to_string(), "VAR2".to_string()]),
-                volumes: Some(vec!["vol1".to_string(), "vol2".to_string()]),
+                volumes: Some(vec!["VOL1_ARG".to_string(), "VOL2_ARG".to_string()]),
                 xargo: Some(false),
                 image: Some("test-image".to_string()),
                 runner: None,
@@ -209,7 +175,7 @@ mod tests {
 
         let test_str = r#"
           [target.aarch64-unknown-linux-gnu]
-          volumes = ["vol1", "vol2"]
+          volumes = ["VOL1_ARG", "VOL2_ARG"]
           passthrough = ["VAR1", "VAR2"]
           xargo = false
           image = "test-image"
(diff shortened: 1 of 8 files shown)


src/cross_toml.rs
```diff
@@ -0,0 +1,222 @@
...
+//! ```
+//!
+//! # `build.env`
+//! With the `build.env` key you can globally set volumes that should be mounted
+//! in the Docker container or environment variables that should be passed through.
+//! For example:
+//!
+//! ```toml
+//! [build.env]
+//! volumes = ["vol1", "vol2"]
```
Emilgardis, line 20: this should be upper case, it's also environment variables

```suggestion
//! volumes = ["VOL1_ARG", "VOL1_ARG"]
```

src/cross_toml.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct CrossTargetConfig {
+    passthrough: Option<Vec<String>>,
+    volumes: Option<Vec<String>>,
+    xargo: Option<bool>,
+    image: Option<String>,
+    runner: Option<String>,
+}
+
+/// Wrapper struct for `Target` -> `CrossTargetConfig` mappings
```
Emilgardis, line 67: Please try to use intra-doc links :D

Emilgardis:
```suggestion
/// Wrapper struct for [`Target`] -> [`CrossTargetConfig`] mappings
```

src/main.rs
```diff
@@ -116,7 +119,8 @@ impl<'a> From<&'a str> for Host {
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
+#[serde(from = "&str")]
```
Emilgardis, line 123: why from `&str` and not `String` ?

src/main.rs
```diff
@@ -87,6 +89,7 @@ impl Host {
         }
     }
 
+    /// Returns the `Target` as target triple string
```
Emilgardis, line 92:
```suggestion
    /// Returns the [`Target`] as target triple string
```

Emilgardis (changes requested):
looks great!

Emilgardis:
> I added some basic documentation as module-level documentation, although I'm not sure whether this is the best place for that. Ideally it should be documentated in the `README.md` at some point.

We can split the documentation out into it's own .md and then do

```rust
#![doc = include_str!("file.md")]
```

src/main.rs
mntns: Since `Host` already implements `From<&str>` and `Target` implements `From<Host>`, I added the transitive `From<&str> for Target`, which is then used by `serde` to deserialize the keys of the map [here](https://github.com/mntns/cross/blob/ea276c1f653b210ebee76591aac3a00b40966874/src/cross_toml.rs#L40). I think there is no big difference between using `String` or a string slice for the deserialization.

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct CrossToml {
+    #[serde(rename = "target")]
+    targets: Option<CrossTargetMapConfig>,
+    build: Option<CrossBuildConfig>,
+}
+
+impl CrossToml {
+    /// Parses the [`CrossToml`] from a string
+    pub fn from_str(toml_str: &str) -> Result<Self> {
+        let cfg: CrossToml = toml::from_str(toml_str)?;
```
Emilgardis, line 54: i think it would make sense to add [`serde-path-to-error`](https://github.com/dtolnay/path-to-error), but that can be done in another pr

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+    passthrough: Option<Vec<String>>,
+    volumes: Option<Vec<String>>,
+    xargo: Option<bool>,
+    image: Option<String>,
+    runner: Option<String>,
+}
+
+/// Wrapper struct for mappings from [`Target`] to [`CrossTargetConfig`]
+///
+/// This is used to circumvent that serde's flatten and field aliases
```
Emilgardis, line 35: this can be a `// TODO:` instead of docstring

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct CrossTargetMapConfig {
+    #[serde(flatten)]
+    inner: HashMap<Target, CrossTargetConfig>,
+}
+
+/// Cross configuration
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct CrossToml {
+    #[serde(rename = "target")]
+    targets: Option<CrossTargetMapConfig>,
```
Emilgardis, line 47: this doesn't need to be an option, just make `CrossTargetMapConfig` default, and deref to `inner` and then finally

```rust
    #[serde(default)]
    targets: CrossTargetMapConfig,
```

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+    build: Option<CrossBuildConfig>,
```
Emilgardis, line 48: Same here, doesn't need to be option

src/cross_toml.rs
Emilgardis: also, is this really applicable here?

```rust
pub struct CrossToml {
    #[serde(rename = "target")]
    targets: HashMap<Target, CrossTargetConfig>
    build: Option<CrossBuildConfig>,
}
```

would work no?

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+    env: Option<CrossBuildEnvConfig>,
+    xargo: Option<bool>,
+    target: Option<String>,
+}
+
+/// Target configuration
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct CrossTargetConfig {
+    passthrough: Option<Vec<String>>,
+    volumes: Option<Vec<String>>,
```
Emilgardis, line 27: this doesn't need to be Option, you can default it instead

src/cross_toml.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::errors::*;
+use crate::Target;
+use serde::Deserialize;
+use std::collections::HashMap;
+
+/// Build environment configuration
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct CrossBuildEnvConfig {
+    volumes: Option<Vec<String>>,
+    passthrough: Option<Vec<String>>,
```
Emilgardis, line 12: this doesn't need to be Option, you can default it instead

Emilgardis (changes requested):
some minor nits, not really too major

src/cross_toml.rs
mntns: Good one!

src/cross_toml.rs
mntns: Oh, true, I don't know why I thought that the wrapper struct would be necessary.

src/cross_toml.rs
mntns: Do you mean the `build` field here? For this I think it makes sense to have the `Option`. For the `targets` field I added the defaulting.

Emilgardis: my bad, yes I mean the build as well, but if you think it makes sense I'll agree with you :)

Emilgardis (approved):
thanks! this is great 😄 thank you!

I want to merge #624 as well, but I think it makes sense to merge this before

please address the remaining clippy issues and add a simple changelog entry and I'll r+ this!

Emilgardis (approved):
bors r+

PR #673, opened 2022-03-24
warn when used toolchain is out-of-date

Emilgardis:
on non-linux hosts, warn when toolchain is out of date.

# Example

```
$ cross build
Warning: using older rustc `1.61.0-nightly (9f4dc0b4d 2022-03-23)` for the target. Current active rustc on the host is `rustc 1.62.0-nightly (879aff385 2022-04-20)`.
 > Update with `rustup update --force-non-host nightly-x86_64-unknown-linux-gnu`
```

### DIFF
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -1,4 +1,7 @@
+use std::path::Path;
 use std::process::Command;
+
+use rustc_version::Version;
 
 use crate::errors::*;
 use crate::extensions::CommandExt;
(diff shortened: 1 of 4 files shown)


src/rustup.rs
```diff
@@ -97,3 +100,32 @@ pub fn component_is_installed(component: &str, toolchain: &str, verbose: bool) -
...
+            .get("pkg")
+            .and_then(|pkg| pkg.get("rust"))
+            .and_then(|rust| rust.get("version"))
+            .and_then(|version| version.as_str())
+        {
+            let mut i = rust_version.splitn(2, ' ');
+            Ok(Some((
+                Version::parse(
+                    i.next().ok_or_else(|| {
+                        eyre::eyre!("no rust version found in {}", path.display())
```
Emilgardis, line 120: does this warrant an error?

src/rustup.rs
```diff
@@ -97,3 +100,32 @@ pub fn component_is_installed(component: &str, toolchain: &str, verbose: bool) -
...
+                    })?,
+                )?,
+                i.next().map(|s| s.to_owned()),
+            )))
+        } else {
+            Ok(None)
+        }
+    } else {
+        Ok(None)
+    }
```
reitermarkus, line 130: These `else`s can be removed if you `return` the `Ok` variant.

src/rustup.rs
reitermarkus: I think the error here should be that the version is invalid rather that not being found.

src/rustup.rs
```diff
@@ -97,3 +100,32 @@ pub fn component_is_installed(component: &str, toolchain: &str, verbose: bool) -
...
+        let contents = std::fs::read(&path)
+            .wrap_err_with(|| format!("couldn't open file `{}`", path.display()))?;
+        let manifest: toml::value::Table = toml::from_slice(&contents)?;
+        if let Some(rust_version) = manifest
+            .get("pkg")
+            .and_then(|pkg| pkg.get("rust"))
+            .and_then(|rust| rust.get("version"))
+            .and_then(|version| version.as_str())
+        {
+            let mut i = rust_version.splitn(2, ' ');
```
reitermarkus, line 116: Use `split_once` here.

Emilgardis:
Done, I think :)

Emilgardis:
oh ic, there's a mistake with `use_codeowners = true`

Emilgardis:
bors r+

maybe it's because of protected branch, lets see

Emilgardis:
> 👎 Rejected because of missing code owner approval

seriously?...

Emilgardis:
bors r+ 

🤞

PR #674, opened 2022-03-25
Specify `edge` for the example of Dockerfile

yykamei:
`README.md` recommends writing `Dockerfile` based on the default
Docker image that cross uses. The example indicates the base
image's tag as `latest`, but `latest` does not seem to exist on container
packages.

https://github.com/orgs/cross-rs/packages?repo_name=cross

So, I updated `README.md` to specify `edge` instead of `latest`
in order for developers to use the snippet just by copying it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ This way you won't have to figure out how to install a cross C toolchain in your
 custom image. Example below:
 
 ``` Dockerfile
-FROM ghcr.io/cross-rs/aarch64-unknown-linux-gnu:latest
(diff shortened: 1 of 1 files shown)


Emilgardis:
This is not something we want. We do not have a latest tag currently due to not having had a new release since 0.2.1 on the new CI. This will be solved when s new release (0.3.0) is done.

Unfortunately, this means there is some desync between the readme and the actual state of things, but it will be solved soon™

Thanks anyway!

yykamei:
Thank you! I'm looking forward to the new release 👍

PR #678, opened 2022-04-03
allow specifying a dockerfile to use for a target and implement "pre-build"ing

### LINKED ISSUE #565: Add pre-build hook

Emilgardis:
Add optional `target.{target}.dockerfile[.file]`, `target.{target}.dockerfile.context` and `target.{target}.dockerfile.build-args` to invoke `docker/podman build` before using an image.

also adds `target.{target}.pre-build` to allow pre-building as per #565

In total, this PR allows a user to do the following:

```toml
[build]
pre-build = [
    "dpkg --add-architecture $CROSS_DEB_ARCH", 
    "apt-get update && apt-get --assume-yes install libssl-dev:$CROSS_DEB_ARCH"
    ]
dockerfile = { file = "./Dockerfile.x86_64_ssl", context = "..", build-args = { ARG1 = "foo" }}
image = "my-image:latest"
pre-build = ["echo 'Hello world!'"]
```

fixes #565

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #795 - added images for additional toolchains maintained by cross-rs.
 - #792 - added `CROSS_CONTAINER_IN_CONTAINER` environment variable to replace `CROSS_DOCKER_IN_DOCKER`.
 - #782 - added `build-std` config option, which builds the rust standard library from source if enabled.
-- #775 - forward Cargo exit code to host
+- #678 - Add optional `target.{target}.dockerfile[.file]`, `target.{target}.dockerfile.context` and `target.{target}.dockerfile.build-args` to invoke docker/podman build before using an image.
+- #678 - Add `target.{target}.pre-build` config for running commands before building the image.
 - #772 - added `CROSS_CONTAINER_OPTS` environment variable to replace `DOCKER_OPTS`.
 - #767, #788 - added the `cross-util` and `xtask` commands.
 - #745 - added `thumbv7neon-*` targets.
@@ -27,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- #775 - forward Cargo exit code to host
 - #762 - re-enabled `x86_64-unknown-dragonfly` target.
 - #747 - reduced android image sizes.
 - #746 - limit image permissions for android images.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 serde_ignored = "0.1.2"
 shell-words = "1.1.0"
+sha1_smol = "1.0.0"
 
 [target.'cfg(not(windows))'.dependencies]
 nix = { version = "0.24", default-features = false, features = ["user"] }
--- a/README.md
+++ b/README.md
@@ -112,10 +112,22 @@ the default one. Normal Docker behavior applies, so:
 
 - If only `tag` is omitted, then Docker will use the `latest` tag.
 
+#### Dockerfiles
+
+If you're using a custom Dockerfile, you can use `target.{{TARGET}}.dockerfile` to automatically build it
+
+``` toml
+[target.aarch64-unknown-linux-gnu.dockerfile]
+dockerfile = "./path/to/where/the/Dockerfile/resides"
+```
+
+`cross` will build and use the image that was built instead of the default image.
+
 It's recommended to base your custom image on the default Docker image that
 cross uses: `ghcr.io/cross-rs/{{TARGET}}:{{VERSION}}` (where `{{VERSION}}` is cross's version).
 This way you won't have to figure out how to install a cross C toolchain in your
-custom image. Example below:
+custom image.
+
 
 ``` Dockerfile
 FROM ghcr.io/cross-rs/aarch64-unknown-linux-gnu:latest
@@ -125,8 +137,23 @@ RUN dpkg --add-architecture arm64 && \
     apt-get install --assume-yes libfoo:arm64
 ```
 
+If you want cross to provide the `FROM` instruction, you can do the following
+
+``` Dockerfile
+ARG CROSS_BASE_IMAGE
+FROM $CROSS_BASE_IMAGE
+
+RUN ...
 ```
-$ docker build -t my/image:tag path/to/where/the/Dockerfile/resides
+
+#### Pre-build hook
+
+`cross` enables you to add dependencies and run other necessary commands in the image before using it.
+This action will be added to the used image, so it won't be ran/built every time you use `cross`.
+
+``` toml
+[target.x86_64-unknown-linux-gnu]
+pre-build = ["dpkg --add-architecture arm64 && apt-get update && apt-get install --assume-yes libfoo:arm64"]
 ```
 
 ### Docker in Docker
--- a/docs/cross_toml.md
+++ b/docs/cross_toml.md
@@ -1,6 +1,7 @@
 The `cross` configuration in the `Cross.toml` file, can contain the following elements:
 
 # `build`
+
 The `build` key allows you to set global variables, e.g.:
 
 ```toml
@@ -11,6 +12,7 @@ default-target = "x86_64-unknown-linux-gnu"
 ```
 
 # `build.env`
+
 With the `build.env` key you can globally set volumes that should be mounted
 in the Docker container or environment variables that should be passed through.
 For example:
@@ -22,6 +24,7 @@ passthrough = ["IMPORTANT_ENV_VARIABLES"]
 ```
 
 # `target.TARGET`
+
 The `target` key allows you to specify parameters for specific compilation targets.
 
 ```toml
@@ -29,10 +32,12 @@ The `target` key allows you to specify parameters for specific compilation targe
 xargo = false
 build-std = false
 image = "test-image"
+pre-build = ["apt-get update"]
 runner = "custom-runner"
 ```
 
 # `target.TARGET.env`
+
 The `target` key allows you to specify environment variables that should be used for a specific compilation target.
 This is similar to `build.env`, but allows you to be more specific per target.
 
@@ -40,4 +45,20 @@ This is similar to `build.env`, but allows you to be more specific per target.
 [target.x86_64-unknown-linux-gnu.env]
 volumes = ["VOL1_ARG", "VOL2_ARG"]
 passthrough = ["IMPORTANT_ENV_VARIABLES"]
+```
+
+# `target.TARGET.dockerfile`
+
+```toml
+[target.x86_64-unknown-linux-gnu.dockerfile]
+file = "./Dockerfile" # The dockerfile to use relative to the `Cargo.toml`
+context = "." # What folder to run the build script in
+build-args = { ARG1 = "foo" } # https://docs.docker.com/engine/reference/builder/#arg
+```
+
+also supports
+
+```toml
+[target.x86_64-unknown-linux-gnu]
+dockerfile = "./Dockerfile"
 ```
--- a/src/config.rs
+++ b/src/config.rs
@@ -68,6 +68,23 @@ impl Environment {
         self.get_target_var(target, "IMAGE")
     }
 
+    fn dockerfile(&self, target: &Target) -> Option<String> {
+        let res = self.get_values_for("DOCKERFILE", target, |s| s.to_string());
+        res.0.or(res.1)
+    }
+
+    fn dockerfile_context(&self, target: &Target) -> Option<String> {
+        let res = self.get_values_for("DOCKERFILE_CONTEXT", target, |s| s.to_string());
+        res.0.or(res.1)
+    }
+
+    fn pre_build(&self, target: &Target) -> Option<Vec<String>> {
+        let res = self.get_values_for("PRE_BUILD", target, |v| {
+            v.split('\n').map(String::from).collect()
+        });
+        res.0.or(res.1)
+    }
+
     fn runner(&self, target: &Target) -> Option<String> {
         self.get_target_var(target, "RUNNER")
     }
@@ -178,16 +195,34 @@ impl Config {
     fn vec_from_config(
         &self,
         target: &Target,
-        env: impl Fn(&Environment, &Target) -> (Option<Vec<String>>, Option<Vec<String>>),
-        config_build: impl for<'a> Fn(&'a CrossToml) -> Option<&'a [String]>,
-        config_target: impl for<'a> Fn(&'a CrossToml, &Target) -> Option<&'a [String]>,
-    ) -> Result<Vec<String>> {
+        env: impl for<'a> Fn(&'a Environment, &Target) -> (Option<Vec<String>>, Option<Vec<String>>),
+        config: impl for<'a> Fn(&'a CrossToml, &Target) -> (Option<&'a [String]>, Option<&'a [String]>),
+        sum: bool,
+    ) -> Result<Option<Vec<String>>> {
         let (env_build, env_target) = env(&self.env, target);
 
-        let mut collected = self.sum_of_env_toml_values(env_build, config_build)?;
-        collected.extend(self.sum_of_env_toml_values(env_target, |t| config_target(t, target))?);
+        if sum {
+            return self.sum_of_env_toml_values(env_target, |t| config(t, target));
+        } else if let Some(env_target) = env_target {
+            return Ok(Some(env_target));
+        }
 
-        Ok(collected)
+        let (build, target) = self
+            .toml
+            .as_ref()
+            .map(|t| config(t, target))
+            .unwrap_or_default();
+
+        // FIXME: let expression
+        if target.is_none() && env_build.is_some() {
+            return Ok(env_build);
+        }
+
+        if target.is_none() {
+            Ok(build.map(ToOwned::to_owned))
+        } else {
+            Ok(target.map(ToOwned::to_owned))
+        }
     }
 
     #[cfg(test)]
@@ -211,22 +246,17 @@ impl Config {
         self.string_from_config(target, Environment::runner, CrossToml::runner)
     }
 
-    pub fn env_passthrough(&self, target: &Target) -> Result<Vec<String>> {
+    pub fn env_passthrough(&self, target: &Target) -> Result<Option<Vec<String>>> {
         self.vec_from_config(
             target,
             Environment::passthrough,
-            CrossToml::env_passthrough_build,
-            CrossToml::env_passthrough_target,
+            CrossToml::env_passthrough,
+            true,
         )
     }
 
-    pub fn env_volumes(&self, target: &Target) -> Result<Vec<String>> {
-        self.vec_from_config(
-            target,
-            Environment::volumes,
-            CrossToml::env_volumes_build,
-            CrossToml::env_volumes_target,
-        )
+    pub fn env_volumes(&self, target: &Target) -> Result<Option<Vec<String>>> {
+        self.vec_from_config(target, Environment::volumes, CrossToml::env_volumes, false)
     }
 
     pub fn target(&self, target_list: &TargetList) -> Option<Target> {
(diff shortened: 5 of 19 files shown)


docs/cross_toml.md
```diff
@@ -29,5 +29,7 @@ volumes = ["VOL1_ARG", "VOL2_ARG"]
 passthrough = ["VAR1", "VAR2"]
 xargo = false
 image = "test-image"
+dockerfile = "./Dockerfile"
+context = "."
```
Emilgardis, line 33: This should maybe be more explicit, or even added as its own table together with dockerfile

So

`dockerfile = {file = "./Dockerfile", context = "." }`

Could even make it take either a table or string

src/docker.rs
```diff
@@ -248,8 +250,51 @@ pub fn run(
         }
     }
 
+    let image = if let Some(dockerfile) = config.dockerfile(target)? {
+        let mut docker_build = docker_command("build")?;
```
Emilgardis, line 254: Should maybe add a flag to allow specifying more flags for this command

src/docker.rs
```diff
@@ -248,8 +250,51 @@ pub fn run(
...
+        } else {
+            let mut image_name = String::new();
+            std::fs::File::open(&iidfile_path)?.read_to_string(&mut image_name)?;
+            image_name
+        };
+        if image_name.is_empty() {
+            eyre::bail!("no image returned from docker build")
+        }
+        std::fs::remove_file(iidfile_path)?;
+        dbg!(image_name)
```
Emilgardis, line 291: Oops

Emilgardis:
I think this is a nice addition to #635

Emilgardis:
I've made this also implement #565

I think I personally like this approach better to #635

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #678 - Add optional target.{target}.dockerfile[.file], target.{target}.dockerfile.context and target.{target}.dockerfile.build-args to invoke docker/podman build before using an image.
+- #678 - Add `pre-build` config for running commands before building the image.
```
Emilgardis, line 9:
```suggestion
- #678 - Add optional `target.{target}.dockerfile[.file]`, `target.{target}.dockerfile.context` and `target.{target}.dockerfile.build-args` to invoke docker/podman build before using an image.
- #678 - Add `target.{target}.pre-build` config for running commands before building the image.
```

src/config.rs
```diff
@@ -152,6 +165,46 @@ impl Config {
...
+        self.toml.as_ref().map_or(Ok(None), |t| {
+            Ok(t.dockerfile_context(target).map(ToOwned::to_owned))
+        })
+    }
+
+    pub fn dockerfile_build_args(
+        &self,
+        target: &Target,
+    ) -> Result<Option<HashMap<String, String>>> {
+        // This value does not support env variables
```
Emilgardis, line 192: would it make sense to allow this? How?

src/cross_toml.rs
```diff
@@ -170,6 +263,8 @@ mod tests {
                 xargo: Some(false),
                 image: Some("test-image".to_string()),
                 runner: None,
+                dockerfile: None,
+                pre_build: None,
```
Emilgardis, line 267: should add tests for this

Emilgardis: including a ci run with them

Emilgardis: fixed with #683 if it lands

README.md
```diff
@@ -122,8 +135,14 @@ RUN dpkg --add-architecture arm64 && \
...
-$ docker build -t my/image:tag path/to/where/the/Dockerfile/resides
+#### Pre-build hook
+
+`cross` enables you to add dependencies and run other necessary commands in the image before using it.
+This action will be added to the used image, so it won't be run every time you use `cross`.
+
+``` toml
+[target.x86_64-unknown-linux-gnu]
+pre-build = ["yum update", "yum install -y libfoo"]
 ```
```
Emilgardis, line 146: Maybe this shouldn't be called `pre-build`.

My concern is that someone would actually like something like what #635 does, e.g run everytime, and this doesn't allow that.

However, that's kinda what `build.rs` is for

Emilgardis:
We can also add a label to our images to signal that an image is intended for cross, and that way we can return those with the tools in #767 

not sure what to label them though, `org.cross-rs.for-cross-target=<target>` ?
(i just bought that domain so we can use it :) )

src/cross_toml.rs
```diff
@@ -69,6 +95,32 @@ impl CrossToml {
         self.get_target(target).and_then(|t| t.image.clone())
     }
 
+    /// Returns the `target.{}.dockerfile` or `target.{}.dockerfile.file` part of `Cross.toml`
```
Alexhuszagh, line 98: This always returns `target.{}.dockerfile.file`, since this is from the config file only, and `file` is not optional now. Should be:

```rust
/// Returns the `target.{}.dockerfile.file` part of `Cross.toml`
```

src/config.rs
Alexhuszagh: I doubt it, since the build args could have characters that we'd expect in any delimiter we choose.

src/docker.rs
```diff
@@ -325,10 +328,143 @@ pub fn run(
...
+        docker_build.env("DOCKER_SCAN_SUGGEST", "false");
+        let image_name = self.image_name(target_triple, package);
+        docker_build.args(["--tag", &image_name]);
+
+        for (key, arg) in build_args.into_iter() {
+            docker_build.args(["--build-arg", &format!("{}={}", key.as_ref(), arg.as_ref())]);
+        }
+
+        match self {
+            Dockerfile::File { path, .. } => {
```
Alexhuszagh, line 411: Would it be preferable to handle the `FROM {{image}}` ourselves? That way the users can just add the components they want afterwards? Not sure about this but I think it might be easier.

Alexhuszagh (changes requested):

src/docker.rs
```diff
@@ -325,10 +328,143 @@ pub fn run(
...
+            package,
+        )
+        .wrap_err("when building dockerfile")?;
+    }
+
+    if let Some(pre_build) = config.pre_build(target)? {
+        let stdin = Dockerfile::Stdin {
+            content: format!(
+                "
+FROM {image}
```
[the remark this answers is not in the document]
Emilgardis: yep i've fixed that with adding pre-build to the tag when building in this "mode"

src/cross_toml.rs
Emilgardis: wrong link, https://github.com/cross-rs/cross/blob/0ecb3279fba1572cbbc52fd33a80603248334110/src/cross_toml.rs#L34-L35

src/docker.rs
```diff
@@ -325,10 +328,147 @@ pub fn run(
...
+        match self {
+            Dockerfile::File { path, .. } => {
+                docker_build.args(["--file", path]);
+                if let Some(context) = self.context() {
+                    docker_build.arg(&context);
+                }
+            }
+            Dockerfile::Stdin { content } => {
+                // XXX: Ugly hack to circumvent platform issues. we could use echo, but it may not exist
+                let stdin_path = tempfile::NamedTempFile::new()
```
Alexhuszagh, line 420: I think we can have a better version: just write to a Dockerfile in the `/target/${target}` directory, and then it will automatically be cleaned up by `cargo clean`, and we don't have to worry about making 100s of tempfiles.

src/docker.rs
Emilgardis: I fail to see the point, you could already do that with 

```toml
pre-build = ["apt-get update && apt-get install --assume-yes --no-install-recommends libclang-3.9-dev clang-3.9"]
```

I see another solution though, make a `target.<target>.dockerfile.provide-base` that would do it instead with the image you provide. But I still don't see the benefit

src/docker.rs
Emilgardis: yep, that's what I mentioned somewhere elsewhere, just slowly working through everything

src/docker.rs
Alexhuszagh: That's just a simple example: if something requires the `ENV` keyword or similar (like, say we want `vcpkg` for providing dependencies of cross-compiled C/C++ libraries), we can't simply use pre-build hooks.

Alexhuszagh: A more complex example would be:

```Dockerfile
    libclang-3.9-dev clang-3.9
RUN git clone https://github.com/Microsoft/vcpkg.git && ./vcpkg/bootstrap-vcpkg.sh
ENV PATH /path/to/vcpkg/bin/dir/:$PATH

# do more stuff here, maybe add an ARG, etc.
# everything is platform generic.
```

Alexhuszagh: Maybe we could have `pre-build` hooks encompass this functionality (`RUN` isn't implicit anymore, and then all this can be easily encompassed there)? That way you can provide a complete Dockerfile, or just add extra commands to the existing one?

Alexhuszagh:
> need #746 for its write function

I'll rebase it then after the musl PR merges and push. Does #746 look good to go?

Emilgardis:
i think so

src/docker.rs
Emilgardis: thinking like `DOCKER_OPTS`, I think a `DOCKER_BUILD_OPTS` is the best way to do it

Alexhuszagh: Ah ok, that's a good suggestion. Maybe also consider then #770, since we handle more than just docker now? If we're introducing new environment variables, we don't need to worry about backwards compatibility.

Cargo.toml
```diff
@@ -16,7 +16,7 @@ dev = ["serde_yaml"]
 [dependencies]
 atty = "0.2"
 clap = { version = "3.1.18", features = ["derive"] }
-color-eyre = "0.6"
+color-eyre = {version = "0.6", default-features = false }
```
Emilgardis, line 19: this is to disable location printing,

waiting on yaahc/color-eyre#106

Cargo.toml
```diff
@@ -28,6 +28,8 @@ serde_json = "1"
 serde_yaml = { version = "0.8", optional = true }
 serde_ignored = "0.1.2"
 shell-words = "1.1.0"
+tempfile = "3"
```
Emilgardis, line 31: this dep is temporary :), will use `target/<target>/` instead

src/cross_toml.rs
```diff
@@ -26,14 +27,41 @@ pub struct CrossBuildConfig {
...
 }
 
+/// Dockerfile configuration
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct CrossTargetDockerfileConfig {
+    file: String,
+    provide_base: bool,
+    context: Option<String>,
+    build_args: Option<HashMap<String, String>>,
```
Emilgardis, line 49: should this be a hashmap?

src/docker.rs
```diff
@@ -351,10 +351,168 @@ pub fn run(
         }
     }
 
+    let mut image = image(config, target)?;
+
+    // TODO: if -p, use that
+    let package = metadata.packages.get(0).unwrap();
```
Emilgardis, line 357: this needs to be fixed, it doesn't even seem to work

src/docker.rs
```diff
@@ -351,10 +351,168 @@ pub fn run(
...
+        };
+
+        if config.dockerfile_provide_base(target).unwrap_or_default() {
+            let mut content = format!("FROM {image}\n");
+            content.push_str(&crate::file::read(&path)?);
+            build = Dockerfile::Stdin {
+                content,
+                from_file: true,
+            };
+        }
```
Emilgardis, line 376: here's the thing I mentioned here https://github.com/cross-rs/cross/pull/678/files#r892810082

src/lib.rs
```diff
@@ -524,3 +524,15 @@ fn toml(metadata: &CargoMetadata) -> Result<Option<CrossToml>> {
...
+    use std::str::FromStr;
+
+    if let Ok(value) = bool::from_str(envvar) {
+        value
+    } else if let Ok(value) = i32::from_str(envvar) {
+        value != 0
+    } else {
+        !envvar.is_empty()
+    }
+}
```
Emilgardis, line 538: leftover from rebase

src/docker.rs
Emilgardis: I've implemented this (i think, not fully tested), but with allowing the user to tell cross to insert `FROM` with `target.TARGET.dockerfile.provide-base`.

Emilgardis:
This is ready for a general re-review, I have not tested everything yet, so maybe expect errors

Emilgardis:
Also waiting on https://github.com/cross-rs/cross/issues/770 to use shellwords for https://github.com/cross-rs/cross/pull/678#discussion_r893806642

Emilgardis:
if this pr is discarded (doubt it but you never know, parts of [ec19c9d8f3d0ed0c45a5d66720c15046898f35ea](https://github.com/cross-rs/cross/commit/ec19c9d8f3d0ed0c45a5d66720c15046898f35ea) has to be included)

src/docker.rs
Emilgardis: and removed with f98a6e630507ebfd92af5dbbfbd72e449dc72e7c

instead we provide a build arg

`ARG CROSS_BUILD_IMAGE` that will be whatever cross would use

src/docker.rs
```diff
@@ -351,10 +348,187 @@ pub fn run(
...
+            .wrap_err("when building dockerfile")?;
+    }
+    let pre_build = config.pre_build(target)?;
+
+    if !pre_build.is_empty() {
+        let custom = Dockerfile::Custom {
+            content: format!(
+                r#"
+FROM {image}
+ARG CROSS_DEB_ARCH=
```
Emilgardis, line 382: this should maybe be documented

Emilgardis (commented):
Would it make sense to allow the following

```toml
[target.aarch64-unknown-linux-gnu.pre-build]
lines = ["/lib.sh"]
[[target.aarch64-unknown-linux-gnu.pre-build.copy]]
files = ["scripts/lib.sh"]
location = "/"
```

Emilgardis:
this should be ready now, I'm happy with the functionality.

src/file.rs
```diff
@@ -44,9 +44,16 @@ pub fn write_file(path: impl AsRef<Path>, overwrite: bool) -> Result<File> {
...
+
+    let mut open = fs::OpenOptions::new();
+    open.write(true);
+
+    if overwrite {
+        open.truncate(true).create(true);
+    } else {
+        open.create_new(true);
+    }
+
```
Alexhuszagh, line 56: Nice fix.

src/extensions.rs
```diff
@@ -7,16 +7,24 @@ use crate::errors::*;
...
     fn run_and_get_stdout(&mut self, verbose: bool) -> Result<String, CommandError>;
     fn run_and_get_output(&mut self, verbose: bool) -> Result<std::process::Output, CommandError>;
 }
 
 impl CommandExt for Command {
     fn print_verbose(&self, verbose: bool) {
         if verbose {
-            println!("+ {:?}", self);
+            if let Some(cwd) = self.get_current_dir() {
+                println!("+ {:?} {:?}", cwd, self);
```
Alexhuszagh, line 24: `cwd` might be quite long, or have unusual characters. Maybe this should be quoted?

src/docker/shared.rs
```diff
@@ -361,11 +366,8 @@ pub(crate) fn docker_seccomp(
             "unconfined".to_string()
         } else {
             #[allow(unused_mut)] // target_os = "windows"
-            let mut path = env::current_dir()
-                .wrap_err("couldn't get current directory")?
-                .canonicalize()
-                .wrap_err_with(|| "when canonicalizing current_dir".to_string())?
-                .join("target")
+            let mut path = metadata
+                .target_directory
```
Alexhuszagh, line 370: Very nice fix, this got through the refactoring.

src/docker/shared.rs
```diff
@@ -385,7 +387,73 @@ pub(crate) fn docker_seccomp(
...
+                metadata,
+                engine,
+                &host_root,
+                config.dockerfile_build_args(target)?.unwrap_or_default(),
+                target,
+                verbose,
+            )
+            .wrap_err("when building dockerfile")?;
+    }
+    let pre_build = config.pre_build(target)?;
```
Alexhuszagh, line 427: This also needs to be added to the wiki or the cross_toml documentation.

src/cross_toml.rs
```diff
@@ -23,6 +24,10 @@ pub struct CrossBuildConfig {
     xargo: Option<bool>,
     build_std: Option<bool>,
     default_target: Option<String>,
+    #[serde(default)]
+    pre_build: Vec<String>,
```
Alexhuszagh, line 28: I think this should be similar to #801, and be optional.

src/cross_toml.rs
```diff
@@ -32,11 +37,36 @@ pub struct CrossTargetConfig {
     xargo: Option<bool>,
     build_std: Option<bool>,
     image: Option<String>,
+    #[serde(default, deserialize_with = "opt_string_or_struct")]
+    dockerfile: Option<CrossTargetDockerfileConfig>,
+    #[serde(default)]
```
Alexhuszagh, line 42: Same thing with `Option<Vec<String>>` here.

src/cross_toml.rs
```diff
@@ -69,12 +99,60 @@ impl CrossToml {
...
+    pub fn dockerfile_context(&self, target: &Target) -> Option<String> {
+        self.get_string(
+            target,
+            |b| b.dockerfile.as_ref().and_then(|c| c.context.as_ref()),
+            |t| t.dockerfile.as_ref().and_then(|c| c.context.as_ref()),
+        )
+    }
+
+    /// Returns the `target.{}.dockerfile.build_args` part of `Cross.toml`
+    pub fn dockerfile_build_args(&self, target: &Target) -> Option<HashMap<String, String>> {
```
Alexhuszagh, line 124: I wonder if this could use a helper? For 2 optional Vecs to merge them?

`merge_vec(build, target)` where we have:

```rust
fn merge_vec(x: Option<String>, y: Option<String>) -> Option<String>;
```

src/cross_toml.rs
```diff
@@ -87,24 +165,14 @@ impl CrossToml {
         self.get_bool(target, |b| b.build_std, |t| t.build_std)
     }
 
-    /// Returns the list of environment variables to pass through for `build`,
-    pub fn env_passthrough_build(&self) -> &[String] {
-        &self.build.env.passthrough
+    /// Returns the list of environment variables to pass through for `build` and `target`
+    pub fn env_passthrough(&self, target: &Target) -> (&[String], &[String]) {
+        self.get_vec(target, |_| &[], |t| &t.env.passthrough)
```
Alexhuszagh, line 170: This will need to be updated due to #801 for the new optional arrays, and can just use `|_| None`.

src/cross_toml.rs
```diff
@@ -120,12 +188,16 @@ impl CrossToml {
...
     }
 
-    fn get_string(
-        &self,
+    fn get_string<'a>(
+        &'a self,
         target: &Target,
-        get: impl Fn(&CrossTargetConfig) -> &Option<String>,
+        get_build: impl Fn(&'a CrossBuildConfig) -> Option<&'a String>,
+        get_target: impl Fn(&'a CrossTargetConfig) -> Option<&'a String>,
```
Alexhuszagh, line 195: I wonder if having `Self::get_target` and `get_target` is going to be confusing. Should be fine, just wondering if there's a better naming scheme. Maybe `get_for_target`?

src/docker/shared.rs
```diff
@@ -1,8 +1,12 @@
+pub mod custom;
```
Alexhuszagh, line 1: Can't `custom` just be under `docker`? It can be it's own module and imported into shared. Not sure we need another level of nesting.

Alexhuszagh (commented):
Minor comments, otherwise looks great.

src/docker/shared.rs
```diff
@@ -385,7 +387,73 @@ pub(crate) fn docker_seccomp(
...
+            .wrap_err("when building dockerfile")?;
+    }
+    let pre_build = config.pre_build(target)?;
+
+    if !pre_build.is_empty() {
+        let custom = Dockerfile::Custom {
+            content: format!(
+                r#"
+    FROM {image}
+    ARG CROSS_DEB_ARCH=
```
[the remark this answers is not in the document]
Emilgardis: I've added it to the readme, will definitely add it to the wiki

docs/cross_toml.md
```diff
@@ -41,3 +46,19 @@ This is similar to `build.env`, but allows you to be more specific per target.
...
 passthrough = ["IMPORTANT_ENV_VARIABLES"]
 ```
+
+# `target.TARGET.dockerfile`
+
+```toml
+[target.x86_64-unknown-linux-gnu.dockerfile]
+file = "./Dockerfile"
+context = "."
+build-args = { ARG1 = "foo" }
```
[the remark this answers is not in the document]
Emilgardis: I can add some `#` comments,

```toml
[target.x86_64-unknown-linux-gnu.dockerfile]
file = "./Dockerfile" # The dockerfile to use relative to the `Cargo.toml`
context = "." # What folder to run the build script in
build-args = { ARG1 = "foo" } # https://docs.docker.com/engine/reference/builder/#arg
```

src/cross_toml.rs
Emilgardis: Definitely, I didn't do the change here because for it I needed to change the get function

src/cross_toml.rs
Emilgardis: can also remove the `get_` prefix, I think I prefer that

src/docker/shared.rs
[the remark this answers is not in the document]
Emilgardis: scratch that, I haven't... wiki seems good

docs/cross_toml.md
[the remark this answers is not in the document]
Alexhuszagh: That's perfect.

src/extensions.rs
Alexhuszagh: I mean that the CWD might be something like....

`/home/ahuszagh/PlayOnLinux's virtual drives/amazon_kindle/drive_c/Program Files/Amazon/Kindle/Microsoft.VC90.CRT`

Which, if you notice, isn't exactly very easy to read . We might want to ellipse some characters or just quote the whole thing.

src/cross_toml.rs
Alexhuszagh: I think it's fine. I'd probably have used a `Vec` since we never actually index the values, but the deduplication is nice. I might have used a `BTreeMap` instead in that case though.

Emilgardis:
all done!

Alexhuszagh (approved):

Alexhuszagh:
Went over all the changes from the comments, and it looks good. Once this merges I'll be adding additional documentation to the wiki.

bors r+

PR #682, opened 2022-04-06
more environment configuration and a bug fix

Emilgardis:

PR #684, opened 2022-04-06
enable cargo workspace and path dependencies to work seamlessly

### LINKED ISSUE #388: Building with local dependencies

Emilgardis:
This will allow cargo workspaces to work seamlessly with cross, assuming they don't reference anything outside the current workspace root. This change also automatically mounts (if needed) path dependencies.

If they do reference things outside the workspace root, and that fact is not know to `cargo metadata` (e.g, it's not a workspace member or a path dependency), you'll have to mount it with

```toml
[build.env]
volumes = ["THING"]
```

and set `$THING="/path/to/thing"` (we need to work on accessibility for this feature too, a lot of reported issues are solved by it)

I'm sure there is an issue directly related to workspaces, but I can't find any.

fixes #388

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #543 - Added environment variables to control the UID and GID in the container
 - #524 - docker: Add Nix Store volume support
 - Added support for mounting volumes.
+- #684 - Enable cargo workspaces to work from any path in the workspace, and make path dependencies mount seamlessly.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Re-enabled `sparc64-unknown-linux-gnu` image
 - #582 - Added `libprocstat.so` to FreeBSD images
 - #665 - when not using [env.volumes](https://github.com/cross-rs/cross#mounting-volumes-into-the-build-environment), mount project in /project
+- #494 - Parse Cargo's --manifest-path option to determine mounted docker root
 
 
 ### Removed
--- a/README.md
+++ b/README.md
@@ -362,15 +362,6 @@ $ QEMU_STRACE=1 cross run --target aarch64-unknown-linux-gnu
 9 exit_group(0)
 ```
 
-## Caveats
-
-- path dependencies (in Cargo.toml) that point outside the Cargo project won't
-  work because `cross` use docker containers only mounts the Cargo project so
-  the container doesn't have access to the rest of the filesystem.
-  However, you may use Cargo's `--manifest-path` option to reference your
-  target crate, executed from a common root directory from which all your
-  dependencies are available.
-
 ## Minimum Supported Rust Version (MSRV)
 
 This crate is guaranteed to compile on stable Rust 1.58.1 and up. It *might*
--- a/README.md
+++ b/README.md
@@ -367,6 +367,9 @@ $ QEMU_STRACE=1 cross run --target aarch64-unknown-linux-gnu
 - path dependencies (in Cargo.toml) that point outside the Cargo project won't
   work because `cross` use docker containers only mounts the Cargo project so
   the container doesn't have access to the rest of the filesystem.
+  However, you may use Cargo's `--manifest-path` option to reference your
+  target crate, executed from a common root directory from which all your
+  dependencies are available.
 
 ## Minimum Supported Rust Version (MSRV)
 
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -135,6 +135,21 @@ EOF
                 popd
 
                 rm -rf "${td}"
+                td=$(mktemp -d)
+                git clone \
+                    --depth 1 \
+                    --recursive \
+                    https://github.com/cross-rs/test-workspace "${td}"
+                
+                pushd "${td}"
+      cross run --target "${TARGET}" -p binary --manifest-path="./workspace/Cargo.toml"
+                pushd "workspace"
+                cross run --target "${TARGET}" -p binary
+                pushd "binary"
+                cross run --target "${TARGET}"
+                popd
+                popd
+                popd
             ;;
         esac
 
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -25,6 +25,13 @@ function retry {
   return ${exit_code}
 }
 
+workspace_test() {
+  cross build --target "${TARGET}" --workspace "${@}"
+  cross run --target "${TARGET}" -p binary "${@}"
+  cross run --target "${TARGET}" --bin dependencies \
+    --features=dependencies "${@}"
+}
+
 main() {
     local td=
 
@@ -142,9 +149,9 @@ EOF
                     https://github.com/cross-rs/test-workspace "${td}"
                 
                 pushd "${td}"
-      cross run --target "${TARGET}" -p binary --manifest-path="./workspace/Cargo.toml"
+                TARGET="${TARGET}" workspace_test --manifest-path="./workspace/Cargo.toml"
                 pushd "workspace"
-                cross run --target "${TARGET}" -p binary
+                TARGET="${TARGET}" workspace_test
                 pushd "binary"
                 cross run --target "${TARGET}"
                 popd
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -26,10 +26,11 @@ function retry {
 }
 
 workspace_test() {
-  cross build --target "${TARGET}" --workspace "${@}"
-  cross run --target "${TARGET}" -p binary "${@}"
+  # "${@}" is an unbound variable for bash 3.2, which is the installed version on macOS
+  cross build --target "${TARGET}" --workspace "$@"
+  cross run --target "${TARGET}" -p binary "$@"
   cross run --target "${TARGET}" --bin dependencies \
-    --features=dependencies "${@}"
+    --features=dependencies "$@"
 }
 
 main() {
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -1,9 +1,12 @@
 use serde::Deserialize;
+use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 
+use crate::cli::Args;
+use crate::config::Config;
 use crate::errors::*;
-use crate::extensions::CommandExt;
+use crate::extensions::{CommandExt, OutputExt};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Subcommand {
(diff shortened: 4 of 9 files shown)


Emilgardis (commented):
https://github.com/cross-rs/cross/tree/14965587d7f9dccac34598495fcd3291c067deba#caveats should be updated

Emilgardis:
We can also use `workspace_members` and mount members as volumes if they are not on the same "root" folder.

we can also walk through all non-workspace packages, and if `source` is `path+file://`, mount that as well.

This will effectively make path dependencies zero-cost, as long as those paths are defined and shown with `cargo metadata`

Emilgardis:
this is possibly an argument for using cargo as a library, but that locks us to the features of the specific cargo version that is used.

src/docker.rs
```diff
@@ -145,6 +151,12 @@ pub fn run(
         docker.args(&["-e", var]);
     }
     let mut env_volumes = false;
+    // FIXME(emilgardis 2022-04-07): This is a fallback so that if it's hard for use to do mounting logic, make it simple(r)
```
Emilgardis, line 151: *us*

Emilgardis:

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #494 - Parse Cargo's --manifest-path option to determine mounted docker root
```
Emilgardis, line 8: need to add own changelog

Emilgardis, line 8: need to add own changelog entry

README.md
```diff
@@ -355,6 +355,9 @@ $ QEMU_STRACE=1 cross run --target aarch64-unknown-linux-gnu
 - path dependencies (in Cargo.toml) that point outside the Cargo project won't
   work because `cross` use docker containers only mounts the Cargo project so
   the container doesn't have access to the rest of the filesystem.
+  However, you may use Cargo's `--manifest-path` option to reference your
+  target crate, executed from a common root directory from which all your
+  dependencies are available.
```
Emilgardis, line 360: modify

ci/test.sh
```diff
@@ -135,6 +135,21 @@ EOF
...
+                
+                pushd "${td}"
+      cross run --target "${TARGET}" -p binary --manifest-path="./workspace/Cargo.toml"
+                pushd "workspace"
+                cross run --target "${TARGET}" -p binary
+                pushd "binary"
+                cross run --target "${TARGET}"
+                popd
+                popd
+                popd
```
Emilgardis, line 152: im not a bash expert, can you tell?

Is it possible to do the pops in one command?

src/cargo.rs
```diff
@@ -52,38 +53,96 @@ impl<'a> From<&'a str> for Subcommand {
...
     }
-}
 
-/// Cargo project root
-pub fn root() -> Result<Option<Root>> {
-    let cd = env::current_dir().wrap_err("couldn't get current directory")?;
+    pub fn path_dependencies(&self) -> impl Iterator<Item = &Path> {
+        // TODO: Also filter out things that are in workspace, but not a workspace member
+        self.non_workspace_members().filter_map(|p| p.crate_path())
+    }
```
Emilgardis, line 74: this todo is not super important, but probably good to have since it minifies the mounts needed

src/cargo.rs
```diff
@@ -52,38 +53,96 @@ impl<'a> From<&'a str> for Subcommand {
...
+        command.args(["--filter-platform", target.triple()]);
+    }
+    let output = command.output()?;
+    let manifest: Option<CargoMetadata> =
+        serde_json::from_slice(&output.stdout).wrap_err_with(|| {
+            format!(
+                "{command:?} returned nothing. {:?}",
+                String::from_utf8(output.stderr)
+            )
+        })?;
```
Emilgardis, line 127: this needs to be modified to be "nicer"

src/cargo.rs
```diff
@@ -52,38 +53,96 @@ impl<'a> From<&'a str> for Subcommand {
...
+    manifest
+        .map(|m| -> Result<_> {
+            Ok(CargoMetadata {
+                target_directory: args
+                    .and_then(|a| a.target_dir.clone())
+                    .map(|p| {
+                        if p.is_relative() {
+                            cd.expect("this is a bug, working directory should be provided here")
+                                .join(p)
```
Emilgardis, line 136: this will fail if `--target-dir` is absolute I think. but if `--target-dir` is absolute then we'd need to figure out if it's mounted already or if we need to mount it

src/docker.rs
```diff
@@ -230,10 +255,21 @@ pub fn run(
...
+        // We do this to avoid clashes with path separators. Windows uses `\` as a path separator on Path::join
+        let cwd = &cwd;
+        let working_dir = Path::new("project").join(cwd.strip_prefix(&metadata.workspace_root)?);
+        // No [T].join for OsStr
+        let mut mount_wd = std::ffi::OsString::new();
+        for part in working_dir.iter() {
+            mount_wd.push("/");
+            mount_wd.push(part);
+        }
+        docker.args(&["-w".as_ref(), mount_wd.as_os_str()]);
```
Emilgardis, line 272: this is sad, but must be done.

Emilgardis:
it works! https://github.com/cross-rs/cross/runs/5873527788?check_suite_focus=true#step:7:1482

If anyone is able to review this before it's actually ready, please do <3

Emilgardis:
ping @wngr

this integrates #494

src/main.rs
```diff
@@ -406,10 +407,10 @@ fn run() -> Result<ExitStatus> {
 
 /// Parses the `Cross.toml` at the root of the Cargo project or from the
 /// `CROSS_CONFIG` environment variable (if any exist in either location).
-fn toml(root: &Root) -> Result<Option<CrossToml>> {
+fn toml(root: &CargoMetadata) -> Result<Option<CrossToml>> {
```
Emilgardis, line 410: rename
```suggestion
fn toml(metadata: &CargoMetadata) -> Result<Option<CrossToml>> {
```

Emilgardis:
bors try

src/cargo.rs
```diff
@@ -52,38 +53,96 @@ impl<'a> From<&'a str> for Subcommand {
...
+            command.args(["--manifest-path".as_ref(), manifest_path.as_os_str()]);
         }
+    } else {
+        command.arg("--no-deps");
     }
-
-    Ok(None)
+    if let Some(target) = args.and_then(|a| a.target.as_ref()) {
+        command.args(["--filter-platform", target.triple()]);
+    }
```
Emilgardis, line 119: should also forward `--features`

ci/test.sh
Alexhuszagh: I think this is the best approach.

You can use `pushd -0 && dirs -c` to go to the bottom of the stack and then clear the stack, but this won't work if there's already a so it's better to just use `popd` multiple times IMO. I believe there's no correct way to do this using arguments to popd, since they just remove items from specific indexes in the `dirs` stack. This could be changed to a single line with `popd; popd; popd`.

The issue with it is as follows:

```bash
#!/bin/bash

call() {
	pushd .local
	pushd share
	pushd applications
	echo "$(ls -la .)"
	echo "dirs before clear: $(dirs)"
	pushd -0 && dirs -c
	# Correct way: popd; popd; popd
	echo "pwd inside call: $(pwd)"
}

pushd ~
call
echo "dirs after clear: $(dirs)"
echo "pwd outside call: $(pwd)"
```

This means that after `call`, we'll be back in the original directory, not in `~`, which is probably what was wanted.

Emilgardis:
bors try --target x86_64

Emilgardis:
bors try --target x86_64

Emilgardis:
I think this is ready

src/docker.rs
```diff
@@ -133,38 +168,28 @@ pub fn run(
...
-    let mut env_volumes = false;
+    let mut mount_volumes = false;
+    // FIXME(emilgardis 2022-04-07): This is a fallback so that if it's hard for us to do mounting logic, make it simple(r)
+    // Preferably we would not have to do this.
+    if cwd.strip_prefix(&metadata.workspace_root).is_err() {
+        mount_volumes = true;
+    }
+
     for ref var in config.env_volumes(target)? {
         validate_env_var(var)?;
```
Alexhuszagh, line 179: I don't believe this fixes the `VAR=VALUE` issue mentioned in #743.

Alexhuszagh (changes requested):
Also, can you squash the commits and rebase?

Emilgardis:
yep!

src/docker.rs
Emilgardis: i never said it would? thats #682

Emilgardis:
I found a bug too which I will fix, been switching between local computers and servers so not 100% sure where the fix is :3

src/docker.rs
Alexhuszagh: Ok ok my bad, I'll mark this as resolved then.

Alexhuszagh (approved):

src/cli.rs
```diff
@@ -82,7 +86,21 @@ pub fn parse(target_list: &TargetList) -> Result<Args> {
...
+                    manifest_path = env::current_dir().ok().map(|cwd| cwd.join(p));
+                }
+            } else if arg.starts_with("--manifest-path=") {
+                manifest_path = arg
+                    .split_once('=')
+                    .map(|x| x.1)
+                    .map(PathBuf::from)
+                    .and_then(|p| env::current_dir().ok().map(|cwd| cwd.join(p)));
+                all.push(arg);
+            } else if let ("+", ch) = arg.split_at(1) {
```
Emilgardis, line 103: we should add a utility to do this operation easier and in a standardized way

Emilgardis:
rebased and squashed.
the "bug" i fixed was that we error on non-existing targets, now we fallback to cargo.

Emilgardis:
bors try --target unknown-linux-gnu

Emilgardis:
bors try --target *unknown-linux-gnu

Alexhuszagh (approved):
Everything looks great and the updated tests all pass. I believe this is ready to merge. Should I `r+` it?

Alexhuszagh:
bors try --target x86_64-apple-darwin

PR #685, opened 2022-04-07
Add android libc version

dalance:
This PR adds libc version of android because bionic is provided as libc on android.
QEMU version is added too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,23 +264,23 @@ terminate.
 | Target                               |  libc  |   GCC   | C++ | QEMU  | `test` |
 |--------------------------------------|-------:|--------:|:---:|------:|:------:|
 | `*-apple-ios` [1]                    | N/A    | N/A     | N/A | N/A   |   ✓    |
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #685 - Add android libc version
```
Emilgardis, line 8: it's fine to just add this to the previous line, no need for a separate entry

Emilgardis (changes requested):

CHANGELOG.md
Emilgardis: or, we can skip modifying changelog entirely

dalance: I removed the commit.
Now changelog check is failed. Is it OK?

Emilgardis: Yes 👍

Emilgardis (approved):
bors r+

PR #687, opened 2022-04-08
Fix broken link in `dropbear.sh`

Jules-Bertholet:
The broken link prevented Docker images from building.

### DIFF
--- a/docker/dropbear.sh
+++ b/docker/dropbear.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 . lib.sh
 
 main() {
-    local version=2020.80
+    local version=2022.82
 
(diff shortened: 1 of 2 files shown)


Emilgardis (dismissed):
bors r+

thanks!

I assume this wasn't caught earlier because of cache. We should maybe build at least one image without cache to catch this.

Emilgardis:
This also bumps the version, but I think that's fine

Emilgardis:
Interesting, I don't understand how these are failing

Emilgardis:
bors try

Emilgardis (approved):
bors r+

thanks!

PR #694, opened 2022-04-21
Upgrade nix to 0.24, limit features

rtzoeller:
This reduces cross's compile time slightly, and removes memoffset as an indirect dependency.


Emilgardis (approved):
thanks!

will merge when ci blocker is fixed

Emilgardis:
bors cancel

PR #703, opened 2022-04-29
fix #701 #702

Emilgardis:


Emilgardis:
bors is great, its documentation is lacking though...

see #701 and #702

PR #707, opened 2022-05-08
Add sysroot to `BINDGEN_EXTRA_CLANG_ARGS` env var

Jules-Bertholet:
This allows projects using [`rust-bindgen`](https://github.com/rust-lang/rust-bindgen) to build properly.

PR #708, opened 2022-05-08
Fix `BINDGEN_EXTRA_CLANG_ARGS`

Jules-Bertholet:
They are supposed to be lowercase 😳
It seems I didn't test #707 as well as I thought…


Emilgardis (approved):
:P 

bors r+

PR #713, opened 2022-05-24
Convert --target-dir to use absolute paths.

### LINKED ISSUE #581: `--target-dir` results in `failed to create directory /target/release`
Cross is running docker with `--user`, and when you use `--target-dir` it adds a volume mount from the directory to `/target`, like so:

```bash
docker run <...> -v mytargetdir:/target:Z <...>
```

The `/target` directory will be owned by root, and cross will fail to create anything there.

Alexhuszagh:
Converts relative target directories to absolute paths, to avoid creating the target directory in the sysroot. This keeps the prior behavior if the provided target directory is an absolute path.

Fixes #581.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,6 +2,7 @@ use std::str::FromStr;
 use std::{env, path::PathBuf};
 
 use crate::cargo::Subcommand;
+use crate::errors::Result;
 use crate::rustc::TargetList;
 use crate::Target;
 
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -15,7 +16,16 @@ pub struct Args {
     pub docker_in_docker: bool,
 }
 
-pub fn parse(target_list: &TargetList) -> Args {
+// Fix for issue 581. target_dir must be absolute.
```
Emilgardis, line 19:
```suggestion
// Fix for issue #581. target_dir must be absolute.
```

Emilgardis (changes requested):

Emilgardis (approved):
bors r+

PR #714, opened 2022-05-24
Use host target directory when calling host cargo.

### LINKED ISSUE #619: cross clean does not work with --target-dir
Running `cross clean --target-dir=/path/to/target-dir/` does not delete anything and no error is printed when running with --verbose.

cross version: 0.2.1
cargo version: 1.57.0

Alexhuszagh:
When using docker, the target directory is mounted at `/target`. However, when falling back to the host cargo, if we provide a custom target directory, we incorrectly pass it as `--target-dir /target`. To rectify this, this PR passes the original command-line arguments to cargo.

Fixes #619.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #714 - use host target directory when falling back to host cargo.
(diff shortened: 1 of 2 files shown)


Emilgardis (changes requested):
there's no need to mention why this specific change was done in the comment, rather we should have done this from the beginning.

Emilgardis (approved):
bors r+

Issue #82, opened 2017-03-21
Add full android test support


malbarbo:
~~Should https://github.com/snipsco/dinghy be used?~~

Support for testing android binaries that depends only on libc was added in https://github.com/japaric/cross/pull/132.

We should add support to test android binaries in the android emulator. See https://github.com/japaric/cross/pull/132 for a brief discussion of the issues of distributing the android emulator.

japaric:
If it can provide a transparent testing experience like the other `cross test` targets, sure!

jan-hudec:
Since https://github.com/rust-lang/cargo/pull/3954 it is possible to run the generated binaries through a wrapper. So there could be a wrapper that would start the emulator (which can be done; [here are Travis' instructions](https://docs.travis-ci.com/user/languages/android/#How-to-Create-and-Start-an-Emulator)) and then run the tests there.

jamesmunns:
@malbarbo is this okay to close since #132 has landed?

malbarbo:
@jamesmunns I updated the issue to make clear what is missing.

Alexhuszagh:
This would require a full-system Android emulator, which I believe is beyond the scope of `cross` itself. The best way to test the binaries would be any of the many Android emulators, which also require a full Android image, including Android Studio, BlueStacks, or the open-source [anbox](https://github.com/anbox/anbox/blob/master/docs/install.md). Anbox supports Snap installs, and so should be compatible with our images.  If anyone wants to document this on the wiki, that would be appreciated, but it is assumed most developers targeting Android already have these tools available.

PR #720, opened 2022-05-25
Add android runner to run android binaries.

Alexhuszagh:
Android binaries currently fail to run linking against certain dependencies, because `libc++_shared.so` is not found, so ensure the library is preloaded via `LD_PRELOAD`. This creates a simple `android-runner` file, so any additional changes can be added there.

This merely links to the proper `libc++_shared.so`, so it doesn't fix other issues. The [hello world project](https://github.com/cross-rs/rust-cpp-hello-word) now compiles and runs on `aarch64-linux-android`, but not `i686-linux-android`, which seems to be entirely unrelated (likely a permission error on the cloud Fedora server I'm running).

```text
error getting old personality value: Operation not permitted
libc: error getting old personality value: Operation not permitted
Aborted (core dumped)
```

Linked Issues: #82

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,11 +147,11 @@ jobs:
             - { target: i686-unknown-linux-musl,          os: ubuntu-latest,                    std: 1, run: 1 }
             - { target: mips-unknown-linux-musl,          os: ubuntu-latest,  cpp: 1,           std: 1, run: 1 }
             - { target: mipsel-unknown-linux-musl,        os: ubuntu-latest,  cpp: 1,           std: 1, run: 1 }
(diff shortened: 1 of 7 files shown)


Emilgardis:
would it be possible to add a runner to the ci?

Alexhuszagh:
> would it be possible to add a runner to the ci?

We already have a runner, but I believe to reproduce this error we require the `cc` dependency. [rust-cpp-hello-world](https://github.com/cross-rs/rust-cpp-hello-word) fails, but `cargo init --bin --name hello . && cross run --target aarch64-linux-android` passes.

Should we try to build against a binary with a `cc` dependency rather than just against an empty hello world project?

Emilgardis:
we can make a new test repo for it to try it on

Emilgardis:
bors try --target android

Emilgardis (approved):
bors r+

Issue #225, opened 2018-11-07
Doc tests are not run


gnzlbg:
Currently, `cross test` does not run documentation tests.

Alexhuszagh:
This still seems to depend on a nightly compiler, as documented in [doctest-xcompile](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile)
- https://github.com/rust-lang/cargo/issues/7040
- https://github.com/rust-lang/rust/issues/64245

If the subcommand is `test` and the compiler is nightly, we may be able to provide the `-Zdoctest-xcompile` flag.

CraftSpider:
As of Rust 1.89, doctest-xcompile has stabilized, meaning doctests should now be able to run

PR #721, opened 2022-05-26
Add preliminary support for running doctests.

Alexhuszagh:
This partially addresses #225.

This only works on nightly, due to the use of the unstable feature [doctest-xcompile](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile)

The relevant tracking issues are:
- https://github.com/rust-lang/cargo/issues/7040
- https://github.com/rust-lang/rust/issues/64245

If the subcommand is `test` and the compiler is nightly, we provide the `-Zdoctest-xcompile` flag if `CROSS_UNSTABLE_ENABLE_DOCTESTS=true`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,6 +21,7 @@ use std::path::PathBuf;
 use std::process::ExitStatus;
 
 use config::Config;
+use rustc_version::Channel;
 use serde::Deserialize;
 
 use self::cargo::{Root, Subcommand};
@@ -314,6 +315,7 @@ fn run() -> Result<ExitStatus> {
                 default_toolchain.to_string()
             };
             sysroot.set_file_name(&toolchain);
+            let mut is_nightly = toolchain.contains("nightly");
 
             let installed_toolchains = rustup::installed_toolchains(verbose)?;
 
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -279,6 +279,7 @@ fn run() -> Result<ExitStatus> {
 
     let host_version_meta =
         rustc_version::version_meta().wrap_err("couldn't fetch the `rustc` version")?;
+    let is_nightly = rustc::is_nightly(&args.channel, &host_version_meta);
```
Emilgardis, line 282: this is wrong, it should check the target meta, not the host meta. see the logic in L325/L324

Emilgardis (changes requested):

src/main.rs
Alexhuszagh: I believe this is fixed by detecting the channel in `rustup::rustc_version`. First we detect if the toolchain contains nightly, and if we can parse the toolchain info via `rustup::rustc_version`, then we update that. This might be slightly redundant, however: I'm not sure if I need both assignments. The logic is effectively:

```rust
let toolchain = ...;
let mut is_nightly = toolchain.contains("nightly");
if let Some((rustc_version, channel, rustc_commit)) = rustup::rustc_version(&sysroot)? {
    is_nightly = channel == Channel::Nightly;
}
```

src/rustup.rs
```diff
@@ -101,7 +101,21 @@ pub fn component_is_installed(component: &str, toolchain: &str, verbose: bool) -
...
+    if version.contains("dev") {
+        Channel::Dev
+    } else if version.contains("nightly") {
+        Channel::Nightly
+    } else if version.contains("beta") {
+        Channel::Beta
+    } else {
+        Channel::Stable
+    }
+}
```
Emilgardis, line 116: this is how rustc_version does this

```suggestion
fn rustc_channel(version: &Version) -> Channel {
    let channel = match version.pre.split('.').next().unwrap() {
        "" => Channel::Stable,
        "dev" => Channel::Dev,
        "beta" => Channel::Beta,
        "nightly" => Channel::Nightly,
        x => return Err(Error::UnknownPreReleaseTag(x.to_owned())),
    };
}
```

I think I prefer that approach. (atleast using match instead of a if else chain)

Emilgardis (changes requested):

Emilgardis:
could you squash the changes? feel free to remove the `co-authored by` for me

Emilgardis (approved):
bors r+

PR #722, opened 2022-05-26
Evaluate boolean environment variables as truthy or falsey.

Alexhuszagh:
Allow values of 0, 1, so values other than `true` or `false` can be provided. `VAR=0`, `VAR=`, `VAR=-0`, or `VAR=false` will evaluate to false, while the rest will evaluate to true.

Affects #661 and #721.


Emilgardis (approved):
bors r+

PR #727, opened 2022-05-29
Add `PKG_CONFIG_PATH` to all `*-linux-gnu` images.

### LINKED ISSUE #404: Define PKG_CONFIG_PATH in Docker images
As stated on https://github.com/rust-embedded/cross/issues/149#issuecomment-610265947 one has to manually define `PKG_CONFIG_PATH` in their custom docker image in order to build Rust code relying on external libs.

Automatically defining it in the `cross` images would make things easier as users would just have to derive from the official images and install their extra packages.
Would you consider such change in the docker files?

Alexhuszagh:
This adds support for pkg-config to all `*-linux-gnu` images, since `pkg-config` is already installed on these images. We export `PKG_CONFIG_PATH` which allows packages installed from `apt` to be automatically added to the search path when linking.

See [cross_dbus_pkgconfig](https://github.com/Alexhuszagh/cross_dbus_pkgconfig) for examples building and linking against a library with a D-BUS dependency.

Closes #404.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #727 - add `PKG_CONFIG_PATH` to all `*-linux-gnu` images.
 - #725 - support `CROSS_DEBUG` and `CROSS_RUNNER` on android images.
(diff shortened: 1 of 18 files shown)


Emilgardis (approved):
bors r+

Issue #695, opened 2022-04-22
Make FreeBSD builds more resiliant


Emilgardis:
FreeBSD frequently breaks our images, removing older resources.

We need a way to select a range of versions to not constantly have to patch them.

Emilgardis:
Is there some kind of authoritative file provided for us query the latest versions from?

Alexhuszagh:
We could always use the archive, or we could query using FTP and a regex to find the latest version:

For example, in [FreeBSD-Archive i386](http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/i386/), we have all older versions of 12 from 12.0-12.2.

For querying the latest release, we could always use curl:

```bash
max_freebsd_12() {
    local best=
    local minor=0
    local m=
    for release in $@; do
        m=$(echo "$release" | cut -d '.' -f 2 | cut -d '-' -f 1)
        if [ "$m" -gt "$minor" ]; then
            best="$release"
            minor="$m"
        fi
    done
    if [[ -z "$best" ]]; then
        echo "Could not find best release for FreeBSD 12." 1>&2
        exit 1
    fi
    echo "$best"
}

releases=($(curl --silent --list-only ftp.freebsd.org/pub/FreeBSD/releases/i386/ | grep RELEASE))
latest=$(max_freebsd_12 ${releases[@]})
```

This would break if FreeBSD 12 is no longer supported, however.

Alexhuszagh:
A maintainer of FreeBSD has also suggested this [URL](https://people.freebsd.org/~jmg/FreeBSD-snap/snapshot.idx.xz), which we can decompress to find the latest, authoritative versions. Each line is off the following:

```text
iso 12.3-RELEASE amd64 bootonly xxx unspec xxx FreeBSD-12.3-RELEASE-amd64-bootonly.iso https://download.freebsd.org/ftp/releases/ISO-IMAGES/12.3/FreeBSD-12.3-RELEASE-amd64-bootonly.iso 20211207210319.1F6683569@freefall.freebsd.org
...
iso 12.3-RELEASE i386 bootonly xxx unspec xxx FreeBSD-12.3-RELEASE-i386-bootonly.iso https://download.freebsd.org/ftp/releases/ISO-IMAGES/12.3/FreeBSD-12.3-RELEASE-i386-bootonly.iso 20211207210319.1F6683569@freefall.freebsd.org
```

[...]

PR #729, opened 2022-05-30
Refactor code to use install_packages and purge_packages.

Alexhuszagh:
Remove all local use of installing manually via the package manager, using `install_packages` and `purge_packages` rather than use redundant code. Simplifies our shell scripts to build docker images.

Still a draft because I'm currently testing it on all images.

### DIFF
--- a/docker/Dockerfile.x86_64-unknown-linux-gnu
+++ b/docker/Dockerfile.x86_64-unknown-linux-gnu
@@ -1,6 +1,7 @@
 FROM ubuntu:16.04
 ARG DEBIAN_FRONTEND=noninteractive
 
+COPY lib.sh /
 COPY linux-image.sh /
 RUN /linux-image.sh x86_64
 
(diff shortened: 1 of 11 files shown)


Alexhuszagh:
@Emilgardis Is there a way to disable the changelog check when submitting a PR? This doesn't really have any significant changes: it's just refactoring code.

Emilgardis:
use the label "no changelog" :)

Alexhuszagh:
bors try

Alexhuszagh:
This works locally on my machine for a few images, so there's no obvious logic error for me (and I checked to see if `dependencies` or `install_packages` or `purge_list` was used elsewhere in these images, and it didn't see so), so I'll try it on bors.

Alexhuszagh:
Unrelated bugs are causing builds to [fail](https://github.com/cross-rs/cross/runs/6658303164?check_suite_focus=true), which I'm addressing in a separate PR. The two are somewhat interlinked, so I can merge this first or merge the other afterwards. I'll be submitting a patch for #695 shortly.

PR #730, opened 2022-05-30
Make FreeBSD builds more resilient.

### LINKED ISSUE #695: Make FreeBSD builds more resiliant

Alexhuszagh:
Do not hardcode exact versions of the FreeBSD release, `openssl`, or `sqlite3` versions. Rather, use the FreeBSD ftp server to determine the latest `12.X` release, and use the package list downloaded to determine the latest `openssl` and `sqlite3` versions.

Since we use a dynamic version of FreeBSD, and have no good way of detecting this later, we write the version and revision to `/opt/freebsd-version`.

Fixes #695.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- #730 - make FreeBSD builds more resilient.
 - #727 - add `PKG_CONFIG_PATH` to all `*-linux-gnu` images.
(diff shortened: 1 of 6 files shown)


Alexhuszagh:
bors try

Emilgardis:
Great! One sad thing though, because we cache from the previous images, this won't invalidate often. We should look into that separately.

Emilgardis (approved):
bors r+

PR #746, opened 2022-06-03
Limit permissions for Android images.

Alexhuszagh:
Remove the use of the `--privileged` flag for Android images and instead use an seccomp permissions. The provided profile is derived from the docker documentation, with slight modifications to allow `clone` and `clone3`.

The documentation is [docker seccomp](https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile), which details the syscalls blocked by docker. The same is true for podman. We merely modified these settings to allow `personality` syscall, which then allows us to use our Android images.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -1,9 +1,11 @@
+use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 use std::{env, fs};
 
 use crate::cargo::CargoMetadata;
 use crate::extensions::{CommandExt, SafeCommand};
+use crate::file::write_file;
 use crate::id;
 use crate::{errors::*, file};
 use crate::{Config, Target};
@@ -14,16 +16,34 @@ const DOCKER_IMAGES: &[&str] = &include!(concat!(env!("OUT_DIR"), "/docker-image
 const CROSS_IMAGE: &str = "ghcr.io/cross-rs";
 const DOCKER: &str = "docker";
 const PODMAN: &str = "podman";
+// secured profile based off the docker documentation for denied syscalls:
+// https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile
+// note that we've allow listed `clone` and `clone3`, which is necessary
+// to fork the process, and which podman allows by default.
+const SECCOMP: &str = include_str!("seccomp.json");
+
+#[derive(Debug, PartialEq, Eq)]
+enum EngineType {
+    Docker,
+    Podman,
+    Other,
+}
 
 // determine if the container engine is docker. this fixes issues with
 // any aliases (#530), and doesn't fail if an executable suffix exists.
-fn get_is_docker(ce: std::path::PathBuf, verbose: bool) -> Result<bool> {
+fn get_engine_type(ce: std::path::PathBuf, verbose: bool) -> Result<EngineType> {
     let stdout = Command::new(ce)
         .arg("--help")
         .run_and_get_stdout(verbose)?
         .to_lowercase();
 
-    Ok(stdout.contains("docker") && !stdout.contains("emulate"))
+    if stdout.contains("podman") {
+        Ok(EngineType::Podman)
+    } else if stdout.contains("docker") && !stdout.contains("emulate") {
+        Ok(EngineType::Docker)
+    } else {
+        Ok(EngineType::Other)
+    }
 }
 
 pub fn get_container_engine() -> Result<std::path::PathBuf, which::Error> {
@@ -172,7 +192,7 @@ pub fn run(
     let runner = config.runner(target)?;
 
     let mut docker = docker_command("run")?;
-    let is_docker = get_is_docker(get_container_engine().unwrap(), verbose)?;
+    let engine_type = get_engine_type(get_container_engine().unwrap(), verbose)?;
 
     for ref var in config.env_passthrough(target)? {
         validate_env_var(var)?;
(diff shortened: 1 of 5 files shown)


Alexhuszagh:
I've confirmed this works on Linux, WSL2, Docker Desktop for Windows. It should work without issue on macOS, but I haven't manually confirmed that yet.

src/docker.rs
```diff
@@ -170,8 +176,25 @@ pub fn run(
...
-    if target.needs_docker_privileged() {
-        docker.arg("--privileged");
+    // docker uses seccomp now on all installations
+    if target.needs_docker_seccomp() {
+        // write to our seccomp profile if it doesn't exist.
+        let mut path = env::current_dir().wrap_err("couldn't get current directory")?;
+        path.push("target");
+        path.push(target.triple());
+        fs::create_dir_all(&path).wrap_err("couldn't make target directory")?;
+        path.push("seccomp.json");
```
Emilgardis, line 186: this should be a function, maybe place it in `file`

Emilgardis (changes requested):

Alexhuszagh:
I should probably add: this still allows you to pass the `--privileged` which will override the `--security opts seccomp=...`, so if users have a crate that requires potentially dangerous syscalls to run, they can always disable these security settings.

src/docker.rs
Emilgardis: I fixed it :)

Alexhuszagh: Ok perfect, I'll modify this then.

Alexhuszagh: Is that better now?

Alexhuszagh:
Update: this doesn't work with long file names on Windows, I'll probably have to fix this to a UNC path or something.

Alexhuszagh:
On Windows Docker, this seems to be a Docker issue which I've filed upstream. We might just have to use `unconfined` instead of the path if we're using Docker rather than Podman and on Windows.

https://github.com/docker/for-win/issues/12760

src/file.rs
```diff
@@ -19,3 +19,19 @@ fn read_(path: &Path) -> Result<String> {
...
     Ok(s)
 }
+
+pub fn write_file(path: impl AsRef<Path>, overwrite: bool) -> Result<File> {
+    let path = path.as_ref();
+    fs::create_dir_all(&path.parent().ok_or_else(|| {
+        eyre::eyre!(format!(
+            "could not find parent directory for `{}`",
+            path.display()
+        ))
```
Emilgardis, line 29:
```suggestion
        eyre::eyre!(
            "could not find parent directory for `{}`",
            path.display()
        )
```

Emilgardis (commented):
nice, do we need to wait for the upstream issue to be fixed? I don't think so

Alexhuszagh:
> nice, do we need to wait for the upstream issue to be fixed? I don't think so

Nope, I've added a workaround for Docker on Windows. I'm going to rebase, then try to check on OS X, and it all passes, I'll make this ready for review.

Alexhuszagh:
Ok I've tested this on the following targets, and this passes on all (except podman on macOS, but for unrelated issues, see #756):

- Windows 10 (Docker Desktop)
- Windows 10 (Podman, not run inside WSL2)
- Windows 10 (Docker, run inside WSL2)
- Windows 10 (Podman, run inside WSL2)
- macOS Catalina x86_64 (Docker Desktop)
- macOS Catalina x86_64 (Podman)
- Fedora 36 (Docker)
- Fedora 36 (Podman)

I've had to do a few special workarounds for Windows, but those work well. This is ready for review and merging.

src/docker.rs
```diff
@@ -14,16 +16,23 @@ const DOCKER_IMAGES: &[&str] = &include!(concat!(env!("OUT_DIR"), "/docker-image
...
+// secured profile based off the docker documentation for denied syscalls:
+// https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile
+// note that we've allow listed `clone` and `clone3`, which is necessary
+// to fork the process, and which podman allows by default.
+const SECCOMP: &str = include_str!("seccomp.json");
 
 // determine if the container engine is docker. this fixes issues with
 // any aliases (#530), and doesn't fail if an executable suffix exists.
-fn get_is_docker(ce: std::path::PathBuf, verbose: bool) -> Result<bool> {
+fn get_engine_type(ce: std::path::PathBuf, verbose: bool) -> Result<(bool, bool)> {
```
Emilgardis, line 27: this could be an enum

src/docker.rs
```diff
@@ -233,8 +243,33 @@ pub fn run(
...
+                .join(target.triple())
+                .join("seccomp.json");
+            if !path.exists() {
+                write_file(&path, false)?.write_all(SECCOMP.as_bytes())?;
+            }
+            #[cfg(target_os = "windows")]
+            if is_podman {
+                // podman weirdly expects a WSL path here, and fails otherwise
+                path = wslpath(&path, verbose)?;
+            }
```
Emilgardis, line 268: I think the windows path thing would be solved by canonicalizing with `file::canonicalize`, haven't tested it myself

Alexhuszagh: It is not, see https://github.com/docker/for-win/issues/12760.

Emilgardis: sorry, meant the podman issue

otavio (approved):

otavio:
bors r=Alexhuszagh

Alexhuszagh:
> bors r=Alexhuszagh

I think that's supposed to be `bors r=Otavio,Emilgardis`.

Emilgardis:
use `bors r=user` when they are the reviewer, in this case you should just do `bors r+`

Emilgardis:
bors r-

bors r=otavio,emilgardis

PR #747, opened 2022-06-03
Reduce Android image sizes.

### LINKED ISSUE #739: Reduce Android Image Sizes
Currently, Android images are very large, and could easily be reduced in size since we only use one toolchain per target. Since the builds use different arguments per architecture, we can't share layers between these builds. These images are ~4GB and ~2.5GB is the `/android-ndk` folder. Each folder in `/android-ndk/sysroot/usr/lib` is ~180MB large, so deleting all but those for the desired toolchain will save ~600MB. The other directories are much less significant, but we could reduce the image size from ~4GB to ~3.4GB.

The other solution could be to allow us to cache layers. Since this is the first step that specifies the architecture in the Dockerfile, we could simply remove the architecture, move the desired logic outside, and then all images would share that layer.

```Dockerfile
FROM ubuntu:20.04
ARG DEBIAN_FRONTEND=noninteractive

COPY common.sh lib.sh /
RUN /common.sh

COPY cmake.sh /
RUN /cmake.sh

COPY xargo.sh /
RUN /xargo.sh

COPY android-ndk.sh /
RUN /android-ndk.sh arm 28
ENV PATH=$PATH:/android-ndk/bin

...
```

Alexhuszagh:
Delete unused toolchains to save ~600MB in our total image sizes. This is done by removing the toolchains for other targets in `/android-ndk/sysroot/usr/lib/`.

Closes #739.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- #747 - reduced android image sizes.
 - #377 - update WINE versions to 7.0.
 - #734 - patch `arm-unknown-linux-gnueabihf` to build for ARMv6, and add architecture for crosstool-ng-based images.
 - #709 - Update Emscripten targets to `emcc` version 3.1.10
--- a/docker/android-ndk.sh
+++ b/docker/android-ndk.sh
@@ -27,6 +27,30 @@ main() {
       --arch "${arch}" \
       --api "${api}"
 
+    # clean up unused toolchains to reduce image size
+    local triple
+    local triples
+    local triple_arch="${arch}"
+    case "${arch}" in
+      arm64)
+        triple_arch="aarch64"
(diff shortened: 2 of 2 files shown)


otavio (approved):

Alexhuszagh:
bors try --target *android*

Alexhuszagh:
Ok I've fixed these issues by converting the provided architecture to the triple prefix:

```bash
    local triple_arch="${arch}"
    case "${arch}" in
      arm64)
        triple_arch="aarch64"
        ;;
      x86)
        triple_arch="i686"
        ;;
    esac
    triples=(...)
    for triple in "${triples[@]}"; do
      if ! [[ "${triple}" =~ ^"${triple_arch}".* ]]; then
        rm -rf "/android-ndk/sysroot/usr/lib/${triple}"
      fi
    done
```

Emilgardis (approved):
bors r+

PR #748, opened 2022-06-03
Allow variable definitions in env.pasthrough.

### LINKED ISSUE #743: Allow Environment Variable Definitions in `env.passthrough`.
### Checklist

### Describe your request

Currently when we specify passthrough variables for cross, we cannot define the passthrough variable. `docker run` supports the `--env VAR=VALUE` syntax, so this should not require much work to support. In fact, removing the following lines and splitting on `=` to ensure that the key is not `CROSS_RUNNER` should fix this.

https://github.com/cross-rs/cross/blob/1689de1d780e8d81de63377d1dae2dd6f873d996/src/docker.rs#L118-L124

Makes it a lot easier to use `Cross.toml` since you will not need wrapper scripts to define the environment variable value, or provide them every time on the command line.

Alexhuszagh:
Allow the following syntax:

```toml
[build.env]
passthrough = ["VAR=VALUE"]
```

Closes #743.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -115,9 +115,10 @@ pub fn run(
     let mut docker = docker_command("run")?;
 
     let validate_env_var = |var: &str| -> Result<()> {
-        if var.contains('=') {
-            bail!("environment variable names must not contain the '=' character");
-        }
+        let var = match var.split_once('=') {
+            Some((key, _)) => key,
+            _ => var,
+        };
 
         if var == "CROSS_RUNNER" {
             bail!("CROSS_RUNNER environment variable name is reserved and cannot be pass through");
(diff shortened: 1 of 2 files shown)


otavio (approved):

src/docker.rs
```diff
@@ -115,9 +115,10 @@ pub fn run(
     let mut docker = docker_command("run")?;
 
     let validate_env_var = |var: &str| -> Result<()> {
-        if var.contains('=') {
-            bail!("environment variable names must not contain the '=' character");
-        }
+        let var = match var.split_once('=') {
+            Some((key, _)) => key,
+            _ => var,
+        };
```
Emilgardis, line 121: the variable can still contain `=`, is that ok?

PR #749, opened 2022-06-03
fixup changelog

Emilgardis:


Emilgardis:
this should hopefully help with #662 a bit, since there are multiple sections meaning less change for conflicts

PR #754, opened 2022-06-04
Allow `Cargo.toml` as configuration source

mntns:
This addresses #657 and enables that `cross` can be configured via package metadata in the `Cargo.toml`.

### DIFF
--- a/src/bin/cross-util.rs
+++ b/src/bin/cross-util.rs
@@ -1,6 +1,6 @@
 #![deny(missing_debug_implementations, rust_2018_idioms)]
 
-use clap::{Parser, Subcommand};
+use clap::{CommandFactory, Parser, Subcommand};
 
 mod commands;
 
@@ -10,6 +10,13 @@ struct Cli {
     /// Toolchain name/version to use (such as stable or 1.59.0).
     #[clap(value_parser = is_toolchain)]
     toolchain: Option<String>,
+    #[clap(subcommand)]
+    command: Commands,
+}
+
+// hidden implied parser so we can get matches without recursion.
+#[derive(Parser, Debug)]
+struct CliHidden {
     #[clap(subcommand)]
     command: Commands,
 }
@@ -25,7 +32,8 @@ fn is_toolchain(toolchain: &str) -> cross::Result<String> {
     if toolchain.starts_with('+') {
         Ok(toolchain.chars().skip(1).collect())
     } else {
-        eyre::bail!("not a toolchain")
+        let _ = <CliHidden as CommandFactory>::command().get_matches();
+        unreachable!();
     }
 }
 
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -10,7 +10,7 @@ pub mod util;
 use std::path::PathBuf;
 
 use ci::CiJob;
-use clap::{Parser, Subcommand};
+use clap::{CommandFactory, Parser, Subcommand};
 use util::ImageTarget;
 
 use self::build_docker_image::BuildDockerImage;
@@ -24,6 +24,13 @@ struct Cli {
     /// Toolchain name/version to use (such as stable or 1.59.0).
     #[clap(value_parser = is_toolchain)]
     toolchain: Option<String>,
+    #[clap(subcommand)]
+    command: Commands,
+}
+
+// hidden implied parser so we can get matches without recursion.
+#[derive(Parser, Debug)]
+struct CliHidden {
     #[clap(subcommand)]
     command: Commands,
 }
@@ -49,7 +56,8 @@ fn is_toolchain(toolchain: &str) -> cross::Result<String> {
     if toolchain.starts_with('+') {
         Ok(toolchain.chars().skip(1).collect())
     } else {
-        eyre::bail!("not a toolchain")
+        let _ = <CliHidden as CommandFactory>::command().get_matches();
+        unreachable!();
     }
 }
 


Alexhuszagh (commented):
Excellent work, could you potentially quash the commits into 1 or 2, and then implement the above changes? I'm in favor, if any other reviewer approves, I'd happily merge this.

Specificially, these two commits sound great:

- Add parsing of CrossToml from Cargo.toml
- Add documentation about config in Cargo.toml

So one commit would just be the source files, and the other would just be the documentation, which should make it easy?

README.md
```diff
@@ -82,9 +82,12 @@ $ cross rustc --target powerpc-unknown-linux-gnu --release -- -C lto
 
 ## Configuration
 
-You can place a `Cross.toml` file in the root of your Cargo project or use a
-`CROSS_CONFIG` environment variable to tweak `cross`'s behavior. The format
-of `Cross.toml` is documented in [docs/cross_toml.md](docs/cross_toml.md).
+You have three options to configure `cross`:
+1. Write the configuration directly in your `Cargo.toml`, under `[package.metadata.cross]`
+2. Place a `Cross.toml` file in the root of your Cargo project
+3. Set the `CROSS_CONFIG` environment variable, in order to tell `cross` where it should search for the config
```
Emilgardis, line 88: it might be confusing with how you'll need to use package.metadata all the time in Cargo.toml, maybe that should be explained or demonstrated

Emilgardis (commented):
lgtm, minor nit

src/main.rs
```diff
@@ -487,27 +487,53 @@ pub(crate) fn warn_host_version_mismatch(
...
-    if path.exists() {
-        let content = file::read(&path)
-            .wrap_err_with(|| format!("could not read file `{}`", path.display()))?;
-
-        let (config, _) = CrossToml::parse(&content)
-            .wrap_err_with(|| format!("failed to parse file `{}` as TOML", path.display()))?;
+    // Attempts to read the cross config from the Cargo.toml
+    let cargo_toml_str =
+        file::read(root.path().join("Cargo.toml")).wrap_err("failed to read Cargo.toml")?;
+    let cross_toml_opt = CrossToml::parse_from_cargo(&cargo_toml_str)?;
```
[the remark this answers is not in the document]
otavio: Maybe `cargo_cross_toml`?

README.md
otavio: We could even promote its use so we can deprecate the old file and eventually remove it.

otavio (changes requested):

Jules-Bertholet:
> We could even promote its use so we can deprecate the old file and eventually remove it

Please don't deprecate `Cross.toml`; `Cargo.toml` is generally checked in to source control, but you may not always want your Cross configuration in your git repo.

otavio:
Considering what @Jules-Bertholet said, I agree with @Emilgardis that `Cargo.toml` should have higher priority.

Jules-Bertholet:
> I agree with @Emilgardis that Cargo.toml should have higher priority.

@otavio Did you mean "I agree with @Emilgardis that *Cross.toml* should have higher priority" ?

otavio:
@Jules-Bertholet yes. Sorry for the confusion. I edited my comment.

src/main.rs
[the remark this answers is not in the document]
mntns: Thx, I changed `cross_toml_opt` into `cargo_cross_toml_opt`.

README.md
mntns: I added some more documentation for this, including an example for the first configuration option.

mntns:
Thx, I squashed the comments (accidentally into one instead of two) and made the requested changes, including giving a slightly better overview of the configuration options.

Alexhuszagh:
Good idea. Anyway, here's the entire patch to fix the merge conflicts relative to the current main, or commit 23fd0af. Just save it to a file, reset hard to main, and then apply it. We should definitely make the changes Emil mentioned though prior to merging.

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 2156124..4042929 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #775 - forward Cargo exit code to host
 - #767 - added the `cross-util` and `cross-dev` commands.
+- #754 - Add `Cargo.toml` as configuration source
 - #745 - added `thumbv7neon-*` targets.
 - #741 - added `armv7-unknown-linux-gnueabi` and `armv7-unknown-linux-musleabi` targets.
 - #721 - add support for running doctests on nightly if `CROSS_UNSTABLE_ENABLE_DOCTESTS=true`.
diff --git a/README.md b/README.md
index 763265f..d6871b0 100644
--- a/README.md
+++ b/README.md
@@ -82,9 +82,27 @@ $ cross rustc --target powerpc-unknown-linux-gnu --release -- -C lto
 
 ## Configuration
 
-You can place a `Cross.toml` file in the root of your Cargo project or use a
-`CROSS_CONFIG` environment variable to tweak `cross`'s behavior. The format
-of `Cross.toml` is documented in [docs/cross_toml.md](docs/cross_toml.md).
+You have three options to configure `cross`. All of these options use the TOML format for configuration and the possible configuration values are documented [here](docs/cross_toml.md).
+
+### Option 1: Configuring `cross` directly in your `Cargo.toml`
+
+You can directly set [configuration values](docs/cross_toml.md) in your `Cargo.toml` file, under the `[package.metadata.cross]` table, i.e. key prefix.
+An example config snippet would look like this:
+
+```
+[package.metadata.cross.target.aarch64-unknown-linux-gnu]
+xargo = false
+image = "test-image"
+runner = "custom-runner"
+```
+
+### Option 2: Configuring `cross` via a `Cross.toml` file
+
+You can put your [configuration](docs/cross_toml.md) inside a `Cross.toml` file in your project root directory.
+
+### Option 3: Using `CROSS_CONFIG` to specify the location of your configuration
+
+By setting the `CROSS_CONFIG` environment variable, you can tell `cross` where it should search for the config file. This way you are not limited to a `Cross.toml` file in the project root.
 
 ### Custom Docker images
 
diff --git a/docs/cross_toml.md b/docs/cross_toml.md
index 662c541..58d390d 100644
--- a/docs/cross_toml.md
+++ b/docs/cross_toml.md
@@ -1,4 +1,6 @@
-The `cross` configuration in the `Cross.toml` file, can contain the following elements:
+The `cross` configuration in the `Cross.toml` file, can contain the elements described below.
+
+If the configuration is given in the `Cargo.toml`, these table headers must be of the form `[package.metadata.cross.<KEY>]`.
 
 # `build`
 The `build` key allows you to set global variables, e.g.:
@@ -38,4 +40,4 @@ This is similar to `build.env`, but allows you to be more specific per target.
 [target.x86_64-unknown-linux-gnu.env]
 volumes = ["VOL1_ARG", "VOL2_ARG"]
 passthrough = ["IMPORTANT_ENV_VARIABLES"]
-```
\ No newline at end of file
+```
diff --git a/src/cross_toml.rs b/src/cross_toml.rs
index fbddb93..f20c6a6 100644
--- a/src/cross_toml.rs
+++ b/src/cross_toml.rs
@@ -2,7 +2,7 @@
 
 use crate::errors::*;
 use crate::{Target, TargetList};
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 use std::collections::{BTreeSet, HashMap};
 
 /// Environment configuration
@@ -46,11 +46,33 @@ pub struct CrossToml {
 impl CrossToml {
     /// Parses the [`CrossToml`] from a string
     pub fn parse(toml_str: &str) -> Result<(Self, BTreeSet<String>)> {
-        let tomld = &mut toml::Deserializer::new(toml_str);
+        let mut tomld = toml::Deserializer::new(toml_str);
+        Self::parse_from_deserializer(&mut tomld)
+    }
 
-        let mut unused = BTreeSet::new();
+    /// Parses the [`CrossToml`] from a string containing the Cargo.toml contents
+    pub fn parse_from_cargo(cargo_toml_str: &str) -> Result<Option<(Self, BTreeSet<String>)>> {
+        let cargo_toml: toml::Value = toml::from_str(cargo_toml_str)?;
+        let cross_metadata_opt = cargo_toml
+            .get("package")
+            .and_then(|p| p.get("metadata"))
+            .and_then(|m| m.get("cross"));
+
+        if let Some(cross_meta) = cross_metadata_opt {
+            Ok(Some(Self::parse_from_deserializer(cross_meta.clone())?))
+        } else {
+            Ok(None)
+        }
+    }
 
-        let cfg = serde_ignored::deserialize(tomld, |path| {
+    /// Parses the [`CrossToml`] from a [`Deserializer`]
+    fn parse_from_deserializer<'de, D>(deserializer: D) -> Result<(Self, BTreeSet<String>)>
+    where
+        D: Deserializer<'de>,
+        D::Error: Send + Sync + 'static,
+    {
+        let mut unused = BTreeSet::new();
+        let cfg = serde_ignored::deserialize(deserializer, |path| {
             unused.insert(path.to_string());
         })?;
 
@@ -204,4 +226,56 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    pub fn parse_from_empty_cargo_toml() -> Result<()> {
+        let test_str = r#"
+          [package]
+          name = "cargo_toml_test_package"
+          version = "0.1.0"
+
+          [dependencies]
+          cross = "1.2.3"
+        "#;
+
+        let res = CrossToml::parse_from_cargo(test_str)?;
+        assert!(res.is_none());
+
+        Ok(())
+    }
+
+    #[test]
+    pub fn parse_from_cargo_toml() -> Result<()> {
+        let cfg = CrossToml {
+            targets: HashMap::new(),
+            build: CrossBuildConfig {
+                env: CrossEnvConfig {
+                    passthrough: vec![],
+                    volumes: vec![],
+                },
+                xargo: Some(true),
+                default_target: None,
+            },
+        };
+
+        let test_str = r#"
+          [package]
+          name = "cargo_toml_test_package"
+          version = "0.1.0"
+
+          [dependencies]
+          cross = "1.2.3"
+
+          [package.metadata.cross.build]
+          xargo = true
+        "#;
+
+        if let Some((parsed_cfg, _unused)) = CrossToml::parse_from_cargo(test_str)? {
+            assert_eq!(parsed_cfg, cfg);
+        } else {
+            panic!("Parsing result is None");
+        }
+
+        Ok(())
+    }
 }
diff --git a/src/lib.rs b/src/lib.rs
index c4c9808..4261926 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,7 +143,7 @@ impl<'a> From<&'a str> for Host {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
-#[serde(from = "&str")]
+#[serde(from = "String")]
 pub enum Target {
     BuiltIn { triple: String },
     Custom { triple: String },
@@ -269,9 +269,9 @@ impl From<Host> for Target {
     }
 }
 
-impl From<&str> for Target {
-    fn from(target_str: &str) -> Target {
-        let target_host: Host = target_str.into();
+impl From<String> for Target {
+    fn from(target_str: String) -> Target {
+        let target_host: Host = target_str.as_str().into();
         target_host.into()
     }
 }
@@ -500,27 +500,50 @@ pub(crate) fn warn_host_version_mismatch(
     Ok(VersionMatch::Same)
 }
 
-/// Parses the `Cross.toml` at the root of the Cargo project or from the
-/// `CROSS_CONFIG` environment variable (if any exist in either location).
+/// Obtains the [`CrossToml`] from one of the possible locations
+///
+/// These locations are checked in the following order:
+/// 1. Package metadata in the Cargo.toml
+/// 2. If the `CROSS_CONFIG` variable is set, it tries to read the config from its value
+/// 3. Otherwise, the `Cross.toml` in the project root is set.
 fn toml(metadata: &CargoMetadata) -> Result<Option<CrossToml>> {
-    let path = match env::var("CROSS_CONFIG") {
+    let root = &metadata.workspace_root;
+    let cross_config_path = match env::var("CROSS_CONFIG") {
         Ok(var) => PathBuf::from(var),
-        Err(_) => metadata.workspace_root.join("Cross.toml"),
+        Err(_) => root.join("Cross.toml"),
     };
 
-    if path.exists() {
-        let content = file::read(&path)
-            .wrap_err_with(|| format!("could not read file `{}`", path.display()))?;
-
-        let (config, _) = CrossToml::parse(&content)
-            .wrap_err_with(|| format!("failed to parse file `{}` as TOML", path.display()))?;
+    // Attempts to read the cross config from the Cargo.toml
+    let cargo_toml_str =
+        file::read(root.join("Cargo.toml")).wrap_err("failed to read Cargo.toml")?;
+    let cargo_cross_toml_opt = CrossToml::parse_from_cargo(&cargo_toml_str)?;
+
+    match (
+        cross_config_path.exists(),
+        cargo_cross_toml_opt.map(|(cfg, _)| cfg),
+    ) {
+        (true, _) => {
+            let content = file::read(&cross_config_path).wrap_err_with(|| {
+                format!("could not read file `{}`", cross_config_path.display())
+            })?;
+
+            let (config, _) = CrossToml::parse(&content).wrap_err_with(|| {
+                format!(
+                    "failed to parse file `{}` as TOML",
+                    cross_config_path.display()
+                )
+            })?;
+
+            Ok(Some(config))
+        }
+        (false, Some(cfg)) => Ok(Some(cfg)),
+        (false, None) => {
+            // Checks if there is a lowercase version of this file
+            if root.join("cross.toml").exists() {
+                eprintln!("There's a file named cross.toml, instead of Cross.toml. You may want to rename it, or it won't be considered.");
+            }
 
-        Ok(Some(config))
-    } else {
-        // Checks if there is a lowercase version of this file
-        if metadata.workspace_root.join("cross.toml").exists() {
-            eprintln!("There's a file named cross.toml, instead of Cross.toml. You may want to rename it, or it won't be considered.");
+            Ok(None)
         }
-        Ok(None)
     }
 }
```

mntns:
I just fixed the merge conflict and would now try implementing the merge mechanism as proposed by @Emilgardis.

Alexhuszagh:
If you can fix the clippy lint and squash the commits, that'd be great. I'll also give it another review shortly.

src/lib.rs
```diff
@@ -512,27 +512,54 @@ pub(crate) fn warn_host_version_mismatch(
...
 }
 
-/// Parses the `Cross.toml` at the root of the Cargo project or from the
-/// `CROSS_CONFIG` environment variable (if any exist in either location).
+/// Obtains the [`CrossToml`] from one of the possible locations
+///
+/// These locations are checked in the following order:
+/// 1. Package metadata in the Cargo.toml
+/// 2. If the `CROSS_CONFIG` variable is set, it tries to read the config from its value
+/// 3. Otherwise, the `Cross.toml` in the project root is set.
```
Alexhuszagh, line 520: I believe we're changing this order, see https://github.com/cross-rs/cross/pull/754#issuecomment-1151546469.

This way, `Cross.toml` has the highest priority, and overrides any settings provided in `Cargo.toml`. So it would be `CROSS_CONFIG` > `Cross.toml` > `Cargo.toml`.

src/lib.rs
```diff
@@ -512,27 +512,54 @@ pub(crate) fn warn_host_version_mismatch(
...
+    let cargo_toml_str =
+        file::read(root.join("Cargo.toml")).wrap_err("failed to read Cargo.toml")?;
+    let cargo_cross_toml_opt = CrossToml::parse_from_cargo(&cargo_toml_str)?;
+
+    match (
+        cross_config_path.exists(),
+        cargo_cross_toml_opt.map(|(cfg, _)| cfg),
+    ) {
+        (true, Some(_)) => {
+            eprintln!("Found a cross configuration (e.g. Cross.toml) and a configuration in your Cargo.toml. Please use only one of these options.");
```
Alexhuszagh, line 538: Since `Cargo.toml` is always checked into CI, but `Cross.toml` might not be, and we're not planning on deprecating either, this warning should probably be removed and we should concatenate both,

Alexhuszagh (changes requested):
Almost there, thanks for all the excellent work on this PR, I just think we should have the following:

First, change the doc comment here to:

```rust
/// Obtains the [`CrossToml`] from one of the possible locations
///
/// These locations are checked in the following order:
/// 1. If the `CROSS_CONFIG` variable is set, it tries to read the config from its value
/// 2. Otherwise, the `Cross.toml` in the project root is set.
/// 3. Package metadata in the `Cargo.toml`
///
/// The values from `CROSS_CONFIG` or `Cross.toml` are concatenated with the package
/// metadata in `Cargo.toml`. with `Cross.toml` having the highest priority.

Add code to merge the two. If you'd like, I can implement this, or feel free to as well. Just let me know.
```

src/cross_toml.rs
```diff
@@ -89,6 +90,52 @@ impl CrossToml {
...
+    /// # Merging of `targets`
+    /// The `targets` entries are merged based on the [`Target`] keys.
+    /// If a [`Target`] key is present in both configs, the [`CrossTargetConfig`]
+    /// in `other` overwrites the one in `self`.
+    ///
+    /// # Merging of `build`
+    /// The `build` fields ([`CrossBuildConfig`]) are merged based on their sub-fields.
+    /// A field in the [`CrossBuildConfig`] will only overwrite another if it contains
+    /// a value, i.e. it is not `None`.
+    pub fn merge(&self, other: &CrossToml) -> Result<CrossToml> {
```
Emilgardis, line 104: Couldn't this be implemented before deserializing the two/three data sources? 

e.g take the file contents, make `toml::Table`s, extend them, and then deserialize back.

If we serialize like this we lose unused keys, losing some information when we report back, ofc, this isn't a problem because we reported it already when making

src/cross_toml.rs
```diff
@@ -67,6 +90,52 @@ impl CrossToml {
...
```
Emilgardis, line 104:
```suggestion
    pub fn merge(self, other: CrossToml) -> Result<CrossToml> {
```

this should take owned values I think

src/cross_toml.rs
```diff
@@ -67,6 +90,52 @@ impl CrossToml {
...
+        let other_targets_map = to_map(&other.targets)?;
+        self_targets_map.extend(other_targets_map);
+        let merged_targets = from_map(self_targets_map)?;
+
+        // Merges build configs
+        let mut self_build_cfg_map = to_map(&self.build)?;
+        let mut other_build_cfg_map = to_map(&other.build)?;
+        other_build_cfg_map.retain(|_, v| !v.is_null());
+        self_build_cfg_map.extend(other_build_cfg_map);
+        let merged_build_cfg = from_map(self_build_cfg_map)?;
```
Emilgardis, line 131: can't this be unified?

they are both maps, and I assume maps can be placed in maps

mntns:
I added a mechanism for merging both config sources, and addressed your requested changes, @Alexhuszagh.

src/cross_toml.rs
```diff
@@ -49,11 +50,33 @@ pub struct CrossToml {
...
+            .get("package")
+            .and_then(|p| p.get("metadata"))
+            .and_then(|m| m.get("cross"));
+
+        if let Some(cross_meta) = cross_metadata_opt {
+            Ok(Some(Self::parse_from_deserializer(cross_meta.clone())?))
+        } else {
+            Ok(None)
+        }
+    }
```
Emilgardis, line 70: this is wrong, it should use `crate::cargo::cargo_metadata_with_args(Some(cwd), None, false)`

Emilgardis (changes requested):
looks good, I'm curious though if simply using `toml::Map` wouldn't work? in effect skipping the requirement for the contents to be serialize.

e.g before deserializing to `CrossToml`, make two `toml::Map`s and extend them with `cargo_toml.extend(cross_toml)`

src/cross_toml.rs
Emilgardis: discard that thought, it does already, my bad!

src/cross_toml.rs
Emilgardis: It might have to be done though now that I think about it, I assume that extending without being this specific will override every value inside `target`, even if that's not the intention


Instead, you'd probably have to descend into the map and merge. See https://docs.rs/serde-toml-merge/0.3.0/src/serde_toml_merge/lib.rs.html#58-63 for example

src/cross_toml.rs
Emilgardis: Unresolving this, you can actually use this fact for the better, instead of taking a string, add `CrossToml` to the `CargoMetadata`, not needed but it might help

mntns:
Just improved the merging mechanism and made it more general. I'll check out what to extend in the current `cargo` module for the configuration metadata to be in there.

Emilgardis (changes requested):

Emilgardis:
good job so far :D

src/lib.rs
```diff
@@ -512,27 +521,57 @@ pub(crate) fn warn_host_version_mismatch(
...
-            .wrap_err_with(|| format!("failed to parse file `{}` as TOML", path.display()))?;
+    // Attempts to read the cross config from the Cargo.toml
+    let cargo_toml_str =
+        file::read(root.join("Cargo.toml")).wrap_err("failed to read Cargo.toml")?;
+    let cargo_cross_toml_opt = CrossToml::parse_from_cargo(&cargo_toml_str)?;
+
+    match (
+        cross_config_path.exists(),
+        cargo_cross_toml_opt.map(|(cfg, _)| cfg),
+    ) {
```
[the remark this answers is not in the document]
Emilgardis: e.g

```toml
[package.metadata.cross.build]
foo = "bar"
```

should report that `build.foo` is not used

src/cross_toml.rs
```diff
@@ -234,4 +311,218 @@ mod tests {
...
+                runner: None,
+            },
+        );
+        targets_expected.insert(
+            Target::Custom {
+                triple: "target2".to_string(),
+            },
+            CrossTargetConfig {
+                env: CrossEnvConfig {
+                    passthrough: vec!["VAR2_PRECEDENCE".to_string()],
```
Alexhuszagh, line 485: I'm not sure this is how I would expect them to be merged.

`Cargo.toml` has the following:

```toml
[package.metadata.cross.target.target2.env]
passthrough = ["VAR2"]
volumes = ["VOL2_ARG"]
```

And `Cross.toml` has:

```toml
[target.target2.env]
passthrough = ["VAR2_PRECEDENCE"]
volumes = ["VOL2_ARG_PRECEDENCE"]
```

I would expect the result to be `["VAR2", "VAR2_PRECEDENCE"]` and `["VOL2", "VOL2_PRECEDENCE"]`, that is, non-scalar values get combined, and scalar values get overwritten.

`build-std` here is correct (it correctly overrides the `true` value with the `false` one, but I think the arrays should be merged, not over written.

src/cross_toml.rs
```diff
@@ -67,6 +90,60 @@ impl CrossToml {
...
+            let mut other_sub_map = other_map
+                .get(field)
+                .cloned()
+                .unwrap()
+                .as_object_mut()
+                .unwrap()
+                .to_owned();
+
+            // Only retain fields in the build config that have a value
+            if field == "build" {
```
Alexhuszagh, line 137: I think we can generalize this to also include `env`, and also handle arrays, since we might want an array value in `build` in the future or an optional value in `env` later on. That is, this should be:

```rust
if field == "build" || field == "env" {
    other_sub_map.retain(|_, v| !v.is_null());
}
```

Alexhuszagh (changes requested):
Very good work, some major improvements. Thank you. Just a few things I think should be changed.

Alexhuszagh:
I'll be updating the merging logic and I can commit it directly.

src/cross_toml.rs
Alexhuszagh: I've made the merging logic more generalizable [here](https://github.com/Alexhuszagh/cross/blob/cargo_toml_overwrite_arrays/src/cross_toml.rs#L120-L154), so it doesn't assume the exact structure of our config file which will make this code easier to maintain.

```rust
        // merge 2 objects. y has precedence over x.
        fn merge_objects(x: &mut ValueMap, y: &ValueMap) -> Option<()> {
            // we need to iterate over both keys, so we need a full deduplication
            let keys: BTreeSet<String> = x.keys().chain(y.keys()).cloned().collect();
            for key in keys {
                let in_x = x.contains_key(&key);
                let in_y = y.contains_key(&key);
                if !in_x && in_y {
                    let yk = y[&key].clone();
                    x.insert(key, yk);
                    continue;
                } else if !in_y {
                    continue;
                }


                let xk = x.get_mut(&key)?;
                let yk = y.get(&key)?;
                if xk.is_null() && !yk.is_null() {
                    *xk = yk.clone();
                    continue;
                } else if yk.is_null() {
                    continue;
                }


                // now we've filtered out missing keys and optional values
                // all key/value pairs should be same type.
                if xk.is_object() {
                    merge_objects(xk.as_object_mut()?, yk.as_object()?)?;
                } else {
                    *xk = yk.clone();
                }
            }


            Some(())
        }
```

I think this is an improvement since it will work no matter how our config file is structured, no matter how many tables or fields we add.

Emilgardis:
with either, is it possible to say make `target.target2.env.volumes` completely empty?

Alexhuszagh:
> with either, is it possible to say make `target.target2.env.volumes` completely empty?

Actually, this might need some refactoring, because I'm pretty sure how we structure the config file right now doesn't differentiate between no key for `volumes` being present and `volumes = []`: I think it treats both as the same.

Alexhuszagh:
Now that #801 has been merged, merging the two config files and properly overwriting arrays is now possible. So the strategy I think we've agreed on is:

1. `CROSS_CONFIG` > `Cross.toml` > `Cargo.toml`.
2. Merge into objects
3. Null values are overwritten by non-null values
4. If both are not-null, any non-object types are overwritten.

That is, given the two sample files:

This is my sample `Cargo.toml` file:

```toml
[package.metadata.cross.build]
xargo = true

[package.metadata.cross.target.target2]
build-std = true

[package.metadata.cross.target.target2.env]
passthrough = ["VAR2"]
volumes = ["VOL2_ARG"]
```

And this is my sample `Cross.toml` file:

```toml
[target.target2]
image = "image2:tag"

[target.target2.env]
passthrough = ["VAR2_PRECEDENCE"]
volumes = ["VOL2_ARG_PRECEDENCE"]
```

The output should be:

```toml
[build]
xargo = true

[target.target2]
build-std = true
image = "image2:tag"

[target.target2.env]
passthrough = ["VAR2_PRECEDENCE"]
volumes = ["VOL2_ARG_PRECEDENCE"]
```

mntns:
I like the recursive merging a lot, @Alexhuszagh, nice solution! I think merging the `build` configuration recursively makes sense, but for the `targets` I'd find it personally more intuitive if it merges them on the given target (key) and does not recurse into e.g. `env`. But obviously this might be different for other users.

Emilgardis:
> I think we need consistency, or else things get convoluted pretty quickly. And we need a way to be able to "unset" variables provided in `Cargo.toml`. @Emilgardis does the recursive merging, overwrite solution sound good?

Yes, it seems like the best approach

naiba:
Hi everyone, is there any update? The new version release seems to be stuck here? How long will it take?

naiba:
Hi everyone, is there any update? The new version release seems to be stuck here? How long will it take?

src/cross_toml.rs
```diff
@@ -49,11 +50,33 @@ pub struct CrossToml {
...
     pub fn parse(toml_str: &str) -> Result<(Self, BTreeSet<String>)> {
-        let tomld = &mut toml::Deserializer::new(toml_str);
+        let mut tomld = toml::Deserializer::new(toml_str);
+        Self::parse_from_deserializer(&mut tomld)
+    }
 
-        let mut unused = BTreeSet::new();
+    /// Parses the [`CrossToml`] from a string containing the Cargo.toml contents
+    pub fn parse_from_cargo(cargo_toml_str: &str) -> Result<Option<(Self, BTreeSet<String>)>> {
+        let cargo_toml: toml::Value = toml::from_str(cargo_toml_str)?;
```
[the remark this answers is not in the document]
mntns: Done :+1:

mntns:
I didn't have time during the past few days, but integrated the new merging code by @Alexhuszagh now (thanks!) and improved the parse function signatures that @Emilgardis proposed.

As far as I'm aware, the only thing that is missing is that `CargoMetadata` should be used? I can work on this tomorrow, @naiba. So if it passes the review, I hope that this won't block a release for longer.

src/cross_toml.rs
```diff
@@ -227,11 +332,225 @@ mod tests {
...
+            assert_eq!(parsed_cfg, cfg);
+        } else {
+            panic!("Parsing result is None");
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    pub fn merge() -> Result<()> {
```
Emilgardis, line 397: Tests are great, but this is a bit too much in my opinion.

I think the better approach is to parse the structs as strings, that way we skip having to update these whenever a field is added to the config structs and it's smaller in length.

src/cross_toml.rs
```diff
@@ -67,6 +101,77 @@ impl CrossToml {
...
+            }
+
+            Some(())
+        }
+
+        // Builds maps of objects
+        let mut self_map = to_map(&self)?;
+        let other_map = to_map(other)?;
+
+        merge_objects(&mut self_map, &other_map).unwrap();
```
Emilgardis, line 171:
```suggestion
        merge_objects(&mut self_map, &other_map).ok_or_else(|| eyre::eyre!("could not merge"));
```

src/cross_toml.rs
```diff
@@ -67,6 +101,77 @@ impl CrossToml {
...
+                Ok(obj.to_owned())
+            } else {
+                panic!("Failed to serialize CrossToml as object");
+            }
+        }
+
+        fn from_map<D: DeserializeOwned>(map: ValueMap) -> Result<D> {
+            let value = serde_json::to_value(map)?;
+            Ok(serde_json::from_value(value)?)
+        }
```
Emilgardis, line 129: most of these should be annotated with `wrap_err("could not ...")`

specifically the `serde_json::from_value`

src/cross_toml.rs
```diff
@@ -67,6 +101,77 @@ impl CrossToml {
...
+    /// A field in the [`CrossBuildConfig`] will only overwrite another if it contains
+    /// a value, i.e. it is not `None`.
+    pub fn merge(self, other: CrossToml) -> Result<CrossToml> {
+        type ValueMap = serde_json::Map<String, serde_json::Value>;
+
+        fn to_map<S: Serialize>(s: S) -> Result<ValueMap> {
+            if let Some(obj) = serde_json::to_value(s)?.as_object() {
+                Ok(obj.to_owned())
+            } else {
+                panic!("Failed to serialize CrossToml as object");
```
Emilgardis, line 122: This should not panic, but return a result
```suggestion
                eyre::bail!("failed to serialize CrossToml as object")
```

note also lowercase "f"

src/cross_toml.rs
```diff
@@ -227,11 +332,225 @@ mod tests {
...
+                    passthrough: vec!["VAR3".to_string(), "VAR4".to_string()],
+                    volumes: vec![],
+                },
+                build_std: Some(true),
+                xargo: Some(false),
+                default_target: Some("aarch64-unknown-linux-gnu".to_string()),
+            },
+        };
+
+        let cfg_merged = cfg1.merge(cfg2).unwrap();
```
Emilgardis, line 551: this test uses `fn() -> Result<..,..>`, so this could use `?` instead of unwrap

Emilgardis (changes requested):

Emilgardis:
im excited for this, thanks @mntns for doing this!

mntns:
@Emilgardis, just pulled your `pr/mergetoml754` and force-pushed it :+1:

PR #755, opened 2022-06-04
Fix docker detection.

### LINKED ISSUE #530: Failure when podman-docker is installed

Alexhuszagh:
Currently, docker detection fails under a few cases:
1. If `podman_docker` is installed, so the filename is docker but the actual executable is podman.
2. If the executable has a suffix, such as `.exe` on Windows, because we check if the executable `ends_with(DOCKER)`.

The only reliable way to fix both these issues, IE, if the actual engine is docker and not an alias, and if the executable does not contain a suffix, is to query the container engine. This might not be ideal for performance reasons, but is the only reliable way to fix these issues.

Closes #530.
Closes #538.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -15,6 +15,17 @@ const CROSS_IMAGE: &str = "ghcr.io/cross-rs";
 const DOCKER: &str = "docker";
 const PODMAN: &str = "podman";
 
+// determine if the container engine is docker. this fixes issues with
+// any aliases (#530), and doesn't fail if an executable suffix exists.
+fn get_is_docker(ce: std::path::PathBuf, verbose: bool) -> Result<bool> {
+    let stdout = Command::new(ce)
+        .arg("--help")
+        .run_and_get_stdout(verbose)?
+        .to_lowercase();
+
+    Ok(stdout.contains("docker") && !stdout.contains("emulate"))
+}
+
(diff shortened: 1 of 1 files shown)


src/docker.rs
```diff
@@ -126,6 +136,7 @@ pub fn run(
     let runner = config.runner(target)?;
 
     let mut docker = docker_command("run")?;
+    let is_docker = get_is_docker(get_container_engine().unwrap(), verbose)?;
```
Alexhuszagh, line 139: Note that this can't fail due to the above command: `docker_command` ensures that `get_container_engine` is not an error.

Emilgardis (approved):
bors r+

Issue #758, opened 2022-06-06
error: linker `i686-w64-mingw32-gcc` not found


brandonros:
```
 $ CROSS_CONTAINER_ENGINE=docker cross build --target i686-pc-windows-gnu
   Compiling vpi v0.0.1 (/Users/brandonros/Desktop/dpdu-http/vpi)
   Compiling remoter v0.0.1 (/Users/brandonros/Desktop/dpdu-http/remoter)
error: linker `i686-w64-mingw32-gcc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `vpi` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `remoter` due to previous error

Brandons-MacBook-Air:dpdu-http brandonros 2022-06-06 13:25:44 $ uname -a
Darwin Brandons-MacBook-Air.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:47:26 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T8101 arm64
```

It feels like it isn't running in Docker?

Alexhuszagh:
Can you provide the output the cross version? Are you installing from git? The latest cross versions will also warn you if you fall back to the host.

Alexhuszagh:
> ```
> $ cross --version
> cross 0.2.1
> cargo 1.61.0 (a028ae42f 2022-04-29)
> ```

Yeah that means you're using the very old version installed from crates.io, which has some major bugs we're fixing in a new release. For a while, we had a lack of maintainers, so we've been patching some errors that had built up over time. If you reinstall cross from git (as shown above), this should fix your issue. You can always use a specific commit if you'd like to lock version dependencies. For example, to use the currently latest commit, and ensure you've audited the correct version, use:

```bash
cargo install --git https://github.com/cross-rs/cross --rev 6a7fa09d6307acb5380475b134ff749c7cd9ed35
```

Thanks for your patience with the lack of recent releases, due to the lack of maintainers.

PR #759, opened 2022-06-06
update readme to mention install from git due to outdated crates version

brandonros:
part of https://github.com/cross-rs/cross/issues/758?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ default to `docker`.
 ## Installation
 
 ```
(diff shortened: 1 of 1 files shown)


Alexhuszagh (approved):

Alexhuszagh:
I think this is good, because of the major changes since the last release, which includes numerous bug fixes and should have no breaking changes so far.
bors r+

PR #762, opened 2022-06-07
Re-enable x86_64-unknown-dragonfly target.

Alexhuszagh:
This undoes #396, and updates the dragonfly version from 4.6.1 to 6.0.1.
This also fixes a few minor issues in the image previously:

- Brace expansion isn't quoted, so it properly expands.
- Archive files are downloaded to the temp directory, and deleted with
  the image.


Alexhuszagh:
bors try --target x86_64-unknown-dragonfly

PR #767, opened 2022-06-08
Add cross-dev and cross-util commands.

### LINKED ISSUE #760: Add subcommand to cleanup images?
### Checklist

### Describe your request

Currently, users have to be proficient with docker or podman to cleanup images, when we should be able to remove any installed images if desired from a cross subcommand.

For example:

```bash
cross remove-images
```

This would then filter for all images starting with `ghcr.io/cross-rs`, and `rustembedded/cross`, and then remove all the matching images. We could also accept a list of targets, or empty for all, to remove all matching targets. For the former, we would want to filter for those matching `main` or `edge`, to avoid removing local images.

For example:

```bash
cross remove-images aarch64-unknown-linux-gnu arm-unknown-linux-gnueabi mips-unknown-linux-gnu
```

We could probably at the same time provide a `list-images` subcommand, to list what images have been previously installed, using the same syntax.

```bash
$ cross list-images
aarch64-unknown-linux-gnu
arm-unknown-linux-gnueabi
mips-unknown-linux-gnu
```

This would further simplify using cross as a "zero-setup" cross-compilation tool, and also avoid users having to manually cleanup numerous images or write a script to automatically remove all images to reduce the storage requirements of cross.

Alexhuszagh:
The cross-dev command has the `target-info` subcommand, which target information (libc, compiler, C++, qemu version) for a given target.

```bash
cross-dev target-info
cross-dev target-info arm-unknown-linux-gnueabihf
```

The cross-util command has the `list-images` and `remove-images` subcommands. `list-images` lists all cross images, including local development files. `remove-images` removes cross images.

```bash
cross-util list-images
cross-util remove-images
cross-util remove-images arm-unknown-linux-gnueabihf
```

Closes #760.


Alexhuszagh:
I've separated this into many different commits, since this is a large PR and refactors some logic. Each I think is a relatively independent commit, and some of them (such as the removal of the duplicate `cpp` tags) is crucial for subsequent steps (`serde_yaml` parsing the matrix).

src/bin/cross-util.rs
```diff
@@ -0,0 +1,193 @@
...
+        verbose: bool,
+        /// Force removal of images.
+        #[clap(short, long)]
+        force: bool,
+        /// Remove local (development) images.
+        #[clap(short, long)]
+        local: bool,
+        /// Container engine (such as docker or podman).
+        #[clap(long)]
+        engine: Option<String>,
```
Emilgardis, line 49: this and the previous one should also be a env var `CROSS_CONTAINER_ENGINE`

src/bin/cross-util.rs
```diff
@@ -0,0 +1,193 @@
...
+        cross::get_container_engine()
+    }
+}
+
+fn is_cross_image(image: &str) -> bool {
+    IMAGE_PREFIXES.iter().any(|i| image.starts_with(i))
+}
+
+fn is_local_image(image: &str) -> bool {
+    // this cannot panic: we've formatted our image as ${repo}:${tag}
```
Emilgardis, line 66:
```suggestion
    // this cannot panic: we've formatted our image list as ${repo}:${tag}
```

src/bin/cross-util.rs
```diff
@@ -0,0 +1,193 @@
...
+            force,
+            local,
+            engine,
+        } => {
+            let engine = get_container_engine(engine)?;
+            if targets.is_empty() {
+                remove_all_images(&engine, *verbose, *force, *local)?;
+            } else {
+                remove_target_images(&engine, targets, *verbose, *force, *local)?;
+            }
```
Emilgardis, line 188: i think there should be a confirm prompt here, or, a we have a `--execute` flag and standard operation without it is a dry-run

src/bin/cross-util.rs
```diff
@@ -0,0 +1,193 @@
...
+        /// Remove local (development) images.
+        #[clap(short, long)]
+        local: bool,
+        /// Container engine (such as docker or podman).
+        #[clap(long)]
+        engine: Option<String>,
+    },
+}
+
+fn get_container_engine(engine: &Option<String>) -> Result<PathBuf, which::Error> {
```
Emilgardis, line 53:
```suggestion
fn get_container_engine(engine: Option<&str>) -> Result<PathBuf, which::Error> {
```

`Option` has a `as_ref` implementation

src/bin/cross-util.rs
```diff
@@ -0,0 +1,193 @@
...
+    images.sort();
+
+    Ok(images)
+}
+
+// the old rustembedded targets had the following format:
+//  repository = (${registry}/)?rustembedded/cross
+//  tag = ${target}(-${version})?
+// our target triple must match `[A-Za-z0-9_-]`
+fn rustembedded_target(tag: &str) -> Option<String> {
```
Emilgardis, line 93: why does this return an option?

Cargo.toml
```diff
@@ -8,9 +8,11 @@ name = "cross"
 repository = "https://github.com/cross-rs/cross"
 version = "0.2.1"
 edition = "2021"
+readme = "docs/crates_io.md"
```
Emilgardis, line 11: why?

cross-dev/Cargo.lock
```diff
@@ -0,0 +1,621 @@
+# This file is automatically @generated by Cargo.
```
Emilgardis, line 1: we should not have multiple lock files imo, instead `cross-dev` should be included in the root workspace, preferably also then without being in the default-members

cross-dev/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+description = "Development utilities for cross"
+documentation = "https://github.com/cross-rs/cross"
+keywords = ["cross", "compilation", "testing", "tool"]
+license = "MIT OR Apache-2.0"
+name = "cross-dev"
+repository = "https://github.com/cross-rs/cross"
+version = "0.0.1"
+edition = "2021"
+readme = "../docs/crates_io.md"
```
Emilgardis, line 10:
```suggestion
edition = "2021"
publish = false
```

cross-dev/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+clap = { version = "3.1.18", features = ["derive"] }
+cross = { path = "..", version = "0.2.1" }
```
Emilgardis, line 14:
```suggestion
cross = { path = ".." }
```

cross-dev/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+struct Matrix {
+    target: String,
+    os: String,
+    #[serde(default)]
+    cpp: i64,
+    #[serde(default)]
+    std: i64,
+    #[serde(default)]
+    run: i64,
+    runners: Option<String>,
```
Emilgardis, line 90: are these values used?

Emilgardis (changes requested):

src/bin/cross-util.rs
Alexhuszagh: It should not. I had originally planned some checking if the triple had <3 components, but I don't think that's a great idea anymore (say, a custom image someone also wants to remove).

Cargo.toml
Alexhuszagh: This is so when published on `crates.io`, it ensures that the README will display [this](https://github.com/Alexhuszagh/cross/blob/14857a333d6ad81e04b0af89aa9a3573bca07193/docs/crates_io.md). On `docs.rs`, it will look similar to the following image:

Screenshot_20220608_091729

cross-dev/src/main.rs
Alexhuszagh: They are not: I considered if they would and they should be removed.

cross-dev/Cargo.lock
Emilgardis: no, sadly you can't. :/

you can't `use bin;` if `bin` is a binary target, you need to have a `lib.rs` to use it in other crates, even if those crates are in the same workspace

Cargo.toml
Emilgardis: the warning would be nice to have in the documentation, but for crates.io it's fine to only have the regular readme I think.

It's not needed to specify README.md as the readme in the manifest, see https://github.com/cross-rs/cross/pull/563

src/bin/cross-util.rs
Emilgardis: https://docs.rs/clap/3.1.18/clap/struct.Arg.html#method.env

Alexhuszagh: How would this work then with `cross::get_container_engine`?

Emilgardis: ah, I see, nevermind then :D

Alexhuszagh:
Is this better now? I incorporated some of the changes, but thought it would be a bit better to provide a `[[bin]]` target for `cross-dev`, since then it can be run easily but disabled by default.

The changes I've made include:
- Moving `cross-dev` to `/cross-dev/main.rs`, and using the main `Cargo.toml` for it. No more `Cargo.lock`, no more need to not publish it.
- Make `cross-dev` a `[[bin]]` target.Since we need `serde_yaml` as a dependency for it, I've changed it so it's only compiled if the `dev` feature is enabled.
- `engine` still uses the same logic. I considered using `default_value_t` and `parse` for the clap options, but it was going to be a lot of code and would have probably been less readable than the existing implementation. This is because we'd have to provide a wrapper for `which::which`, then use `default_value_t = cross::get_container_engine()`.
- Changed the image output to `{{.Repository}}:{{.Tag}} {{.ID}}`. This is then parsed to an `Image` (see below). This is because images may be tagged as `${repository}:<none>`, which would cause on error on removing the image. Removing it by ID works in this case.
- I've added the `--execute` flag, so it's a dry-run by default.
- I've removed the unnecessary `Matrix` fields.
- I've changed `get_container_engine` to take an `Option<&str>`.
- `rustembedded_target` no longer returns an option.
- No longer use a custom readme in `Cargo.toml`. `lib.rs` still warns about this for crates.io.

```rust
#[derive(Debug, PartialOrd, Ord, PartialEq, Eq)]
struct Image {
    repository: String,
    tag: String,
    // need to remove images by ID, not just tag
    id: String,
}
```

docs/crates_io.md
```diff
@@ -0,0 +1,11 @@
+[![crates.io](https://img.shields.io/crates/v/cross.svg)](https://crates.io/crates/cross)
```
Emilgardis, line 1: this file should be renamed, it can even be ommited fully and just put in the lib.rs directly

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![doc = include_str!("../docs/crates_io.md")]
```
Emilgardis, line 1: as mentioned, this could be inlined

src/bin/cross-util.rs
```diff
@@ -0,0 +1,229 @@
...
+
+fn get_image_target(image: &Image) -> cross::Result<String> {
+    if let Some(stripped) = image.repository.strip_prefix(GHCR_IO) {
+        Ok(stripped.to_string())
+    } else if let Some(tag) = image.tag.strip_prefix(RUST_EMBEDDED) {
+        Ok(rustembedded_target(tag))
+    } else if let Some(tag) = image.tag.strip_prefix(DOCKER_IO) {
+        Ok(rustembedded_target(tag))
+    } else {
+        eyre::bail!("Cannot get target for image {}", image.name())
```
Emilgardis, line 139:
```suggestion
        eyre::bail!("cannot get target for image {}", image.name())
```

Emilgardis (changes requested):

.github/workflows/ci.yml
```diff
@@ -249,6 +249,10 @@ jobs:
         if: matrix.deploy
         run: cargo install --path . --force
 
+      - name: Install cross-dev
+        if: matrix.deploy
+        run: cargo install --path cross-dev --force
```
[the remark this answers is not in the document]
Emilgardis: specifying `--bin cross-dev` should be done too, to avoid unsync between needed features and the ci stage

Emilgardis (changes requested):

Emilgardis (approved):
bors r+

Alexhuszagh:
> bors r-
> 
> oh yeah you wanted to squash?

Already did.

bors r=Emilgardis

PR #768, opened 2022-06-08
Update README for target information.

Alexhuszagh:
Updated target information using the `target-info` command from `cross-dev`. Also added test detection for `cross-dev`, and fixed a bug with `powerpc64le-unknown-linux-gnu` using the wrong qemu architecture. I also removed targets for which we don't provide images, these are commented out below the table, so if we re-enable them they aren't gone entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,74 +272,75 @@ terminate.
 
 | Target                               |  libc  |   GCC   | C++ | QEMU  | `test` |
 |--------------------------------------|-------:|--------:|:---:|------:|:------:|
-| `*-apple-ios` [1]                    | N/A    | N/A     | N/A | N/A   |   ✓    |
-| `aarch64-linux-android` [2]          | 9.0.0  | 4.9     | ✓   | 5.1.0 |   ✓    |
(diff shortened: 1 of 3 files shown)


Alexhuszagh:
We probably could remove some of these footnotes and remove some of the commented-out targets entirely.

This isn't due to cross, since we don't provide an image and is just handled by the host toolchain.

```markdown
[1] iOS cross compilation is supported on macOS hosts.
```

This can just say no std component since I hope no one is using a compiler pre-2018 anymore, which will get duplicated if we ever add back `x86_64-sun-solaris`, which is also tier 3 now.

```markdown
[3] No `std` component available as of 2017-01-10
```

We can comment this out currently as well until we restore the emscripten targets:

```markdown
[6] libc = musl, gcc = emcc. The Docker images for these targets are currently not built automatically
due to a [compiler bug](https://github.com/rust-lang/rust/issues/85821), you will have to build them yourself for now.
```

So we should be able to remove `[1]`, shift everything else down by 1, modify `[3]`, and then comment out `[6]`.

otavio (approved):

Emilgardis:
missed the update! lgtm

bors r=otavio

PR #771, opened 2022-06-09
Fix parsing of docker options.

Alexhuszagh:
Fix parsing `DOCKER_OPTS` to correctly parse them as shell words, rather than splitting on whitespace.

Closes #769.

PR #772, opened 2022-06-09
Add `CROSS_CONTAINER_OPTS` environment variable.

### LINKED ISSUE #770: Rename `DOCKER_OPTS` to `CROSS_CONTAINER_OPTS`
### Checklist

### Describe your request

Currently, we don't use Docker in all instances (I almost entirely use podman), but to provide custom options have to define `DOCKER_OPTS`. Slightly more general, and namespaced, would be `CROSS_CONTAINER_OPTS`. For backwards compatibility, we could support both for a while, preferring the latter, then the former, and then a few releases later warn the user if they're still using `DOCKER_OPTS`.

_No response_

Alexhuszagh:
Rename `DOCKER_OPTS` to `CROSS_CONTAINER_OPTS`, and currently prefer `CROSS_CONTAINER_OPTS` to `DOCKER_OPTS`, although both are still valid.

Closes #770.

### DIFF
--- a/src/docker.rs
+++ b/src/docker.rs
@@ -94,6 +94,10 @@ fn validate_env_var(var: &str) -> Result<(&str, Option<&str>)> {
     Ok((key, value))
 }
 
+fn parse_docker_opts(value: &str) -> Result<Vec<String>> {
+    shell_words::split(value).wrap_err_with(|| format!("could not parse docker opts of {}", value))
+}
+
 #[allow(unused_variables)]
 pub fn mount(cmd: &mut Command, val: &Path, verbose: bool) -> Result<PathBuf> {
     let host_path = file::canonicalize(&val)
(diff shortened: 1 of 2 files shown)


Alexhuszagh:
Currently a draft since #771 will require changing how we parse the values here.

src/docker.rs
```diff
@@ -311,11 +315,14 @@ pub fn run(
...
-        let opts = shell_words::split(&value)
-            .wrap_err_with(|| format!("could not parse docker opts of {}", value))?;
-        docker.args(&opts);
-    }
+    if let Ok(value) = env::var("CROSS_CONTAINER_OPTS") {
+        if env::var("DOCKER_OPTS").is_ok() {
+            eprintln!("Warning: using both `CROSS_CONTAINER_OPTS` and `DOCKER_OPTS`.");
+        }
+        docker.args(&parse_docker_opts(&value)?);
+    } else if let Ok(value) = env::var("DOCKER_OPTS") {
```
otavio, line 323: We could warn users about the old environment variable and suggest using the new one. So we could eventually remove or error out.

otavio (changes requested):

src/docker.rs
Emilgardis: no need to error out, think it's fine to silently remove it after a few releases

otavio: It'd be nice to add a comment in code so we can easily find it in future.

Alexhuszagh: I've added a FIXME which will let us know to remove it upon deprecating `DOCKER_OPTS`.

https://github.com/cross-rs/cross/pull/772/files#diff-4cf895b046f53b02266b843eb9440dec88b857e039a624a5310502e3dd7eb858R304-R306

Emilgardis (approved):
bors r+

Emilgardis:
ping @otavio

otavio (approved):

otavio:
bors r=Emilgardis

Issue #773, opened 2022-06-09
Incorrect Environment Variable/Config Logic


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

The code for parsing some environment variable logic (for example, `xargo`) is as follows:

```rust
    pub fn xargo(&self, target: &Target) -> Result<Option<bool>> {
        let (build_xargo, target_xargo) = self.env.xargo(target);
        let (toml_build_xargo, toml_target_xargo) = if let Some(ref toml) = self.toml {
            toml.xargo(target)
        } else {
            (None, None)
        };

        match (build_xargo, toml_build_xargo) {
            (Some(xargo), _) => return Ok(Some(xargo)),
            (None, Some(xargo)) => return Ok(Some(xargo)),
            (None, None) => {}
        };

        match (target_xargo, toml_target_xargo) {
            (Some(xargo), _) => return Ok(Some(xargo)),
            (None, Some(xargo)) => return Ok(Some(xargo)),
            (None, None) => {}
        };
        Ok(None)
    }
```

This means that if `build.env.xargo = true`, or `CROSS_BUILD_XARGO`, these will override the settings present in `target.$(...).env.xargo` and `CROSS_TARGET_$(...)_XARGO`. We should prefer more specific to less specific: if someone wants to enable `xargo` for all but a single target, it should respect that.

For example:

```toml
[build.env]
xargo = true

[target.aarch64-unknown-linux-gnu.env]
xargo = false
```

This means we should use `xargo` for all targets **but** `aarch64-unknown-linux-gnu`, but we currently use it for all targets.

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [ ] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [ ] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [ ] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.1 (ee2fc1b 2022-06-09)

PR #774, opened 2022-06-09
Simplify config internals.

Alexhuszagh:
Simplifies the `config.rs` and `cross_toml.rs` to have more reusable functions, and implement logic in terms of these functions, so we can add more config options in the future. This also simplifies maintenance, since we can use 1-line logic for most config variables now.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,8 +1,8 @@
 use crate::{CrossToml, Result, Target, TargetList};
 
-use crate::errors::*;
 use std::collections::HashMap;
 use std::env;
+use std::str::FromStr;
 
 #[derive(Debug)]
 struct Environment(&'static str, Option<HashMap<&'static str, &'static str>>);
@@ -23,6 +23,19 @@ impl Environment {
             .or_else(|| env::var(name).ok())
     }
 
+    fn get_values_for<T>(
+        &self,
+        var: &str,
+        target: &Target,
+        convert: impl Fn(&str) -> T,
+    ) -> (Option<T>, Option<T>) {
+        let target_values = self.get_target_var(target, var).map(|ref s| convert(s));
+
+        let build_values = self.get_build_var(var).map(|ref s| convert(s));
+
+        (build_values, target_values)
+    }
+
     fn target_path(target: &Target, key: &str) -> String {
         format!("TARGET_{target}_{key}")
     }
@@ -39,27 +52,8 @@ impl Environment {
         self.get_var(&self.build_var_name(&Self::target_path(target, key)))
     }
 
-    fn xargo(&self, target: &Target) -> Result<(Option<bool>, Option<bool>)> {
-        let (build_xargo, target_xargo) = (
-            self.get_build_var("XARGO"),
-            self.get_target_var(target, "XARGO"),
-        );
-        let build_env = if let Some(value) = build_xargo {
-            Some(value.parse::<bool>().wrap_err_with(|| {
-                format!("error parsing {value} from XARGO environment variable")
-            })?)
-        } else {
-            None
-        };
-        let target_env = if let Some(value) = target_xargo {
-            Some(value.parse::<bool>().wrap_err_with(|| {
-                format!("error parsing {} from XARGO environment variable", value)
-            })?)
-        } else {
-            None
-        };
-
-        Ok((build_env, target_env))
+    fn xargo(&self, target: &Target) -> (Option<bool>, Option<bool>) {
+        self.get_values_for("XARGO", target, bool_from_envvar)
     }
 
     fn image(&self, target: &Target) -> Option<String> {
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -23,6 +23,19 @@ impl Environment {
             .or_else(|| env::var(name).ok())
     }
 
+    fn get_values_for<T, Convert: Fn(&str) -> T>(
+        &self,
+        var: &str,
+        target: &Target,
+        convert: Convert,
+    ) -> (Option<T>, Option<T>) {
```
Emilgardis, line 31:
```suggestion
    fn get_values_for<T>(
        &self,
        var: &str,
        target: &Target,
        convert: impl Fn(&str) -> T,
    ) -> (Option<T>, Option<T>) {
```

src/config.rs
```diff
@@ -136,72 +124,113 @@ impl Config {
...
+    fn bool_from_config<Env, Config>(
+        &self,
+        target: &Target,
+        env: Env,
+        config: Config,
+    ) -> Option<bool>
+    where
+        Env: Fn(&Environment, &Target) -> (Option<bool>, Option<bool>),
+        Config: Fn(&CrossToml, &Target) -> (Option<bool>, Option<bool>),
+    {
```
Emilgardis, line 136:
```suggestion
    fn bool_from_config(
        &self,
        target: &Target,
        env: impl Fn(&Environment, &Target) -> (Option<bool>, Option<bool>),
        config: impl Fn(&CrossToml, &Target) -> (Option<bool>, Option<bool>),
    ) -> Option<bool> {
```

src/config.rs
```diff
@@ -136,72 +124,113 @@ impl Config {
...
+    fn string_from_config<Env, Config>(
+        &self,
+        target: &Target,
+        env: Env,
+        config: Config,
+    ) -> Result<Option<String>>
+    where
+        Env: Fn(&Environment, &Target) -> Option<String>,
+        Config: Fn(&CrossToml, &Target) -> Option<String>,
+    {
```
Emilgardis, line 168: ditto

src/config.rs
```diff
@@ -136,72 +124,113 @@ impl Config {
...
+        target: &Target,
+        env: Env,
+        config_build: ConfigBuild,
+        config_target: ConfigTarget,
+    ) -> Result<Vec<String>>
+    where
+        Env: Fn(&Environment, &Target) -> (Option<Vec<String>>, Option<Vec<String>>),
+        ConfigBuild: Fn(&CrossToml) -> Vec<String>,
+        ConfigTarget: Fn(&CrossToml, &Target) -> Vec<String>,
+    {
```
Emilgardis, line 189: ditto

src/cross_toml.rs
```diff
@@ -116,6 +111,37 @@ impl CrossToml {
     fn get_target(&self, target: &Target) -> Option<&CrossTargetConfig> {
         self.targets.get(target)
     }
+
+    fn get_string<Get>(&self, target: &Target, get: Get) -> Option<String>
+    where
+        Get: Fn(&CrossTargetConfig) -> &Option<String>,
+    {
+        self.get_target(target).and_then(|t| get(t).clone())
```
Emilgardis, line 119: I'd love to get rid of this clone

src/cross_toml.rs
```diff
@@ -116,6 +111,37 @@ impl CrossToml {
...
+
+        (build, target)
+    }
+
+    fn get_vec<Get>(&self, target: &Target, get: Get) -> Vec<String>
+    where
+        Get: for<'a> Fn(&'a CrossEnvConfig) -> &'a Vec<String>,
+    {
+        self.get_target(target)
+            .map_or(Vec::new(), |t| get(&t.env).clone())
```
Emilgardis, line 143: same here, we could return `&[String]`

Emilgardis (commented):
awesome!

I don't like the look of having long-named `Fn` constraints, instead I think it's more readable if they are just `impl` bounds.

src/cross_toml.rs
Alexhuszagh: I've changed all of these to return `&[String]`. The only major change required was then to ensure that the sum had the appropriate lifetimes.

```rust
fn sum_of_env_toml_values<'a>(
    &'a self,
    env_values: Option<Vec<String>>,
    toml_getter: impl FnOnce(&'a CrossToml) -> &'a [String],
) -> Result<Vec<String>> {
    let mut collect = vec![];
    if let Some(mut vars) = env_values {
        collect.append(&mut vars);
    } else if let Some(toml_values) = self.toml.as_ref().map(toml_getter) {
        collect.extend(toml_values.iter().cloned());
    }

    Ok(collect)
}
```

Alexhuszagh:
I've changed all the generic `Fn` constraints to `impl Fn`, and then also changed the `Vec<String>` returns to `&[String]`.

Alexhuszagh:
@Emilgardis Any chance I can get a review on this? I'm working on adding `build-std` support on top of this. Should I merge this in with the patch for #773 into one PR?

Emilgardis (approved):
lgtm! I do feel there's more simplification possible here, but this is a nice step forward.

bors r+

PR #775, opened 2022-06-09
Forward Cargo exit code

Jules-Bertholet:
Forward Cargo's exit code. For example, if `cargo test` returns a non-zero exit code due to a failing test, `cross` will also return a non-zero exit code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #775 - forward Cargo exit code to host
 - #767 - added the `cross-util` and `cross-dev` commands.
 - #745 - added `thumbv7neon-*` targets.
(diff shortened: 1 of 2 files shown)


Alexhuszagh (approved):
lgtm. A simple test shows it works as desired, and properly returns the error code as if cargo has run it.

Alexhuszagh:
@Emilgardis Can you merge this manually? Only failing test is `dragonfly`, which is because it's a tier 3 target and we shouldn't be running tests on it. I'll submit a CI PR for this later.

Emilgardis:
we have xargo for a reason, making it use nightly would fix it, no?

Alexhuszagh:
> we have xargo for a reason, making it use nightly would fix it, no?

Yes it should. Let me test locally first.

Emilgardis (approved):
bors r+

Emilgardis (approved):
Thanks!

bors r=Alexhuszagh,Emilgardis

PR #776, opened 2022-06-09
Bug fix for dragonfly and CI tests.

Alexhuszagh:
Added `build-std` flag to CI, so for targets that support `std` but are tier 3 targets, we can use the experimental `build-std` support to ensure they build. Also patches linking on dragonfly to ensure we use `libgcc_s` instead of `libgcc_pic`, which no longer exists. We just create a symlink for that purpose.

Necessary to merge #775.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
             # Disabled for now, see https://github.com/rust-lang/rust/issues/85821
             #- { target: asmjs-unknown-emscripten,         os: ubuntu-latest,  cpp: 1,           std: 1, run: 1 }
             #- { target: wasm32-unknown-emscripten,        os: ubuntu-latest,  cpp: 1,           std: 1, run: 1 }
-            - { target: x86_64-unknown-dragonfly,         os: ubuntu-latest,  cpp: 1, dylib: 1 }
+            - { target: x86_64-unknown-dragonfly,         os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, build-std: 1 }
             - { target: i686-unknown-freebsd,             os: ubuntu-latest,          dylib: 1, std: 1 }
             - { target: x86_64-unknown-freebsd,           os: ubuntu-latest,          dylib: 1, std: 1 }
             - { target: x86_64-unknown-netbsd,            os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1 }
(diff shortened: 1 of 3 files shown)


Alexhuszagh:
bors try --target *-dragonfly

otavio (approved):

Emilgardis:
it's not working

Alexhuszagh:
Checking it, yeah don't merge it yet.

Alexhuszagh:
Ok I think I've fixed everything. I've just changed the logic so instead of `cross`, we use `${CROSS[@]}` and `"${CROSS_FLAGS[@]}"` everywhere, which then sets the toolchain version and if we need `-Zbuild-std`. It works locally on my machine.

bors try --target x86_64-unknown-dragonfly --target arm-unknown-linux-gnueabihf --target thumbv6m-none-eabi

Emilgardis:
im not a fan of this, instead I think it'd suffice if we make https://github.com/cross-rs/cross/blob/ee2fc1b55d3e5fd4aee37d3ce3ffa88016fd26cc/src/lib.rs#L352-L354 actually properly determine if build-std is needed, and then provide the needed flags if possible. Obviously, that's a bigger change than this simple fix, but I think that's the long-term solution, especially if we add more t3 targets to be "builtins"

Alexhuszagh:
> im not a fan of this, instead I think it'd suffice if we make
> 
> https://github.com/cross-rs/cross/blob/ee2fc1b55d3e5fd4aee37d3ce3ffa88016fd26cc/src/lib.rs#L352-L354
> actually properly determine if build-std is needed, and then provide the needed flags if possible. Obviously, that's a bigger change than this simple fix, but I think that's the long-term solution, especially if we add more t3 targets to be "builtins"

Probably a more logical solution, should I incorporate this into this PR then? It's probably better to have this logic outside our test suite for sure.

Emilgardis:
I'm fine with merging this already or with the suggested "real" fix.

Emilgardis (approved):
bors r+

src/lib.rs
```diff
@@ -349,9 +355,20 @@ pub fn run() -> Result<ExitStatus> {
...
+                    !available_targets.contains(&target)
+                        || needs_build_std(&toolchain, &target, verbose)?
+                }
+            };
+            if !is_nightly && uses_build_std {
+                eyre::bail!(
+                    "no rust-std component available for {}: must use nightly",
+                    target.triple()
+                );
+            }
```
Emilgardis, line 371: what if we're `no_std`?

Alexhuszagh: > what if we're `no_std`?

Uh true, so then just disable it? I guess we should just let cargo take care of it then. I should probably also add a changelog entry. I should probably also check if `rust-src` is available.... Hmmm.

PR #777, opened 2022-06-10
Update CHANGELOG for SSH dependency support.

Alexhuszagh:


CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #543 - Added environment variables to control the UID and GID in the container
 - #524 - docker: Add Nix Store volume support
 - Added support for mounting volumes.
-- #684 - Enable cargo workspaces to work from any path in the workspace, and make path dependencies mount seamlessly.
+- #684 - Enable cargo workspaces to work from any path in the workspace, and make path dependencies mount seamlessly. Also added support for private SSH dependencies.
```
Emilgardis, line 22: it's kindof misleading, it's more that we fetch ahead of time, no?

Alexhuszagh: > it's kinda misleading, it's more that we fetch ahead of time, no?

I'll modify it then. But I wanted to avoid getting into the internals, but since we fetch it on the host in the same command, is there a practical difference?

Emilgardis: meh, let's keep it as is

otavio (approved):

PR #778, opened 2022-06-10
Fix target extraction from `rustembedded` images.

Alexhuszagh:
Previously, our extraction using `rustembedded_target` could fail if there was a character that did not match `\w` in our triple. This failed with targets like `thumbv8m.main-none-eabi`, and therefore should be considered just as a precaution.


Alexhuszagh:
I know this is a very minor bug fix, since our old `rustembedded` targets didn't support any architectures with a non-`\w` character in them, but architectures such as `thumbv8m.main-none-eabi` are supported, and I've added tests to ensure these work.

otavio (approved):

Alexhuszagh:
bors r=otavio

PR #779, opened 2022-06-10
Limit calls to which to find container engine path.

Alexhuszagh:
Store the path to the engine in a variable, so we can limit the number of calls to `get_container_engine` so we can only use it once, avoiding using multiple commands per docker command.

This is relevant for docker remote support, where we might make 20+ docker commands. Here, since we only make 2, it's much less relevant but still good practice.


Emilgardis (approved):
bors r+

nice

PR #781, opened 2022-06-11
Prefer target.env values over build.env config.

### LINKED ISSUE #773: Incorrect Environment Variable/Config Logic

Alexhuszagh:
We currently use `build.env` over `target.$(...).env`, which means we cannot set a default value and then override it for a specific target. Given the following config file:

```toml
[build.env]
xargo = true

[target.aarch64-unknown-linux-gnu.env]
xargo = false
```

We currently would use `xargo` even on `aarch64-unknown-linux-gnu`, when we should not, and only use `xargo` for every other target.

Closes #773.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #781 - ensure `target.$(...)` config options override `build` ones.
 - #771 - fix parsing of `DOCKER_OPTS`.
 - #727 - add `PKG_CONFIG_PATH` to all `*-linux-gnu` images.
 - #722 - boolean environment variables are evaluated as truthy or falsey.
--- a/src/config.rs
+++ b/src/config.rs
@@ -137,13 +137,13 @@ impl Config {
             (None, None)
         };
 
-        match (env_build, toml_build) {
+        match (env_target, toml_target) {
             (Some(value), _) => return Some(value),
             (None, Some(value)) => return Some(value),
             (None, None) => {}
         };
 
-        match (env_target, toml_target) {
+        match (env_build, toml_build) {
             (Some(value), _) => return Some(value),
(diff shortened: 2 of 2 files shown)


Emilgardis (approved):
bors r+

PR #782, opened 2022-06-11
Add build-std config option.

### LINKED ISSUE #692: drop xargo in favor of `build-std`
I think it makes sense to drop xargo integration in cross for almost all targets.

https://github.com/rust-lang/wg-cargo-std-aware

Alexhuszagh:
Detects if `rust-std` is not an available component for a target, and if `build-std = true` in `Cross.toml` under `[build]` or `[target.(...)]`, then use the `-Zbuild-std` flag. This has higher priority than `xargo`: `xargo` will only be used if a target is not a built-in.

A sample `Cross.toml` file is the following:

```toml
[target.x86_64-unknown-dragonfly]
build-std = true
```

Closes #692.


Emilgardis:
what if we want to build std with our compiler options even if std is already provided?

Alexhuszagh:
> what if we want to build std with our compiler options even if std is already provided?

Would it be a good idea then to make it always active when set to true? So remove the following checks and only allow it to be enabled on nightly in all cases? I'll have to confirm how it works for targets with pre-built standard libraries, if it still needs nightly to compile std from source.

https://github.com/cross-rs/cross/pull/782/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R361-R366

Emilgardis:
you need nightly, and that's probably always going to be true due to some features std uses.

Alexhuszagh:
I've updated it so `build-std` is always enabled if provided as a config option, and tested it on both `x86_64-unknown-dragonfly` and `arm-unknown-linux-gnueabihf` to ensure it builds the Rust standard library in both cases.

Emilgardis:
lgtm, please edit the pr description (as that's what bors uses for commit message)

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #783, opened 2022-06-11
Fix clippy lints on nightly.

Alexhuszagh:


Alexhuszagh:
I recently added the following as a pre-commit hook:

```bash
echo "Running rustfmt and clippy checks."
cargo fmt -- --check
cargo clippy -- --deny warnings
cargo +nightly clippy -- --deny warnings
```

The nightly clippy warnings produced a few suggestions, and although checking it in our CI might cause commits to break because of frequent updates to the lints, periodic checks might be good to ensure clippy passes on nightly.

Emilgardis (approved):
bors r+

Issue #724, opened 2022-05-27
glibc version error in nightly targets


Alexhuszagh:
### Version Info

```bash
$ cross --version
cross 0.2.1 (459e03e 2022-05-26)
Warning: Falling back to `cargo` on the host.
cargo 1.63.0-nightly (39ad1039d 2022-05-25)
```

Sample code to reproduce:

```bash
$ git clone https://github.com/cross-rs/rust-cpp-hello-word
$ cd rust-cpp-hello-word
$ cross run --target powerpc64-unknown-linux-gnu --verbose
+ "rustc" "--print" "sysroot"
+ "rustup" "toolchain" "list"
+ "rustup" "target" "list" "--toolchain" "nightly-x86_64-unknown-linux-gnu"
+ "rustup" "component" "list" "--toolchain" "nightly-x86_64-unknown-linux-gnu"
+ "/usr/bin/docker" "run" "--userns" "host" "-e" "PKG_CONFIG_ALLOW_CROSS=1" "--rm" "--user" "1000:1000" "-e" "XARGO_HOME=/xargo" "-e" "CARGO_HOME=/cargo" "-e" "CARGO_TARGET_DIR=/target" "-e" "USER=ahuszagh" "-e" "CROSS_RUNNER=" "-v" "/home/ahuszagh/.xargo:/xargo:Z" "-v" "/home/ahuszagh/.cargo:/cargo:Z" "-v" "/cargo/bin" "-v" "/home/ahuszagh/Desktop/cross/rust-cpp-hello-word:/project:Z" "-v" "/home/ahuszagh/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu:/rust:Z,ro" "-v" "/home/ahuszagh/Desktop/cross/rust-cpp-hello-word/target:/target:Z" "-w" "/project" "-i" "-t" "ghcr.io/cross-rs/powerpc64-unknown-linux-gnu:main" "sh" "-c" "PATH=$PATH:/rust/bin cargo run --target powerpc64-unknown-linux-gnu --verbose"
       Fresh cc v1.0.73
   Compiling hellopp v0.1.0 (/project)
     Running `/target/debug/build/hellopp-0a565c6e09d5b0ce/build-script-build`
error: failed to run custom build command for `hellopp v0.1.0 (/project)`

Caused by:
  process didn't exit successfully: `/target/debug/build/hellopp-0a565c6e09d5b0ce/build-script-build` (exit status: 1)
  --- stderr
[...]

Issue #90, opened 2017-04-17
Issues with Docker Daemon when user namespace is enabled


lawliet89:
When the Docker daemon has [user namespace](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-user-namespace-options) enabled, the following will fail

 - Reading/writing to bind mount volumes
 - `--privileged` flag

The fix is as simple as adding `--userns=host` to the `docker` command. BUT this will break on older Docker client versions that do not support this. I can probably fix this easily, but I am not sure how I should proceed.

EDIT: I suggest this: Use a `lazy_static` or something similar to do a docker engine version check. (`docker version --format '{{.Server.APIVersion}}'`) Then if the semver is newer than `1.24`, we automatically add the flag `--userns=host`.

PR #785, opened 2022-06-12
Initial commit with basic support for remote docker.

### LINKED ISSUE #248: Cross fails to mount if docker daemon not running on same machine

### LINKED ISSUE #273: Recommended GitLab usage
Just wondering if there’s any recommended guidelines for using cargo cross with GitLab CI. Eg whether to do docker-in-docker or to use the target architecture dockerfiles/images directly.

Alexhuszagh:
This supports the volume-based structure, and uses some nice
optimizations to ensure that only the desired toolchain and cargo items
are copied over. It also uses drops to ensure scoped deletion of
resources, to avoid complex logic ensuring their cleanup.

It also supports persistent data volumes, through `cross-util`. In order to setup a persistent data volume, use:

```bash
cross-util create-crate-volume --target arm-unknown-linux-gnueabihf
```

Make sure you provide your `DOCKER_HOST` or correct engine type to ensure these are being made on the remote host. Then, run your command as before:

```bash
CROSS_REMOTE=true cross build --target arm-unknown-linux-gnueabihf
```

Finally, you can clean up the generated volume using:

```bash
cross-util remove-crate-volume --target arm-unknown-linux-gnueabihf
```

A few other utilities are present in `cross-util`:
- `list-volumes`: list all volumes created by cross.
- `remove-volumes`: remove all volumes created by cross.
- `prune-volumes`: prune all volumes unassociated with a container.
- `list-containers`: list all active containers created by cross.
- `remove-containers`: remove all active containers created by cross.

The initial implementation was done by Marc Schreiber, https://github.com/schrieveslaach.

Fixes #248.
Fixes #273.
Closes #449.


Alexhuszagh:
This is a **massive** PR, and I'll likely be incrementally improving the code quality, but reviews are greatly appreciated, as our beta-testers.

Alexhuszagh:
I believe this might break private SSH dependency support, but I'm not sure if there's a great solution to this anyway.

src/docker.rs
```diff
@@ -98,145 +193,412 @@ fn parse_docker_opts(value: &str) -> Result<Vec<String>> {
...
     {
-        mount_path = host_path.clone();
+        Ok(path.to_path_buf())
     }
-    cmd.args(&[
+}
+
+fn canonicalize_host_path(path: &Path) -> Result<PathBuf> {
+    file::canonicalize(path).wrap_err_with(|| format!("when canonicalizing path `{path:?}`"))
+}
```
Emilgardis, line 333: why it's own function?

src/docker.rs
```diff
@@ -98,145 +193,412 @@ fn parse_docker_opts(value: &str) -> Result<Vec<String>> {
...
+) -> Result<()> {
+    // only need to copy the rustlib files for our current target.
+    let triple = target.triple();
+    let relpath = Path::new("lib").join("rustlib").join(&triple);
+    let src = xargo_dir.join(&relpath);
+    let dst = mount_prefix.join("xargo").join(&relpath);
+    if Path::new(&src).exists() {
+        create_volume_dir(engine, container, dst.parent().unwrap(), verbose)?;
+        copy_volume_files(engine, container, &src, &dst, verbose)?;
     }
```
Emilgardis, line 394: why are there their own functions? can they not be summarized to one?

Alexhuszagh: > why are these their own functions? can they not be summarized to one?

Sure they could, but they're also somewhat different functionality each so I figured it would be handy (especially during testing) the separate them.

src/docker.rs
Alexhuszagh: I just broke it up because it's used in many places and wanted the same error message upon failure. I could move it easily into `file::canonicalize` though, that is a good idea.

Emilgardis: add the error message to `file::canonicalize`, good to have it standardized

src/docker.rs
Emilgardis: resolved for now, will revisit when this is more fleshed out.

do you want to convert this to a draft for now maybe to signal clearly it's in progress?

Emilgardis:
this is great!

will have to take some time to take it all in

src/docker.rs
```diff
@@ -343,9 +755,302 @@ pub fn run(
...
+
+    // 2. create our volume to copy all our data over to
+    if let VolumeId::Discard(ref id) = volume {
+        volume_create(&engine, id, verbose)?;
+    }
+    let _volume_deletter = DeleteVolume(&engine, &volume, verbose);
+
+    // 3. create our start container command here
+    let mut docker = docker_subcommand(&engine, "run");
+    docker.args(&["--userns", "host"]);
```
fenollp, line 822: Probably off topic but how is `userns=host` beneficial? https://docs.docker.com/engine/security/userns-remap/#disable-namespace-remapping-for-a-container didn't clear things up for me.

Alexhuszagh: There's a few reasons, and this has been around since 2017. Some of them are historical with our design (I'm a relatively new maintainer, so I might not know everything), but the reasons are partially documented in #90. Our android images used to require the `--privileged` flag, which is incompatible with userns remapping (until #746, where we removed this need and mitigated the permissions a container could use). It also interferes with bind mounts, which we make extensive use of. 

Finally, we remap the user in the container to the current user anyway using the `--user 1000:1000` flag (normally, if run as root it can be `0:0` I believe), so this really isn't an issue anyway. Stuff run as `--user 1000:1000` means it's a generic, unprivileged user so the permissions and owner of the files generated will be the same as your current user, but also the commands will be run unprivileged. This means user namespace remapping isn't that useful for us anyway.

I hope that explains it? Also, in general, anyone using cross has access to a Docker or Podman daemon, which in the former probably has root permissions (unless using rootless docker) on the machine, so it's really only a concern if you're worried the crate's build system is dangerous (a valid concern), but we have some mitigations in place for that anyway.

Alexhuszagh:
Just a few things as considerations:

1. I currently skip the cargo registry, as this is generally a good idea, however, it might break private SSH dependency support. This could be enabled with an environment variable (I was previously using `CROSS_REMOTE_COPY_REGISTRY`.
2. The target directory gets to be quite large, and this can often be slower than doing a full rebuild. It's easy to identify anything that **should not** be copied over, since Rust uses [cache directory tagging](https://bford.info/cachedir/). For example, a project I had not run `cargo clean` on in 4 years had target directory size of 90GB.

Basically, a config file or 2 environment variable settings saying:
- `CROSS_REMOTE_COPY_REGISTRY`: do copy the registry
- `CROSS_REMOTE_COPY_CACHE`: do copy cache files, including the target directory

Alexhuszagh:
Ok I've added the registry support and cache support, which should further speed up execution time.

Not setting `CROSS_REMOTE_COPY_CACHE` will effectively disable incremental compilation unless a persistent volume is used, which in most cases is still probably faster unless doing a release build.

Setting `CROSS_REMOTE_COPY_REGISTRY` will also copy the entire registry, which **should** allow the use of private SSH dependencies.

Emilgardis (commented):
the files are getting rather large, could you split out the remote support stuff into it's own module, e.g inside `docker/remote.rs`

src/bin/cross-util.rs
```diff
@@ -51,6 +55,96 @@ enum Commands {
...
+        /// Force removal of containers.
+        #[clap(short, long)]
+        force: bool,
+        /// Remove containers. Default is a dry run.
+        #[clap(short, long)]
+        execute: bool,
+        /// Container engine (such as docker or podman).
+        #[clap(long)]
+        engine: Option<String>,
+    },
```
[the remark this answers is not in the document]
Alexhuszagh: Yep was just thinking that, and it should simplify maintaining any changes and rebasing as well on my end.

src/commands/containers.rs
```diff
@@ -0,0 +1,370 @@
+use std::path::Path;
```
Emilgardis, line 1: can this not be inside bin?

Alexhuszagh: I believe we'd have to turn off autobin discovery, and then manually register our binaries. I'm fine with that, if you are:
https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery

Emilgardis: No it seems to work to just move the commands mod you placed in `src/commands` in `src/bin/commands` instead and fix the imports so that they are public, could probably make all cross mods public

Alexhuszagh: Ah ok, I thought you meant move them to the root directory. Will do shortly, I've also done some refactoring inside docker.

Alexhuszagh:
I've refactored the codebase more thoroughly:

1. `docker` is now a directory, with the `remote`, `local`, `shared`, and `engine` submodules.
2. `commands` is now in `/src/bin`.
3. Since we now glob import from `docker`, I've prefixed a few of the names.
4. Another thing I've fixed (which should be a separate PR) is fixed docker_in_docker support, which currently expects a hard-coded docker and can just use the engine value we've already gotten.

This should help isolated the behavior better between local and shared, and help

Alexhuszagh:
I've rebased after #794 and updated everything to use the remote options if `CROSS_REMOTE` is active. I'm not sure if building images will work if that's the case, but it's likely better to error out than try to build them on the local machine if they're trying to use a remote host.

src/bin/cross-util.rs
```diff
@@ -20,6 +20,20 @@ enum Commands {
...
+    /// Prune volumes not used by any container.
+    PruneVolumes(commands::PruneVolumes),
+    /// Create a persistent data volume for the current crate.
+    CreateCrateVolume(commands::CreateCrateVolume),
+    /// Remove a persistent data volume for the current crate.
+    RemoveCrateVolume(commands::RemoveCrateVolume),
+    /// List cross containers in local storage.
+    ListContainers(commands::ListContainers),
+    /// Stop and remove cross containers in local storage.
+    RemoveContainers(commands::RemoveContainers),
```
Emilgardis, line 36: maybe these should be refactored into further subcommands?

`cross-util volumes prune` etc

Alexhuszagh: Sounds good, should probably be done with `images` too: `images list` and `images remove`.

Alexhuszagh:
Oh another issue: we use a temporary directory which won't necessarily get cleaned up if it's interrupted by `Ctrl + C` during execution, and this should probably be:
1. Prefixed with `cross` and stored in a temporary directory that should get cleaned up with the rust install.
2. Cleaned up if we have a non-fatal termination of execution (anything besides SIGKILL).

src/docker/mod.rs
```diff
@@ -24,15 +27,31 @@ pub fn run(
...
-        metadata,
-        config,
-        uses_xargo,
-        sysroot,
-        verbose,
-        docker_in_docker,
-        cwd,
-    )
+    if engine.is_remote {
+        remote_run(
```
Emilgardis, line 31:
```suggestion
        remote::run
```

src/docker/mod.rs
```diff
@@ -1,8 +1,10 @@
 mod engine;
 mod local;
+pub mod remote;
 mod shared;
 
 pub use self::engine::*;
+pub use self::remote::*;
```
Emilgardis, line 7: I think glob importing should be used when it's a prelude or alike, not for this

Alexhuszagh:
Ok I've rebased off of the better tempfile support and removed the glob importing from `remote`. The imports are kind of... [verbose](https://github.com/cross-rs/cross/commit/dbc474cbc99d62e4f0031dadd59d2a935997e721#diff-7f0024537bbf8484f3e68ec502bd9cff7805e1f6eaee0eb072647a14e1bdb36fR7-R11) now, but they do work:

```rust
pub use self::remote::{
    container_rm, container_state, container_stop, copy_volume_container_cargo,
    copy_volume_container_rust, copy_volume_container_xargo, unique_container_identifier,
    volume_create, volume_exists, volume_rm, ContainerState,
};
```

It's possible I could move these to shared or something, but that doesn't make much sense. I could also juse make `remote` a public module and then prefix everything with `cross::docker::remote::`.

Emilgardis:
> could also juse make `remote` a public module and then prefix everything with `cross::docker::remote::`.

Or

```rust
use cross::docker::remote;

remote::blabla()
```

Alexhuszagh:
`docker::remote` no longer has a glob import, and is now a public module. I believe this is ready for review, and it's undergone some basic testing.

src/docker/remote.rs
```diff
@@ -0,0 +1,653 @@
...
+    if toolchain_path.exists() {
+        copy_volume_files(engine, container, &toolchain_path, &dst_rustlib, verbose)?;
+    }
+
+    // now we need to copy over the host toolchain too, since it has
+    // some requirements to find std libraries, etc.
+    let rustc = sysroot.join("bin").join("rustc");
+    let libdir = Command::new(rustc)
+        .args(&["--print", "target-libdir"])
+        .run_and_get_stdout(verbose)?;
```
Emilgardis, line 302: this assumes the host is `x86_64-unknown-linux-gnu`

on my mac, I'm getting 

````
emil@Emils-MacBook-Pro rust-cpp-hello-word % CROSS_REMOTE=1 cross -v build --target x86_64-unknown-linux-gnu
+ cargo metadata --format-version 1 --filter-platform x86_64-unknown-linux-gnu
+ rustc --print sysroot
+ rustup toolchain list
+ rustup target list --toolchain stable-x86_64-unknown-linux-gnu
+ rustup component list --toolchain stable-x86_64-unknown-linux-gnu
+ /usr/local/bin/docker
+ /usr/local/bin/docker volume inspect cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e-keep
+ /usr/local/bin/docker ps -a --filter 'name=cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e' --format {{.State}}
+ /usr/local/bin/docker volume inspect cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker volume create cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker run --userns host --name cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e -v cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross -e 'PKG_CONFIG_ALLOW_CROSS=1' -e 'XARGO_HOME=/xargo' -e 'CARGO_HOME=/cargo' -e 'CARGO_TARGET_DIR=/target' -e 'CROSS_RUNNER=' -e 'USER=emil' -v /cross/cargo/bin -d -t ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main sh -c 'sleep infinity'
394cab661e2fc43808f47e32065124403ce3e58a83740565d489ad4ae62aacf6
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/cargo'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.cargo/bin cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/cargo
+ /usr/local/bin/docker cp -a /Users/emil/.cargo/env cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/cargo
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/rust'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/libexec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/etc cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a '/Users/emil/Library/Application Support/cross-rs/tmp/.tmpZBMzqs' cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust/lib
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/rust/lib/rustlib'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust/lib/rustlib
+ /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc --print target-libdir
+ /usr/local/bin/docker stop cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker rm cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker volume rm cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
Error: 
   0: `/Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc --print target-libdir` failed with exit status: 126

Stderr:
   /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc: /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc: cannot execute binary file
```

Emilgardis: or, assumes the host can run the toolchain

Emilgardis: this assumes the host is `x86_64-unknown-linux-gnu`

on my mac, I'm getting 

````
emil@Emils-MacBook-Pro rust-cpp-hello-word % CROSS_REMOTE=1 cross -v build --target x86_64-unknown-linux-gnu
+ cargo metadata --format-version 1 --filter-platform x86_64-unknown-linux-gnu
+ rustc --print sysroot
+ rustup toolchain list
+ rustup target list --toolchain stable-x86_64-unknown-linux-gnu
+ rustup component list --toolchain stable-x86_64-unknown-linux-gnu
+ /usr/local/bin/docker
+ /usr/local/bin/docker volume inspect cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e-keep
+ /usr/local/bin/docker ps -a --filter 'name=cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e' --format {{.State}}
+ /usr/local/bin/docker volume inspect cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker volume create cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker run --userns host --name cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e -v cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross -e 'PKG_CONFIG_ALLOW_CROSS=1' -e 'XARGO_HOME=/xargo' -e 'CARGO_HOME=/cargo' -e 'CARGO_TARGET_DIR=/target' -e 'CROSS_RUNNER=' -e 'USER=emil' -v /cross/cargo/bin -d -t ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main sh -c 'sleep infinity'
394cab661e2fc43808f47e32065124403ce3e58a83740565d489ad4ae62aacf6
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/cargo'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.cargo/bin cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/cargo
+ /usr/local/bin/docker cp -a /Users/emil/.cargo/env cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/cargo
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/rust'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/libexec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/etc cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust
+ /usr/local/bin/docker cp -a '/Users/emil/Library/Application Support/cross-rs/tmp/.tmpZBMzqs' cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust/lib
+ /usr/local/bin/docker exec cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e sh -c 'mkdir -p '\''/cross/rust/lib/rustlib'\'''
+ /usr/local/bin/docker cp -a /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e:/cross/rust/lib/rustlib
+ /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc --print target-libdir
+ /usr/local/bin/docker stop cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker rm cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
+ /usr/local/bin/docker volume rm cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
cross-hellopp-x86_64-unknown-linux-gnu-27448-47e2d-fe5b13d681f25ee6474be29d748c65adcd91f69e
Error: 
   0: `/Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc --print target-libdir` failed with exit status: 126

Stderr:
   /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc: /Users/emil/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc: cannot execute binary file
```

Emilgardis:
```
let host_toolchain_path = sysroot.join("lib/rustlib").join(target.triple());
```

should just work, if the toolchain is malformed, there's something wrong

src/docker/remote.rs
```diff
@@ -0,0 +1,653 @@
...
+pub fn container_state(engine: &Engine, container: &str, verbose: bool) -> Result<ContainerState> {
+    let stdout = subcommand(engine, "ps")
+        .arg("-a")
+        .args(&["--filter", &format!("name={container}")])
+        .args(&["--format", "{{.State}}"])
+        .run_and_get_stdout(verbose)?;
+    ContainerState::new(stdout.trim())
+}
+
+fn path_hash(path: &Path) -> String {
```
Emilgardis, line 360: this fn should be used in `Dockerfile::image_name` too

src/bin/commands/containers.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// List cross data volumes in local storage.
+    List(ListVolumes),
+    /// Remove cross data volumes in local storage.
+    Remove(RemoveVolumes),
+    /// Prune volumes not used by any container.
+    Prune(PruneVolumes),
+    /// Create a persistent data volume for the current crate.
+    CreateCrate(CreateCrateVolume),
+    /// Remove a persistent data volume for the current crate.
+    RemoveCrate(RemoveCrateVolume),
```
Emilgardis, line 122: these should have an alias I think

`create-crate` is a mouthful

simply doing `cross-util volumes create` and `cross-util volumes remove` is enough I think

src/bin/commands/containers.rs
```diff
@@ -0,0 +1,499 @@
...
+    // create a dummy running container to copy data over
+    let mount_prefix = Path::new("/cross");
+    let mut docker = docker::subcommand(engine, "run");
+    docker.args(&["--name", &container]);
+    docker.args(&["-v", &format!("{}:{}", volume, mount_prefix.display())]);
+    docker.arg("-d");
+    if atty::is(Stream::Stdin) && atty::is(Stream::Stdout) && atty::is(Stream::Stderr) {
+        docker.arg("-t");
+    }
+    docker.arg("ubuntu:16.04");
```
Emilgardis, line 361: does this have to be `ubuntu:16.04`? Is there a way to be smart about it, and use whatever `ubuntu` image is available.

src/docker/remote.rs
```diff
@@ -0,0 +1,653 @@
...
+
+    // the logic is broken into the following steps
+    // 1. get our unique identifiers and cleanup from a previous run.
+    // 2. create a data volume to store everything
+    // 3. start our container with the data volume and all envvars
+    // 4. copy all mounted volumes over
+    // 5. create symlinks for all mounted data
+    // 6. execute our cargo command inside the container
+    // 7. copy data from target dir back to host
+    // 8. stop container and delete data volume
```
Emilgardis, line 428: this doesn't apply if a `-keep` volume is available

src/docker/remote.rs
```diff
@@ -0,0 +1,653 @@
...
+        .unwrap_or(host_version_meta.short_version_string);
+
+    let workspace_root = &metadata.workspace_root;
+    let package = metadata
+        .packages
+        .iter()
+        .find(|p| p.manifest_path.parent().unwrap() == workspace_root)
+        .unwrap_or_else(|| metadata.packages.get(0).unwrap());
+
+    let name = &package.name;
```
Emilgardis, line 386: is this unwrap safe?

Also, can we apply this logic in `Dockerfile::image_name` :)

src/docker/remote.rs
```diff
@@ -0,0 +1,653 @@
...
+        .run_and_get_status(verbose, false)
+        .map_err(Into::into)
+}
+
+pub fn container_rm(engine: &Engine, container: &str, verbose: bool) -> Result<ExitStatus> {
+    subcommand(engine, "rm")
+        .arg(container)
+        .run_and_get_status(verbose, false)
+        .map_err(Into::into)
+}
```
Emilgardis, line 349: stop, create and rm for all these are the same, could be unified into one helper function

`command(engine, "action", "on", verbose)`

src/bin/commands/clean.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub verbose: bool,
+    /// Force removal of images.
+    #[clap(short, long)]
+    pub force: bool,
+    /// Remove local (development) images.
+    #[clap(short, long)]
+    pub local: bool,
+    /// Remove images. Default is a dry run.
+    #[clap(short, long)]
+    pub execute: bool,
```
Emilgardis, line 20: not providing `--execute` should probably display a note that to execute, pass --execute

Emilgardis: same should be done in `images` subcommand

Emilgardis (changes requested):
Looks good!

Need to fix the assumption about running the toolchains rustc, andd then some things I think can be better, but otherwise this looks great.

I think the temp file handling could do with some tracing, but it's not in the scope of this pr and not having it is fine imo

src/docker/remote.rs
Emilgardis: if it is safe, use `expect` instead

Emilgardis:
with keep, instead of having a volume for each crate, would it be fair to instead have a volume for the wanted toolchain?

This would also make things more portable, e.g you can simply do `cross-util volumes create --target powerpc-unknown-linux-gnu` once and you can now use this implied toolchain/cross volume with the target available for every future crate you use with `CROSS_REMOTE`

Alexhuszagh:
> with keep, instead of having a volume for each crate, would it be fair to instead have a volume for the wanted toolchain?
> 
> This would also make things more portable, e.g you can simply do `cross-util volumes create --target powerpc-unknown-linux-gnu` once and you can now use this implied toolchain/cross volume with the target available for every future crate you use with `CROSS_REMOTE`

Sounds like a good idea, since we have to copy over the project data over every time anyway.

Emilgardis:
> Sounds like a good idea, since we have to copy over the project data over every time anyway.

Just to check that I'm understanding what this means.

assuming we're currently on toolchain stable-x86_64-unknown-linux-gnu:

* `cross-util volume create --target x86_64-unknown-linux-gnu`
  This will create the volume, copy the toolchain binaries and everything else that is needed and finally install the target `x86_64-unknown-linux-gnu` inside too.
  The volume will be `cross-{toolchain_triple}-{toolchain_hash}-{commit_hash}-keep`
* `cross-util volume create --target powerpc-unknown-linux-gnu`
  this will mount the `powerpc-unknown-linux-gnu` target in the same volume
* `cross-util +nightly volume create --target powerpc-unknown-linux-gnu`
  This will create a new toolchain volume, since the commit hash is not the same

src/bin/commands/containers.rs
Alexhuszagh: > does this have to be `ubuntu:16.04`? Is there a way to be smart about it, and use whatever `ubuntu` image is available.

Sure, this was just similar to what binfmt was doing in `register`. To avoid another docker call, we could probably just store our base image somewhere in the code, and just make a note to increment it (such as in `increment_versions`) whenever we upgrade it. Right now we almost always use 16.04, but soon that will be universally 20.04. Or we could do similar to what is done in `get_cross_containers`, where we run ` docker images --filter reference='ubuntu:*' --format "{{.Tag}}"`, and then select the latest version.

This would give output similar to:

```bash
$  docker images --filter reference='ubuntu:*' --format "{{.Tag}}"
22.04
latest
16.04
20.04
```

Emilgardis: might be something to consider for later.

this can be dismissed, I personally dislike the idea of an extra docker call

Alexhuszagh:
> Just to check that I'm understanding what this means.
> 
> assuming we're currently on toolchain stable-x86_64-unknown-linux-gnu:
> 
>     * `cross-util volume create --target x86_64-unknown-linux-gnu`
>       This will create the volume, copy the toolchain binaries and everything else that is needed and finally install the target `x86_64-unknown-linux-gnu` inside too.

Yep, although I believe the `x86_64-unknown-linux-gnu` might have to change as referenced above. It should be the only toolchain with the `bin` subdirectory in it, which could make this a bit more general.

>       The volume will be `cross-{toolchain_triple}-{toolchain_hash}-{commit_hash}-keep`
> 
>     * `cross-util volume create --target powerpc-unknown-linux-gnu`
>       this will mount the `powerpc-unknown-linux-gnu` target in the same volume
> 
>     * `cross-util +nightly volume create --target powerpc-unknown-linux-gnu`
>       This will create a new toolchain volume, since the commit hash is not the same

Yep. It would be based on the toolchain name and hash, since this also prevents us from say, if `stable-x86_64-unknown-linux-gnu` is updated to a new version, the hash is no longer the same so we have a different toolchain and should therefore use a new volume.

src/bin/commands/containers.rs
Alexhuszagh: Same, it's probably a good idea to store a base version though to avoid having multiple references to the same image scattered throughout the codebase.

src/bin/commands/containers.rs
Alexhuszagh: We already have `volumes remove` though and it has different functionality. We could do `list-all` and `remove-all`, but this is slightly different than the `images` API.

src/docker/remote.rs
Alexhuszagh: There's a need also for more than just the target toolchain, otherwise there's errors in trying to find `rust-src`. It always needs the directory with the `bin` subdirectory, of which there is only one (and refers to the host toolchain), however. So it needs both the target and host toolchains, I believe, always. I'll do more testing on Windows (not WSL) and macOS.

Emilgardis: I think there's a terminology error here.

The `toolchain` is what you use to compile a target, say `stable-x86_64-unknown-linux-gnu`, this toolchain is compiled to run on a host that runs x86_64-unknown-linux-gnu. The binaries for the toolchain live under same directory as `$(rustup which rustc --toolchain <toolchain>)`, the parent of this directory is the sysroot (`rustc --print sysroot`), and the target libraries live under `$sysroot/lib/rustlib/<target>`.

You're already doing everything that is needed for host toolchain + host-target + target. The only thing wrong here is that you assume that the host can run the toolchain.

Emilgardis: on a related note, it shouldn't be called `host_toolchain_path`, but simply `toolchain_path`

src/temp.rs
```diff
@@ -0,0 +1,121 @@
...
+    let file = tempfile::NamedTempFile::new_in(&parent)?;
+    FILES.push(file);
+    Ok(FILES.last().unwrap().path())
+}
+
+/// # Safety
+/// Safe as long as we have single-threaded execution.
+unsafe fn pop_tempfile() -> Option<tempfile::NamedTempFile> {
+    FILES.pop()
+}
```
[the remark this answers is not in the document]
Alexhuszagh: We could probably leave a log file in the cross directory? Unless you want fancier tracing?

src/bin/commands/containers.rs
Alexhuszagh: Ok I've changed it to `list`, `remove-all` for containers and volumes, and then changed it to `create` and `remove` for persistent volumes, which now mention it's for the toolchain and not for the current crate.

src/bin/commands/containers.rs
Alexhuszagh: This now has a public constant with a note so we can update in future releases:

https://github.com/Alexhuszagh/cross/blob/d140c6dceee1fd294801a633f30a238ab858554a/src/docker/shared.rs#L17-L18

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #803 - added `CROSS_CUSTOM_TOOLCHAIN` to disable automatic installation of components for use with tools like `cargo-bisect-rustc`
 - #795 - added images for additional toolchains maintained by cross-rs.
 - #792 - added `CROSS_CONTAINER_IN_CONTAINER` environment variable to replace `CROSS_DOCKER_IN_DOCKER`.
+- #785 - added support for remote container engines through data volumes through setting the `CROSS_REMOTE` environment variable. also adds in utility to commands to create and remove persistent data volumes.
```
Emilgardis, line 15:
```suggestion
- #785 - added support for remote container engines through data volumes through setting the `CROSS_REMOTE` environment variable. also adds in utility commands to create and remove persistent data volumes.
```

typo

src/docker/remote.rs
Alexhuszagh: I've added a `run_and_get_status` function locally that does this, since the code can only be shared between this subset of commands easily.

```rust
fn run_and_get_status(engine: &Engine, args: &[&str], verbose: bool) -> Result<ExitStatus> {
    command(engine)
        .args(args)
        .run_and_get_status(verbose, false)
        .map_err(Into::into)
}
```

src/bin/commands/containers.rs
```diff
@@ -0,0 +1,502 @@
...
+            Containers::RemoveAll(l) => l.verbose,
+        }
+    }
+}
+
+fn get_cross_volumes(engine: &docker::Engine, verbose: bool) -> cross::Result<Vec<String>> {
+    let stdout = docker::subcommand(engine, "volume")
+        .arg("list")
+        .arg("--format")
+        .arg("{{.Name}}")
```
Emilgardis, line 230:
```suggestion
        .args(&["--format", "{{.Name}}"])
```

this could also be `.args(["foo", "bar"])` but we're using slices everywhere else so let's not change it. This is since 1.53.0 where `[T]: IntoIterator<Item=T>` now

src/bin/commands/containers.rs
Emilgardis: perfect

src/temp.rs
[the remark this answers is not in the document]
Emilgardis: i don't know what I want :)

src/bin/commands/containers.rs
```diff
@@ -0,0 +1,502 @@
...
+    if force {
+        command.arg("--force");
+    }
+    command.args(&volumes);
+    if volumes.is_empty() {
+        Ok(())
+    } else if execute {
+        command.run(verbose, false).map_err(Into::into)
+    } else {
+        eprintln!("note: this is a dry run. to remove the volumes, pass the `--execute` flag.");
```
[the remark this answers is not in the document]
Alexhuszagh: Probably a good idea for consistency, since I think we use it for warnings. The only exceptions are in this PR, so definitely. We use capitalization for `eprintln!("Warning: ...");`. so this should be changed.

src/lib.rs
```diff
@@ -42,10 +43,10 @@ use serde::Deserialize;
 pub use self::cargo::{cargo_command, cargo_metadata_with_args, CargoMetadata, Subcommand};
 use self::cross_toml::CrossToml;
 use self::errors::Context;
-use self::rustc::{TargetList, VersionMetaExt};
 
-pub use self::errors::{install_panic_hook, Result};
+pub use self::errors::{install_panic_hook, install_termination_hook, Result};
 pub use self::extensions::{CommandExt, OutputExt};
+pub use self::rustc::{get_sysroot, target_list, version_meta, TargetList, VersionMetaExt};
```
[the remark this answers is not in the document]
Alexhuszagh: We could just make it `pub mod rustc` then, which I was planning on doing.

src/docker/remote.rs
Alexhuszagh: Going to wait on #826 before changing this.

xtask/src/util.rs
```diff
@@ -84,9 +81,8 @@ pub fn format_repo(registry: &str, repository: &str) -> String {
     output
 }
 
-pub fn pull_image(engine: &Path, image: &str, verbose: bool) -> cross::Result<()> {
-    let mut command = Command::new(engine);
-    command.arg("pull");
+pub fn pull_image(engine: &cross::docker::Engine, image: &str, verbose: bool) -> cross::Result<()> {
+    let mut command = cross::docker::subcommand(engine, "pull");
```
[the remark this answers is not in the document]
Alexhuszagh: I've changed many of the longer statements (those that use `cross::docker::{{X}}` many times, especially `cross::docker::Engine` to `use cross::docker` and then `docker::{{X}}`, which also simplifies the statements a lot.

src/docker/remote.rs
Emilgardis: Why? The assumption cross does today (which #817 tries to fix) is that all images run on x86_64-unknown-linux-gnu. If we don't replace it, we could possibly mount the hosts (the one running `CROSS_REMOTE=1 cross ...` ) toolchain, which could be aarch64, powerpc or whatever.

Alexhuszagh: No this is the issue for my code specifically, is that we need the toolchain that has the bin directory, which is the one for the toolchain we're currently using. Say for example, if we're using `x86_64-pc-windows-msvc`, it doesn't need it to compile, but it does need the other subdirectories, and that's why it views it's hardcoding `x86_64-unknown-linux-gnu`.

Emilgardis: > Say for example, if we're using x86_64-pc-windows-msvc

Is this the target or the "toolchain"?

Alexhuszagh: The toolchain. The default toolchain. We need both the target and the default toolchain. So if I'm cross-compiling to `arm-unknown-linux-gnueabihf` on `x86_64-unknown-linux-gnu`, I need both, but on `x86_64-pc-windows-msvc`, I need `arm-unknown-linux-gnueabihf` and `x86_64-pc-windows-msvc` but should not need `x86_64-unknown-linux-gnu`. This is because that directory has `rust-src` or `rust-std` (depending on if building from source or not) that are needed, and aren't present elsewhere in the specific toolchains.

Emilgardis: Aha! That's wrong and where the misunderstanding is.

To compile for a target, we don't need its toolchain, you only need the components you require in `sysroot/lib/rustlib/<target>`. A toolchain has only one binary folder.

The toolchain we use in our images are always of sort `.rustup/toolchains/<channel>-<opt_date>-x86_64-unknown-linux-gnu`. We call this the `sysroot`. The current sysroot function makes it so that we pull the same toolchain "version" for x86_64-unknown-linux-gnu as the host.
We can't use the toolchain compiled for `x86_64-pc-windows-msvc` inside our images (yet).

The intended effect of replacing the host_triple in `<channel>-<opt_date>-<host_triple>` with `x86_64-unknown-linux-gnu` is so that we know it can run on the images. We also make sure to `rustup toolchain add --toolchain <channel>-<opt_date>-x86_64-unknown-linux-gnu`

The binaries for a toolchain lives in `sysroot/bin`, the binaries looks for the needed components in `sysroot/lib/rustlib/<target>`. You don't have multiple toolchains, you have multiple targets in that toolchain. see https://rust-lang.github.io/rustup/cross-compilation.html

The issue with the current changeset is that it pulls in whatever the host toolchain is, that will only ever be valid when you are on `x86_64-unknown-linux-gnu` yourself


So, if you're on cross-compiling to `arm-unknown-linux-gnueabihf` on `x86_64-unknown-linux-gnu`, you only need

`rustup toolchain add x86_64-unknown-linux-gnu` and `rustup target add arm-unknown-linux-gnueabihf --toolchain x86_64-unknown-linux-gnu`

Emilgardis: ofc, as you pointed out, some components live in the host target folder, so that should always follow. But you already do that here, the only wrong thing is that we may not be able to run the `<channel>-<opt_date>-x86_64-unknown-linux-gnu`s rustc.

The call to `rustc --print target-libdir` is redundant, it's not needed, we already know where it would live

Alexhuszagh: I guess on non-Linux systems we probably will also need something more.

Emilgardis: No, not really.

src/docker/shared.rs
```diff
@@ -310,12 +337,7 @@ fn wslpath(path: &Path, verbose: bool) -> Result<PathBuf> {
         .arg("-a")
         .arg(path)
         .run_and_get_stdout(verbose)
-        .wrap_err_with(|| {
-            format!(
-                "could not get linux compatible path for `{}`",
-                path.display()
-            )
-        })
+        .wrap_err_with(|| format!("could not get linux compatible path for `{path:?}`",))
```
Emilgardis, line 340: why this change?

src/file.rs
```diff
@@ -2,10 +2,61 @@ use std::borrow::Cow;
...
+                    Component::CurDir => push_posix_path(&mut output, "."),
+                    Component::ParentDir => push_posix_path(&mut output, ".."),
+                    Component::Normal(path) => push_posix_path(&mut output, path.to_utf8()?),
+                }
+            }
+            Ok(output)
+        } else {
+            self.to_utf8().map(|x| x.to_string())
+        }
+    }
```
Emilgardis, line 57: can we use this in `docker::shared::docker_cwd` too

src/docker/shared.rs
Alexhuszagh: I'm explaining in a subsequent, separate PR shortly.

src/file.rs
Alexhuszagh: Yep, absolutely. I can submit this in a separate PR with my path changes in general.

src/docker/remote.rs
Alexhuszagh: `path_hash` has been moved to `docker::shared`, and is used in `image_name` and `unique_toolchain_identifier` (which is used in `unique_container_identifier`.

Alexhuszagh:
I've got to test this on Windows and macOS, but I've fixed the following issues:
- Fixed all the path issues previously used in #828.
- Made persistent data volumes for the toolchain, using `cross-{triple}-{toolchain_hash}-{commit_hash}`.
- Made temporary data volumes an extension of the persistent data volume, or `{toolchain_id}-{name}-{project_hash}`.

This prevents any conflict, and so the persistent data volume might be named `cross-arm-unknown-linux-gnueabihf-16b8c-fe5b13d681f25ee6474be29d748c65adcd91f69e` while the container and temporary data volume would be named `cross-arm-unknown-linux-gnueabihf-de1a6-fe5b13d681f25ee6474be29d748c65adcd91f69e-hellopp-c1f13`.

src/docker/remote.rs
```diff
@@ -0,0 +1,652 @@
...
+    let commit_hash = if let Some(version) = rustup::rustc_version_string(sysroot)? {
+        rustc::hash_from_version_string(&version, 1)
+    } else {
+        rustc::version_meta()?.commit_hash()
+    };
+
+    let triple = target.triple();
+    let toolchain_hash = path_hash(sysroot)?;
+    Ok(format!("cross-{triple}-{toolchain_hash}-{commit_hash}"))
+}
```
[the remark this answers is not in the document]
Alexhuszagh: Yeah, it does currently. I can make this more general still, but it would need some memo to ensure that I know if the toolchain for the triple has been previously copied over. I guess this could also solve the issue of having `rustup` not having previously installed the cross-compile target toolchain prior to creating the persistent volume.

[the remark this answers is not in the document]
Emilgardis: > but it would need some memo to ensure that I know if the toolchain for the triple has been previously copied over.

you mean for the target? Yes that a good point is there a good way to add that to some kind of metadata in the volume? maybe having a separate volume per toolchain/target is fine then, just sad about duplicating files

src/rustc.rs
```diff
@@ -31,6 +32,48 @@ impl VersionMetaExt for VersionMeta {
...
+        .and_then(|meta| meta.split_once(' '))
+    {
+        if is_hash(commit) && is_date(date) {
+            return short_commit_hash(commit);
+        }
+    }
+
+    // fallback: can't extract the hash. just create a hash of the version string.
+    let buffer = const_sha1::ConstBuffer::from_slice(version.as_bytes());
+    short_commit_hash(&const_sha1::sha1(&buffer).to_string())
```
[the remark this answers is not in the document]
Alexhuszagh: > does this return the sha1 in hex?

Yes, well, of the version string that was provided, and only the first 9 characters of the sha1 hexdigest, as an absolute fallback.

src/docker/remote.rs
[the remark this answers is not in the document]
Alexhuszagh: > > but it would need some memo to ensure that I know if the toolchain for the triple has been previously copied over.
> 
> you mean for the target? Yes that a good point is there a good way to add that to some kind of metadata in the volume? maybe having a separate volume per toolchain/target is fine then, just sad about duplicating files

Yeah, so currently there's a unique volume per target, mostly because I didn't want to have to deal with updating the rustlib directory every single time a new target was added, and wasn't sure if there was a good way to do so efficiently. There probably is, and it's probably a better solution which I'll look into.

Alexhuszagh:
So a few things have been resolved in the latest commit:

1. The persistent data volume can support more than 1 target. It does this by checking if the target directory in `rustlib` already exists, and if it does, doesn't copy over the files. If it does not exist, it copies over the files and recopies all the manifest files as well. This is a pretty cheap check, and reduces volume size a lot.
2. The volume names have gotten a lot nicer, for example, `cross-stable-x86_64-unknown-linux-gnu-16b8c-fe5b13d68`. Here, we have `cross-{toolchain_name}-{sysroot_hash}-{rustc_commit}`, where `toolchain_name` is the filename of the toolchain sysroot, `toolchain_hash` is the hash of the path to the sysroot, and `toolchain_commit` is the 9 hex-digit commit identifier of the rustc version used. If we use a non-persistent data volume, we have `cross-{toolchain_name}-{sysroot_hash}-{rustc_commit-{triple}-{name}-{project_hash}`, where `triple` is the target triple, `name` is the project name, and `project_hash` is the hash of the path to the project root. This means everything is unique, with minimal chance for conflict.

However.... #724 is reproducing, and without a way to run `cargo clean` remotely, there's no way to fix this besides just deleting the data volume currently. So we probably need to add `cargo clean` into the list of commands that runs in the container, at least remotely, and this can be quite a showstopper.

Other than that, I'll test this more but I think it's ready.

Alexhuszagh:
For `cargo clean`, it actually only makes sense to run in the image if all the conditions are met, and then it **still** needs to be run outside the container:
- Persistent data volume
- Is remote

So it's worth adding it as a subcommand, but only if all those conditions are met also running it in the container, and then also falling back to the host. So I'll add this as a separate commit to the PR.

Emilgardis:
why do we need to run cargo clean on the host as well?

Alexhuszagh:
> why do we need to run cargo clean on the host as well?

Because it copies the data back over to the host, although I guess we could skip this unless the copy cache is set, since a primary use of cross is to distribute cross-compiled binaries.

Emilgardis:
was mostly wondering, we copy over the target folder so makes sense to me to clean on the host.

Alexhuszagh:
This is ready for review, and adds support for `cargo clean` on the remote host as well. I've tested it works on:
- Windows (not WSL2)
- Linux
- macOS

A few things we can add to our wishlist later:
- Only copy source/project files that have a changed timestamp to minimize network traffic on incremental builds. This would require using some sort of memo like Rust does.
- Tracing for the tempfiles created.

Alexhuszagh:
> Awesome!
> 
> I think this all looks good, one todo is trying to figure out if we're in remote context and spit out a nice hint to use `CROSS_REMOTE`.
> 
> feel free to r= me

That might be good, it can be quite tricky to unambiguously determine if we in a remote context, but it incurs some seriously overhead (see the comment [here](https://github.com/cross-rs/cross/pull/449#issuecomment-1151359408)), which might matter when normally running on localhost.

src/rustc.rs
```diff
@@ -31,6 +32,48 @@ impl VersionMetaExt for VersionMeta {
...
+        .splitn(index + 1, ' ')
+        .nth(index)
+        .and_then(|meta| meta.strip_prefix('('))
+        .and_then(|meta| meta.strip_suffix(')'))
+        .and_then(|meta| meta.split_once(' '))
+    {
+        if is_hash(commit) && is_date(date) {
+            return short_commit_hash(commit);
+        }
+    }
```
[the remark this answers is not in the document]
Alexhuszagh: It's slightly different but based on that code, since I found it difficult to merge the two into a single function.

Alexhuszagh:
bors r-

There's a merge conflict I think, or we might get a bad merge. Going to rebase and merge.

PR #786, opened 2022-06-13
replace build script with rust

Emilgardis:


Alexhuszagh:
Should this be added to `cross-dev`?

Emilgardis:
> Should this be added to `cross-dev`?

what do you mean? it is `cross-dev`

Alexhuszagh:
> > Should this be added to `cross-dev`?
> 
> what do you mean? it is `cross-dev`

My... mistake.

Emilgardis:
bors try

Emilgardis:
bors try

Emilgardis:
bors try

.github/workflows/ci.yml
```diff
@@ -247,14 +252,6 @@ jobs:
...
           RUNNERS: ${{ matrix.runners }}
         shell: bash
-
-      - name: Install cross
-        if: matrix.deploy
-        run: cargo install --path . --force
-
-      - name: Build cross-dev
-        run: cargo build --features=dev --bin cross-dev
-
```
Emilgardis, line 257: this is not needed anymore as we use cross-dev in ci now

Emilgardis: hmm, maybe it should be kept, just replaced with install https://github.com/cross-rs/cross/runs/6853748188?check_suite_focus=true

Emilgardis:
bors try

Cargo.toml
```diff
@@ -11,11 +11,11 @@ edition = "2021"
 
 [features]
 default = []
-dev = ["serde_yaml"]
+dev = ["serde_yaml", "walkdir"]
 
 [dependencies]
 atty = "0.2"
-clap = { version = "3.1.18", features = ["derive"] }
+clap = { version = "3.1.18", features = ["derive", "env"] }
```
Alexhuszagh, line 18: As of clap 3.2.0, this produces numerous spurious warnings that are unrelated to the actual code (I believe, see https://github.com/clap-rs/clap/issues/3822) that may cause CI to fail. It might be worth locking clap to `=3.1.18` for the meantime.

cross-dev/build_docker_image.rs
```diff
@@ -0,0 +1,158 @@
...
+    }: BuildDockerImage,
+    engine: &Path,
+) -> cross::Result<()> {
+    let metadata = cross::cargo_metadata_with_args(
+        Some(Path::new(env!("CARGO_MANIFEST_DIR"))),
+        None,
+        verbose,
+    )?
+    .ok_or_else(|| eyre::eyre!("could not find cross workspace and its current version"))?;
+    let version = metadata
```
Alexhuszagh, line 50: This isn't necessarily correct: if a build script is used or another package is compiled earlier, then this won't be the package desired.

If you have a sample package with a `cc` dependency in a build script, which we might add eventually, then the output of the following will be something as follows:

**Code**

```rust
    let metadata = cargo_metadata_with_args(
        Some(Path::new(env!("CARGO_MANIFEST_DIR"))),
    ).unwrap()
    .ok_or_else(|| eyre::eyre!("could not find cross workspace and its current version")).unwrap();
    eprintln!("metadata.packages={:?}", metadata.packages);
```

**Output**

```text
[Package { id: "cc 1.0.73 (registry+https://github.com/rust-lang/crates.io-index)", manifest_path: "/home/ahuszagh/.cargo/registry/src/github.com-1ecc6299db9ec823/cc-1.0.73/Cargo.toml", source: Some("registry+https://github.com/rust-lang/crates.io-index"), version: "1.0.73" }
```

```

Alexhuszagh:
Other than those 2 comments, lgtm.

cross-dev/build_docker_image.rs
Emilgardis: I can't replicate this scenario, we're running `cargo metadata --no-deps` so the extra package will never show up

Alexhuszagh: Oh that's weird, it wasn't working on mine. Hmmm maybe I thought `--no-deps` was getting added when it wasn't.

Alexhuszagh:

Just testing the clap issues, if it will fail CI.

Alexhuszagh:
> we're using --locked, so it shouldn't

Ok good, I'll probably submit a PR to fix these CI issues.

Alexhuszagh (approved):

Emilgardis:
I can add the pin here, no need for separate pr imo.

PR #791, opened 2022-06-14
more features for `build-docker-image`

Emilgardis:


Emilgardis:
bors try

xtask/Cargo.toml
```diff
@@ -12,7 +12,7 @@ cross = { path = "..", features = ["dev"] }
 walkdir = { version = "2" }
 color-eyre = "0.6"
 eyre = "0.6"
-clap = { version = "~3.1", features = ["derive", "env"] }
+clap = { version = "3.1", features = ["derive", "env", "unstable-v4"] }
```
Emilgardis, line 15: needs to be specified `3.2.2`

Emilgardis:
oh yeah, since we run the ci on a different branch, the annotations won't show up here :/

might have to use gh api for that

Cargo.toml
```diff
@@ -18,10 +18,10 @@ members = ["xtask"]
 
 [dependencies]
 atty = "0.2"
-# Pinned until further action to migrate to newer clap, see https://github.com/clap-rs/clap/issues/3822#issuecomment-1154069623
-clap = { version = "~3.1", features = ["derive"] }
+clap = { version = "3.2.2", features = ["derive", "unstable-v4"] }
```
Alexhuszagh, line 21: Not sure we need this soon, check https://github.com/epage/clap/commit/7cb5bba81132a8840d5ebaba259201d35abe725f. They've changed it so people can opt-in to deprecations which should be released shortly.

Emilgardis: I think it's fine to keep it, when the feature is removed it'll be 4.0, since removing feature flags is breaking behaviour (requiring a major bump)

xtask/src/build_docker_image.rs
```diff
@@ -142,25 +187,77 @@ pub fn build_docker_image(
...
         docker_build.arg(".");
 
-        if force || !push || std::env::var("GITHUB_ACTIONS").is_ok() {
-            docker_build.run(verbose)?;
+        if !dry_run && (force || !push || gha) {
+            let result = docker_build.run(verbose);
+            if gha {
+                if let Err(e) = &result {
+                    // TODO: Determine what instruction errorred, and place warning on that line with appropriate warning
+                    println!("::error file=docker/{dockerfile},title=Build failed::{}", e)
```
Emilgardis, line 203: I'm not sure how valuable this error is 

https://github.com/cross-rs/cross/runs/6881463959?check_suite_focus=true

It would be better if it could capture/tee the stderr/stdout and determine the actual error instead of spitting out what it currently does

Alexhuszagh: Won't it cause a failure in the bors action for that target anyway, so we'll know which specific Dockerfile is failing already? I don't think this is necessary.

.github/actions/cargo-install-upload-artifacts/action.yml
```diff
@@ -13,10 +13,8 @@ runs:
       run: |
```
Alexhuszagh, line 13: Also, the title on this action (on L11) should probably be changed, since we no longer check if the crate has binaries.

.github/actions/cargo-install-upload-artifacts/action.yml
```diff
@@ -13,10 +13,8 @@ runs:
...
         metadata="$(cargo metadata --format-version 1 --no-deps)"
 
-        package_name="$(jq -r '.packages[0].name' <<< "${metadata}")"
-        has_binaries="$(jq '.packages[0].targets | any(.kind | contains(["bin"]))' <<< "${metadata}")"
+        package_name="cross"
         echo "::set-output name=package-name::${package_name}"
-        echo "::set-output name=has-binaries::${has_binaries}"
```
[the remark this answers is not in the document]
Emilgardis: This does not remove the binary artifacts, it simply removes the check for if there is a binary or not.

This is done because rn, `packages[0]` is the xtask workspace member, but if we add another workspace member without a binary, and that happens to become the first package in the metadata, we'd silently not upload an artifact

xtask/src/build_docker_image.rs
Emilgardis: I was hoping that with it, it would show the annotation in the pr, but since it doesn't do that it doesn't make much sense.

There is a reason to keep it however if there are multiple targets, I'll change this

Alexhuszagh (approved):

Alexhuszagh:
I'm fine with it as-is, or if you remove the warning mentioned above. Feel free to `r=me` once you're ready.

Emilgardis:
> I'm fine with it as-is, or if you remove the warning mentioned above. Feel free to `r=me` once you're ready.

I've made it so it's only shown in multi target builds, which we have none of right now.

PR #794, opened 2022-06-14
Refactor docker and cross-util commands.

Alexhuszagh:
Refactor the internals of the docker module and separate the cross-util commands from the actual binary. Will simplify the internals for adding remote container support later.


src/bin/cross-util.rs
```diff
@@ -1,252 +1,61 @@
...
-    fn name(&self) -> String {
-        format!("{}:{}", self.repository, self.tag)
-    }
+    RemoveImages(commands::RemoveImages),
 }
 
-fn get_container_engine(engine: Option<&str>) -> Result<PathBuf, which::Error> {
-    if let Some(ce) = engine {
-        which::which(ce)
+fn is_toolchain(toolchain: &str) -> cross::Result<String> {
```
Alexhuszagh, line 25: Allow us to use `cross-util +nightly`, or similar, rather than `cross-util --toolchain nightly`, which is consistent with the `cross` API.

src/docker/engine.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum EngineType {
+    Docker,
+    Podman,
+    PodmanRemote,
+    Other,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Engine {
```
Alexhuszagh, line 20: Add the engine with the engine type, so we can simplify this. Also adds `PodmanRemote`, which is a separate command and essentially encompasses the same functionality as `podman --remote`.

src/docker/local.rs
```diff
@@ -0,0 +1,93 @@
...
+use super::engine::*;
+use super::shared::*;
+use crate::cargo::CargoMetadata;
+use crate::errors::Result;
+use crate::extensions::CommandExt;
+use crate::{Config, Target};
+use atty::Stream;
+
+#[allow(clippy::too_many_arguments)] // TODO: refactor
+pub(crate) fn local_run(
```
Alexhuszagh, line 13: Rename `run` to `local_run`, so we can use `remote_run` later for if the command is remote.

src/docker/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use std::path::Path;
+use std::process::ExitStatus;
+
+use crate::cargo::CargoMetadata;
+use crate::errors::*;
+use crate::{Config, Target};
+
+#[allow(clippy::too_many_arguments)] // TODO: refactor
+pub fn container_run(
```
Alexhuszagh, line 18: Rename `run` to `container_run`, so we prefix it correctly since we'll need to use glob imports or very lengthy imports in the future.

Emilgardis (changes requested):
one nit, otherwise looks good

src/lib.rs
```diff
@@ -44,8 +44,7 @@ use self::cross_toml::CrossToml;
 use self::errors::Context;
 use self::rustc::{TargetList, VersionMetaExt};
 
-pub use self::docker::get_container_engine;
-pub use self::docker::CROSS_IMAGE;
+pub use self::docker::*;
```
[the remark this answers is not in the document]
Emilgardis: oh yes, definitely. 

so local_run would be invoked as `docker::local::run` instead. That makes sense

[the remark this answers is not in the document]
Alexhuszagh: Ok I've removed the glob import, and made `docker` a public module. I've also removed the prefixes for most of the public commands.

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #795, opened 2022-06-14
Add additional toolchains maintained by cross-rs.

Alexhuszagh:
These toolchains are not automatically tested by CI, and must be built manually by the user. However, they are confirmed to build at least once, and are likely to work as-is.

### DIFF
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -48,6 +48,23 @@ pub struct BuildDockerImage {
     targets: Vec<String>,
 }
 
+fn locate_dockerfile(
+    target: String,
+    docker_root: &Path,
+    cross_toolchain_root: &Path,
+) -> cross::Result<(String, String)> {
+    let dockerfile_name = format!("Dockerfile.{target}");
+    let dockerfile_root = if cross_toolchain_root.join(&dockerfile_name).exists() {
+        &cross_toolchain_root
+    } else if docker_root.join(&dockerfile_name).exists() {
+        &docker_root
+    } else {
+        eyre::bail!("unable to find dockerfile for target \"{target}\"");
+    };
+    let dockerfile = dockerfile_root.join(dockerfile_name).display().to_string();
+    Ok((target, dockerfile))
+}
+
 pub fn build_docker_image(
     BuildDockerImage {
         ref_type,
@@ -102,22 +119,27 @@ pub fn build_docker_image(
         }
     }
     let gha = std::env::var("GITHUB_ACTIONS").is_ok();
+    let docker_root = metadata.workspace_root.join("docker");
+    let cross_toolchains_root = docker_root.join("cross-toolchains").join("docker");
+    let targets = targets
+        .into_iter()
+        .map(|t| locate_dockerfile(t, &docker_root, &cross_toolchains_root))
+        .collect::<cross::Result<Vec<_>>>()?;
+
     let mut results = vec![];
-    for target in &targets {
+    for (target, dockerfile) in &targets {
         if gha && targets.len() > 1 {
             println!("::group::Build {target}");
         }
         let mut docker_build = Command::new(engine);
         docker_build.args(&["buildx", "build"]);
-        docker_build.current_dir(metadata.workspace_root.join("docker"));
+        docker_build.current_dir(&docker_root);
 
         if push {
             docker_build.arg("--push");
         } else {
             docker_build.arg("--load");
         }
-
-        let dockerfile = format!("Dockerfile.{target}");
 
         let (target, sub) = if let Some((t, sub)) = target.split_once('.') {
             (t.to_owned(), format!("-{sub}"))
@@ -192,7 +214,7 @@ pub fn build_docker_image(
             docker_build.args(&["--label", label]);
         }
 
-        docker_build.args(&["-f", &dockerfile]);
+        docker_build.args(&["-f", dockerfile]);
 
         if gha || progress == "plain" {
             docker_build.args(&["--progress", "plain"]);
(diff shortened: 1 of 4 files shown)


xtask/src/build_docker_image.rs
```diff
@@ -107,17 +107,25 @@ pub fn build_docker_image(
...
         } else {
             docker_build.arg("--load");
         }
 
-        let dockerfile = format!("Dockerfile.{target}");
+        let dockerfile_name = format!("Dockerfile.{target}");
+        let dockerfile_root = if docker_root.join(&dockerfile_name).exists() {
+            docker_root.clone()
+        } else {
+            docker_root.join("cross-toolchains").join("docker")
```
Emilgardis, line 125: this would make the `docker buildx build` say that a file at `cross-toolchains` does not exist. I think it's better to instead of doing this, to check if the file exists at `cross-toolchains`, and if it does not fallback to `docker/`.

Emilgardis, line 125: seems we're always using the string version of the file, so it could be converted to one here already

Emilgardis (changes requested):

xtask/src/build_docker_image.rs
Alexhuszagh: Done, however, the error message is still fairly opaque. Here's a sample where I replace `s390x-unknown-linux-gnu-cross` with `s390x-unknown-linux-gnu-crossx`:

```bash
+ "/home/ahuszagh/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo" "metadata" "--format-version=1" "--no-deps"                                                                         
+ "/usr/bin/docker" "buildx" "build" "--load" "--pull" "--cache-from" "type=registry,ref=ghcr.io/cross-rs/s390x-unknown-linux-gnu-crossx:main" "--tag" "ghcr.io/cross-rs/s390x-unknown-linux-gnu-cro
ssx:local" "-f" "/home/ahuszagh/git/cross/docker/Dockerfile.s390x-unknown-linux-gnu-crossx" "--progress" "auto" "."                                                                                 
[+] Building 0.2s (2/2) FINISHED                                                                                                                                                                    
 => [internal] load build definition from Dockerfile.s390x-unknown-linux-gnu-crossx                                                                                                            0.1s 
 => => transferring dockerfile: 2B                                                                                                                                                             0.0s 
 => [internal] load .dockerignore                                                                                                                                                              0.1s 
 => => transferring context: 2B                                                                                                                                                                0.0s 
error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount1322602709/Dockerfile.s390x-
unknown-linux-gnu-crossx: no such file or directory                                                                                                                                                 
Error:                                                                                                                                                                                              
   0: encountered error(s)                                                                                                                                                                          
                                                                                                                                                                                                    
Location:                                                                                                                                                                                           
   xtask/src/build_docker_image.rs:245                                                                                                                                                              
                                                                                                                                                                                                    
Error:                                                                                                                                                                                              
   0: `"/usr/bin/docker" "buildx" "build" "--load" "--pull" "--cache-from" "type=registry,ref=ghcr.io/cross-rs/s390x-unknown-linux-gnu-crossx:main" "--tag" "ghcr.io/cross-rs/s390x-unknown-linux-gn
u-crossx:local" "-f" "/home/ahuszagh/git/cross/docker/Dockerfile.s390x-unknown-linux-gnu-crossx" "--progress" "auto" "."` failed with exit code: exit status: 1                                     
[ahuszagh@ahuszagh cross]$ git log 
```

xtask/src/build_docker_image.rs
```diff
@@ -107,17 +107,28 @@ pub fn build_docker_image(
...
+        let dockerfile_name = format!("Dockerfile.{target}");
+        let cross_toolchains_root = docker_root.join("cross-toolchains").join("docker");
+        let dockerfile_root = if cross_toolchains_root.join(&dockerfile_name).exists() {
+            &cross_toolchains_root
+        } else if docker_root.join(&dockerfile_name).exists() {
+            &docker_root
+        } else {
+            eyre::bail!("unable to find dockerfile for target \"{target}\"");
+        };
+        let dockerfile = dockerfile_root.join(dockerfile_name).display().to_string();
```
Emilgardis, line 130: can we move this check to before the loop?

```rs
fn locate_file(target: &String, cross_toolchain_root: &Path) -> {
... (this changeset)
}

targets = targets.into_iter().map(|target| locate_file(&target, cross_toolchain_root)).collect::<Result<Vec<_>>>()?;

for target in &targets {
    ...
}
```

Alexhuszagh:
I think that's everything. We now locate the Dockerfiles ahead of time, allowing mixing and matching in the docker root or the cross-toolchain docker root:

```rust
fn locate_dockerfile(
    target: String,
    docker_root: &Path,
    cross_toolchain_root: &Path,
) -> cross::Result<(String, String)> {
    let dockerfile_name = format!("Dockerfile.{target}");
    let dockerfile_root = if cross_toolchain_root.join(&dockerfile_name).exists() {
        &cross_toolchain_root
    } else if docker_root.join(&dockerfile_name).exists() {
        &docker_root
    } else {
        eyre::bail!("unable to find dockerfile for target \"{target}\"");
    };
    let dockerfile = dockerfile_root.join(dockerfile_name).display().to_string();
    Ok((target, dockerfile))
}

...

    let targets = targets
        .into_iter()
        .map(|t| locate_dockerfile(t, &docker_root, &cross_toolchains_root))
        .collect::<cross::Result<Vec<_>>>()?;
```

Emilgardis (approved):
bors r+

PR #801, opened 2022-06-15
Make config vectors optional.

Alexhuszagh:
Allows us to differentiate between an empty array and an array that was not provided when merging config files.

https://github.com/cross-rs/cross/pull/754#issuecomment-1156722398

Required for #754.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -179,8 +179,8 @@ impl Config {
         &self,
         target: &Target,
         env: impl Fn(&Environment, &Target) -> (Option<Vec<String>>, Option<Vec<String>>),
-        config_build: impl for<'a> Fn(&'a CrossToml) -> &'a [String],
-        config_target: impl for<'a> Fn(&'a CrossToml, &Target) -> &'a [String],
+        config_build: impl for<'a> Fn(&'a CrossToml) -> Option<&'a [String]>,
+        config_target: impl for<'a> Fn(&'a CrossToml, &Target) -> Option<&'a [String]>,
     ) -> Result<Vec<String>> {
         let (env_build, env_target) = env(&self.env, target);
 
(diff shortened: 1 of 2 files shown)


src/cross_toml.rs
```diff
@@ -9,9 +9,9 @@ use std::collections::{BTreeSet, HashMap};
 #[derive(Debug, Deserialize, PartialEq, Eq, Default)]
 pub struct CrossEnvConfig {
     #[serde(default)]
-    volumes: Vec<String>,
+    volumes: Option<Vec<String>>,
     #[serde(default)]
-    passthrough: Vec<String>,
+    passthrough: Option<Vec<String>>,
```
Emilgardis, line 14: the `#[serde(default)]` shouldn't be needed anymore

src/cross_toml.rs
```diff
@@ -89,22 +89,22 @@ impl CrossToml {
...
     pub fn env_volumes_build(&self) -> &[String] {
-        &self.build.env.volumes
+        self.build.env.volumes.as_deref().unwrap_or(&[])
     }
 
     /// Returns the list of environment variables to pass through for `target`,
     pub fn env_volumes_target(&self, target: &Target) -> &[String] {
-        self.get_vec(target, |e| &e.volumes)
+        self.get_vec(target, |e| e.volumes.as_deref().unwrap_or(&[]))
     }
```
Emilgardis, line 108: I think these should return opt as well, instead the decision to throw away this fact should be done at the caller

Emilgardis (changes requested):

src/cross_toml.rs
Alexhuszagh: I had those originally as Options, it makes the calling code a bit more difficult (specifically, the accumulation code). I can revert it back, however.

Emilgardis: I've made some changes in #678 that touches on that as well, fyi

src/config.rs
```diff
@@ -241,12 +241,12 @@ impl Config {
...
         &'a self,
         env_values: Option<Vec<String>>,
-        toml_getter: impl FnOnce(&'a CrossToml) -> &'a [String],
+        toml_getter: impl FnOnce(&'a CrossToml) -> Option<&'a [String]>,
     ) -> Result<Vec<String>> {
         let mut collect = vec![];
         if let Some(mut vars) = env_values {
             collect.append(&mut vars);
-        } else if let Some(toml_values) = self.toml.as_ref().map(toml_getter) {
+        } else if let Some(Some(toml_values)) = self.toml.as_ref().map(toml_getter) {
```
Emilgardis, line 249: very nice, buut

```suggestion
        } else if let Some(toml_values) = self.toml.as_ref().and_then(toml_getter) {
```

src/cross_toml.rs
Alexhuszagh: > I've made some changes in #678 that touches on that as well, fyi

Currently reviewing that as we speak, I'll look into that shortly.

Emilgardis (changes requested):

src/config.rs
Alexhuszagh: Even better...

Emilgardis: This is a good oppurtunity for a clippy lint

Emilgardis (approved):
bors r+

PR #802, opened 2022-06-15
more hooks

Emilgardis:
- add pre-push hook that tests
- fix rustfmt edition


Alexhuszagh:
Might also be worth adding a shellcheck, if installed?

Emilgardis:
oo, smart

xtask/src/pre-commit
```diff
@@ -1,13 +0,0 @@
...
-
-set -ex
-
-flags=(--all-features --all-targets --workspace)
-cargo fmt -- --check
-if cargo +nightly >/dev/null 2>&1; then
-    cargo +nightly clippy "${flags[@]}" -- --deny warnings
-else
-    cargo clippy "${flags[@]}" -- --deny warnings
-fi
```
Emilgardis, line 12: why does git think this file was removed...

ill fix it

Emilgardis: ...

Alexhuszagh: No idea, it has the proper permissions at least so it seems fine.

Alexhuszagh (approved):

Emilgardis:
bors r-

just a minute, think there's something wrong

xtask/src/pre-commit.sh
```diff
@@ -0,0 +1,24 @@
...
+    cargo clippy "${flags[@]}" -- --deny warnings
+fi
+
+if shellcheck -V >/dev/null 2>&1; then
+    echo "Running shellcheck"
+    changed_shell_files=$(
+        git diff --cached --name-only --diff-filter=ACM |
+        xargs grep -lE '^#!/.*(sh|bash|ksh)'
+    )
+    if [ -n "${changed_shell_files[*]}" ]; then
```
Emilgardis, line 21: this check is not working when there's no changed sh files

PR #587, opened 2021-09-04
Update to latest stable Qemu release. Fixes #557.

Alexhuszagh:
Update Qemu version to the latest stable release, v6.1.0. The latest Qemu release contains numerous bug fixes, and other improvements, and the existing patches various issues reported in older Qemu versions.

The only changes required are incrementing the Qemu version and adding `ninja-build` as a temporary dependency, since Qemu now uses it for the build system.

PR #803, opened 2022-06-16
add `CROSS_CUSTOM_TOOLCHAIN` to disable automatic target component downloading

### LINKED ISSUE #699: How to use cross with custom toolchains?
**Maintainer's Note:** The original title was "How to use cargo-bisect-rustc with cross?", however, the issue is more general and deals with working with custom toolchains broadly, or at least specifically custom toolchains that are installed from Rust sources.

----

When running `cargo-bisect-rustc -vv --script=./test.sh` where `test.sh` is:

```
#!/bin/sh

exec cross run --target powerpc64le-unknown-linux-gnu
```

The `cross` command failed with:

```
error: `"rustup" "target" "list" "--toolchain" "bisector-nightly-2022-04-26-x86_64-unknown-linux-gnu"` failed with exit code: Some(1)
```

If I just run `cargo-bisect-rustc -vv` without using `cross`, then everything runs without problem.

Is there any way to fix the problem?

Emilgardis:
This adds an environment variable `CROSS_CUSTOM_TOOLCHAIN` to disable some rustup commands.

The way to use this with `cargo-bisect-rustc` is `cargo bisect-rustc --script=./bisect.sh --target powerpc64le-unknown-linux-gnu` and the script is

```sh
#!/usr/bin/env bash

export CROSS_CUSTOM_TOOLCHAIN=1
exec cross run --target powerpc64le-unknown-linux-gnu
```

I've filed https://github.com/rust-lang/cargo-bisect-rustc/pull/159 to avoid the pitfall that was discovered when not specifying `--target`

resolves #699

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #803 - added `CROSS_CUSTOM_TOOLCHAIN` to disable automatic installation of components for use with tools like `cargo-bisect-rustc`
 - #795 - added images for additional toolchains maintained by cross-rs.
 - #792 - added `CROSS_CONTAINER_IN_CONTAINER` environment variable to replace `CROSS_DOCKER_IN_DOCKER`.
 - #782 - added `build-std` config option, which builds the rust standard library from source if enabled.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -416,35 +416,37 @@ pub fn run() -> Result<ExitStatus> {
                 is_nightly = channel == Channel::Nightly;
             }
 
-            // build-std overrides xargo, but only use it if it's a built-in
-            // tool but not an available target or doesn't have rust-std.
-            let available_targets = rustup::available_targets(&toolchain, verbose)?;
             let uses_build_std = config.build_std(&target).unwrap_or(false);
             let uses_xargo =
                 !uses_build_std && config.xargo(&target).unwrap_or(!target.is_builtin());
-            if !is_nightly && uses_build_std {
-                eyre::bail!(
-                    "no rust-std component available for {}: must use nightly",
-                    target.triple()
-                );
-            }
+            if std::env::var("CROSS_CUSTOM_TOOLCHAIN").is_err() {
+                // build-std overrides xargo, but only use it if it's a built-in
+                // tool but not an available target or doesn't have rust-std.
+                let available_targets = rustup::available_targets(&toolchain, verbose)?;
 
-            if !uses_xargo
-                && !available_targets.is_installed(&target)
-                && available_targets.contains(&target)
-            {
-                rustup::install(&target, &toolchain, verbose)?;
-            } else if !rustup::component_is_installed("rust-src", &toolchain, verbose)? {
-                rustup::install_component("rust-src", &toolchain, verbose)?;
-            }
+                if !is_nightly && uses_build_std {
+                    eyre::bail!(
+                        "no rust-std component available for {}: must use nightly",
+                        target.triple()
+                    );
+                }
 
-            if args
-                .subcommand
-                .map(|sc| sc == Subcommand::Clippy)
-                .unwrap_or(false)
-                && !rustup::component_is_installed("clippy", &toolchain, verbose)?
-            {
-                rustup::install_component("clippy", &toolchain, verbose)?;
+                if !uses_xargo
+                    && !available_targets.is_installed(&target)
+                    && available_targets.contains(&target)
+                {
+                    rustup::install(&target, &toolchain, verbose)?;
+                } else if !rustup::component_is_installed("rust-src", &toolchain, verbose)? {
+                    rustup::install_component("rust-src", &toolchain, verbose)?;
+                }
+                if args
+                    .subcommand
+                    .map(|sc| sc == Subcommand::Clippy)
+                    .unwrap_or(false)
+                    && !rustup::component_is_installed("clippy", &toolchain, verbose)?
+                {
+                    rustup::install_component("clippy", &toolchain, verbose)?;
+                }
             }
 
             let needs_interpreter = args
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -4,7 +4,7 @@ use std::process::Command;
 use rustc_version::{Channel, Version};
 
 use crate::errors::*;
-use crate::extensions::CommandExt;
+pub use crate::extensions::{CommandExt, OutputExt};
 use crate::Target;
 
 #[derive(Debug)]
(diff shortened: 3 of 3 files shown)


Emilgardis:
ping @NobodyXu

does this work for you? I've been trying to replicate your errors but I can't manage to do it.

Anyway, this should solve the issue of using cross with a custom toolchain (that is rustup compatible)

Alexhuszagh:
I don't think this will work because the custom toolchains are installed dynamically and won't necessarily have the desired target added for cross-compiling. This would work if the custom toolchain was installed by the user with every feature desired already, but `cargo-bisect-rustc` doesn't do that I believe.

Emilgardis:
`cargo-bisect-rustc` plays nice here and installs everything that we explicitly install ourselves in the commands "protected" by the env var (except clippy)

I think I've tried it enough, it works

Emilgardis:
The way to use this with `cargo-bisect-rustc` is `cargo bisect-rustc --script=./bisect.sh --target powerpc64le-unknown-linux-gnu` and the script is

```sh
#!/usr/bin/env bash

export CROSS_CUSTOM_TOOLCHAIN=1
exec cross run --target powerpc64le-unknown-linux-gnu
```

NobodyXu:
Let me have a try.

Alexhuszagh:
Once I finish a few commits, I'll try as well with a guaranteed failing command and see how it works.

NobodyXu:
Working on my machine:

```shell
$ cargo bisect-rustc -vv --script=./test.sh --target powerpc64le-unknown-linux-gnu
installing nightly-2022-06-16
testing...
+ export CROSS_CUSTOM_TOOLCHAIN=1
+ CROSS_CUSTOM_TOOLCHAIN=1
+ exec cross run -vv --target powerpc64le-unknown-linux-gnu
+ "/home/nobodyxu/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo" "metadata" "--format-version=1" "--filter-platform" "powerpc64le-unknown-linux-gnu"
+ "rustc" "--print" "sysroot"
+ "rustup" "toolchain" "list"
+ "/usr/bin/docker" "--help"
+ "/usr/bin/docker" "run" "--userns" "host" "-e" "PKG_CONFIG_ALLOW_CROSS=1" "-e" "XARGO_HOME=/xargo" "-e" "CARGO_HOME=/cargo" "-e" "CARGO_TARGET_DIR=/target" "-e" "CROSS_RUNNER=" "-e" "USER=nobodyxu" "--rm" "--user" "1000:1000" "-v" "/home/nobodyxu/.xargo:/xargo:Z" "-v" "/home/nobodyxu/.cargo:/cargo:Z" "-v" "/cargo/bin" "-v" "/home/nobodyxu/tests:/project:Z" "-v" "/home/nobodyxu/.rustup/toolchains/bisector-nightly-2022-06-16-x86_64-unknown-linux-gnu:/rust:Z,ro" "-v" "/home/nobodyxu/tests/target-bisector-nightly-2022-06-16-x86_64-unknown-linux-gnu:/target:Z" "-w" "/project" "ghcr.io/cross-rs/powerpc64le-unknown-linux-gnu:main" "sh" "-c" "PATH=$PATH:/rust/bin cargo run -vv --target powerpc64le-unknown-linux-gnu"
   Compiling tests v0.1.0 (/project)
     Running `CARGO=/rust/bin/cargo CARGO_BIN_NAME=tests CARGO_CRATE_NAME=tests CARGO_MANIFEST_DIR=/project CARGO_PKG_AUTHORS='' CARGO_PKG_DESCRIPTION='' CARGO_PKG_HOMEPAGE='' CARGO_PKG_LICENSE='' CARGO_PKG_LICENSE_FILE='' CARGO_PKG_NAME=tests CARGO_PKG_REPOSITORY='' CARGO_PKG_RUST_VERSION='' CARGO_PKG_VERSION=0.1.0 CARGO_PKG_VERSION_MAJOR=0 CARGO_PKG_VERSION_MINOR=1 CARGO_PKG_VERSION_PATCH=0 CARGO_PKG_VERSION_PRE='' CARGO_PRIMARY_PACKAGE=1 LD_LIBRARY_PATH='/target/debug/deps:/rust/lib' rustc --crate-name tests --edition=2018 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C metadata=1eaf9da53e7e178c -C extra-filename=-1eaf9da53e7e178c --out-dir /target/powerpc64le-unknown-linux-gnu/debug/deps --target powerpc64le-unknown-linux-gnu -C linker=powerpc64le-linux-gnu-gcc -C incremental=/target/powerpc64le-unknown-linux-gnu/debug/incremental -L dependency=/target/powerpc64le-unknown-linux-gnu/debug/deps -L dependency=/target/debug/deps`
    Finished dev [unoptimized + debuginfo] target(s) in 0.55s
     Running `/linux-runner powerpc64le /target/powerpc64le-unknown-linux-gnu/debug/tests`
3.2
-inf
RESULT: nightly-2022-06-16, ===> No
uninstalling nightly-2022-06-16
```

and my `test.sh`:

```
#!/bin/bash

set -euxo pipefail

export CROSS_CUSTOM_TOOLCHAIN=1
exec cross run -vv --target powerpc64le-unknown-linux-gnu
```

Emilgardis:
> Worked on my machine:

Awesome, however the interesting part is when it installs an "non-normal" toolchain, e.g one that rustup reports as a custom toolchain (e.g installed with `rustup toolchain link`)

In an unrelated note to this specific pr, I can't replicate the behaviour you suggest in your rustc issue, e.g I can't find a date which reproduces the issue of `3.2_f32` printing as `0`

Alexhuszagh:
Nevermind this does work as long as you provide the `--target` both on the outside and inside. lgtm. I'll add this to the things-to-add-to-the-wiki as well.

Emilgardis:
> Nevermind this does work as long as you provide the `--target` both on the outside and inside. lgtm. I'll add this to the things-to-add-to-the-wiki as well.

don't think you need to add it, I've filed https://github.com/rust-lang/cargo-bisect-rustc/pull/159 which would fix it automagically

Alexhuszagh:
> > Worked on my machine:
> 
> Awesome, however the interesting part is when it installs an "non-normal" toolchain, e.g one that rustup reports as a custom toolchain (e.g installed with `rustup toolchain link`)
> 
> In an unrelated note to this specific pr, I can't replicate the behaviour you suggest in your rustc issue, e.g I can't find a date which reproduces the issue of `3.2_f32` printing as `0`

This was a Qemu error 🙃. This was actually the exact bug for the exact target I started contributing to this repository in the first place.

NobodyXu:
> > > Worked on my machine:
> > 
> > 
> > Awesome, however the interesting part is when it installs an "non-normal" toolchain, e.g one that rustup reports as a custom toolchain (e.g installed with `rustup toolchain link`)
> > In an unrelated note to this specific pr, I can't replicate the behaviour you suggest in your rustc issue, e.g I can't find a date which reproduces the issue of `3.2_f32` printing as `0`
> 
> This was a Qemu error 🙃. This was actually the exact bug for the exact target I started contributing to this repository in the first place.

Oops

Never thought it was a Qemu error...
Is that because the Qemu used by `cross` is too old?

Alexhuszagh (approved):
lgtm. I'm assuming we want to just check if the envvar is set, rather than it's truthy?

Emilgardis:
the specific qemu error was fixed as part of #609

there might have been a mix of qemu errors and alignment errors on rustc, hard to say for sure without knowing exactly what was done

Alexhuszagh:
> 

The old releases for cross used a much older Qemu version. We then pinned it to Qemu 3.x because of a similar bug, but this causes other issues that were fixed in 4.x. We now installed off-of-main use 5.x, which fixes all these issues.

Emilgardis:
> lgtm. I'm assuming we want to just check if the envvar is set, rather than it's truthy?

I think (hope) no one sets it when not needed, worst case that happens when it's set is that the user doesn't get easy downloading of needed components

lets ship it

bors r=Alexhuszagh

NobodyXu:
It totally didn't occur to me that it involves Qemu bugs

Emilgardis:
> It totally didn't occur to me that it involves Qemu bugs

;) do note also that cross, while using newer qemu, doesn't still use the newest, for that we need #587

PR #805, opened 2022-06-16
Add shellcheck to git hooks and convert to xtask.

Alexhuszagh:
Convert git hooks to use `xtask` commands, and add `shellcheck` to `xtask check`. Can run on staged or all files.

### DIFF
--- a/xtask/src/hooks.rs
+++ b/xtask/src/hooks.rs
@@ -1,0 +1,161 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader, ErrorKind};
+use std::path::Path;
+use std::process::Command;
+
+use clap::Args;
+use cross::CommandExt;
+
+const CARGO_FLAGS: &[&str] = &["--all-features", "--all-targets", "--workspace"];
+
+#[derive(Args, Debug)]
+pub struct Check {
+    /// Provide verbose diagnostic output.
+    #[clap(short, long)]
+    verbose: bool,
+    /// Run shellcheck on all files, not just staged files.
+    #[clap(short, long)]
+    all: bool,
+}
+
+#[derive(Args, Debug)]
+pub struct Test {
+    /// Provide verbose diagnostic output.
+    #[clap(short, long)]
+    verbose: bool,
+}
+
+fn has_nightly(verbose: bool) -> cross::Result<bool> {
+    Command::new("cargo")
+        .arg("+nightly")
+        .run_and_get_output(verbose)
+        .map(|o| o.status.success())
+        .map_err(Into::into)
+}
+
+fn get_channel_prefer_nightly(
+    verbose: bool,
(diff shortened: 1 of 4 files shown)


xtask/src/hooks.rs
```diff
@@ -0,0 +1,161 @@
...
+fn is_shell_script(path: impl AsRef<Path>) -> cross::Result<bool> {
+    if path.as_ref().is_dir() {
+        // is a directory if a git submodule
+        return Ok(false);
+    }
+    let file = File::open(path.as_ref())?;
+    let reader = BufReader::new(file);
+
+    match reader.lines().next() {
+        Some(Ok(line)) => Ok(line.starts_with("#!") && line.contains("sh")),
```
Alexhuszagh, line 110: A more sophisticated way would be to actually evaluate the shebang line, but this is identical to what we did before, and it's going to work unless we **really** try to mess it up, so I think this is fine.

xtask/src/hooks.rs
```diff
@@ -0,0 +1,161 @@
...
+    verbose: bool,
+}
+
+fn has_nightly(verbose: bool) -> cross::Result<bool> {
+    Command::new("cargo")
+        .arg("+nightly")
+        .run_and_get_output(verbose)
+        .map(|o| o.status.success())
+        .map_err(Into::into)
+}
```
Emilgardis, line 34: technically, this should be
```suggestion
fn has_nightly(verbose: bool) -> cross::Result<bool> {
    Command::new("cargo")
        .arg("+nightly")
        .run_and_get_output(verbose)
        .map(|o| o.stdout.contains("-nightly"))
        .map_err(Into::into)
}
```

but that's just being very pedantic

Emilgardis (approved):

xtask/src/hooks.rs
```diff
@@ -0,0 +1,161 @@
...
+                .args(&scripts)
+                .run(verbose, false)?;
+        }
+    }
+
+    Ok(())
+}
+
+pub fn check(Check { verbose, all }: Check, toolchain: Option<&str>) -> cross::Result<()> {
+    println!("Running rustfmt and clippy checks.");
```
Emilgardis, line 143: missing shellsheck

xtask/src/hooks.rs
Alexhuszagh: Why though? I've just checked that running `cargo +nightly` produces a non-zero error code: if it works, it works. If someone wants to rename it something funny, then sure. If I do:

```bash
$ cargo +sofa
error: toolchain 'sofa' is not installed
```

Emilgardis: exactly, it's very uneccessary :)

Alexhuszagh:
bors r=Emilgardis

PR #806, opened 2022-06-16
improve error messages

Emilgardis:
this makes error messages a bit better by displaying stdout/stderr if it was captured

also improves command output in verbose mode or in errors

new:
the new output on an error on verbose mode

old:
the old output on an error on verbose mode, has a some unnecessary double quoutes and doesn't tell what the actual reported error was/is

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -28,7 +28,7 @@ pub enum CommandError {
 }
 
 impl CommandError {
-    /// Attach valuable information to this CommandError
+    /// Attach valuable information to this [`CommandError`](Self)
     pub fn to_section_report(self) -> eyre::Report {
         match &self {
             CommandError::NonZeroExitCode { stderr, stdout, .. } => {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -10,10 +10,35 @@ pub fn install_panic_hook() -> Result<()> {
 
 #[derive(Debug, thiserror::Error)]
 pub enum CommandError {
-    #[error("`{1}` failed with exit code: {0}")]
-    NonZeroExitCode(std::process::ExitStatus, String),
-    #[error("could not execute `{0}`")]
-    CouldNotExecute(#[source] Box<dyn std::error::Error + Send + Sync>, String),
+    #[error("`{command}` failed with exit code: {status}")]
+    NonZeroExitCode {
+        status: std::process::ExitStatus,
+        command: String,
+        stderr: Vec<u8>,
+        stdout: Vec<u8>,
+    },
+    #[error("could not execute `{command}`")]
+    CouldNotExecute {
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync>,
+        command: String,
+    },
     #[error("`{0:?}` output was not UTF-8")]
     Utf8Error(#[source] std::string::FromUtf8Error, std::process::Output),
+}
+
+impl CommandError {
+    /// Attach valuable information to this CommandError
+    pub fn to_section_report(self) -> eyre::Report {
+        match &self {
+            CommandError::NonZeroExitCode { stderr, stdout, .. } => {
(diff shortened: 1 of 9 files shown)


Emilgardis:
sample (artificial):

image

Emilgardis:
converting to draft, I want to try to remove the uneccessary quotes when possible

Emilgardis:
new:
image

Emilgardis:
old:
image

Alexhuszagh:
I think this needs to fix CWD quoting when building Docker images, which I still think is worth it.

2022-06-16 09_28_42-Window

Alexhuszagh:
Only issue I see is for some commands, if an absolute path is provided for the executable, then it isn't quoted. This might get confusing in a few cases, say, for long toolchain install locations.

2022-06-16 09_28_42-Window

Alexhuszagh:
This should work, I'll add unittests and then commit it, you can squash the commit history as needed:

```rust
fn pretty_path(path: impl AsRef<Path>) -> String {
    // Need to remove UNC paths on windows.
    let path = file::canonicalize(path.as_ref()).unwrap_or_else(|_| path.as_ref().to_path_buf());
    let string = path.display().to_string();
    if string.contains('"') {
        format!("{path:?}")
    } else if string.contains(' ') {
        format!("\"{string}\"")
    } else {
        string
    }
}
```

Emilgardis:
ill add it directly 👍🏽

src/file.rs
```diff
@@ -57,3 +108,55 @@ pub fn write_file(path: impl AsRef<Path>, overwrite: bool) -> Result<File> {
     open.open(&path)
         .wrap_err(format!("couldn't write to file `{}`", path.display()))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
Alexhuszagh, line 114: Oh we should probably have a platform-independent test too, just for simple stuff. Ensures it works on macOS as well.

Emilgardis: does mac os have any quirks not found in *nix?

src/extensions.rs
```diff
@@ -65,7 +81,8 @@ impl CommandExt for Command {
         } else {
             Err(CommandError::NonZeroExitCode {
                 status,
-                command: self.command_pretty(),
+                command: self
+                    .command_pretty(verbose, |ref cmd| STRIPPED_BINS.iter().any(|f| f == cmd)),
```
Alexhuszagh, line 85: Does this work? Don't we provide commands using the path of docker? Also, this should probably also support `cargo` as well. For example, we do `which::which`. We could map the values for the path to the actual unprefixed value?

Emilgardis: The const includes cargo, so it does it already

it works as in we never display `/usr/bin/cargo` unless we're in verbose

Emilgardis: > I guess we could just discard the result of which::which, since it has to be in-the-path anyway, or provided as a full path if using CROSS_CONTAINER_ENGINE and not in-the-path.

That makes sense to me, but there could be a point to having the full path.

Alexhuszagh (approved):

src/errors.rs
```diff
@@ -10,10 +10,37 @@ pub fn install_panic_hook() -> Result<()> {
...
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync>,
+        command: String,
+    },
     #[error("`{0:?}` output was not UTF-8")]
     Utf8Error(#[source] std::string::FromUtf8Error, std::process::Output),
 }
+
+//  color_eyre::SectionExt::header(stderr.trim().to_string(), "Stderr:")
+
```
Emilgardis, line 31: oops!

Emilgardis (commented):
bors r-

Alexhuszagh:
Ah feel free to merge it whenever.

Emilgardis:
bors r=Alexhuszagh

PR #807, opened 2022-06-16
Update Qemu version to v6.1.0 if possible.

Alexhuszagh:
The latest Qemu release contains numerous bug fixes, and other improvements, and the existing patches various issues reported in older Qemu versions. The only changes required are incrementing the Qemu version and adding ninja-build as a temporary dependency, since Qemu now uses it for the build system. However, Ubuntu 16.04, the base image for numerous dependencies we use, still ships with Python3.5, and Qemu 6.1.0 requires Python3.6+, as well as the ninja-build system. We therefore add a check to see if we have a compatible Python version, and if we do, build the last Qemu version, so our code can be fully backwards-compatible.

Closes #587.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- #807 - update Qemu to 6.1.0 on images using Ubuntu 18.04+ with python3.6+.
 - #775 - forward Cargo exit code to host
 - #762 - re-enabled `x86_64-unknown-dragonfly` target.
 - #747 - reduced android image sizes.
--- a/README.md
+++ b/README.md
@@ -302,49 +302,49 @@ terminate.
 
 | Target                               |  libc  |   GCC   | C++ | QEMU  | `test` |
 |--------------------------------------|-------:|--------:|:---:|------:|:------:|
-| `aarch64-linux-android` [1]          | 9.0.8  | 9.0.8   | ✓   | 5.1.0 |   ✓    |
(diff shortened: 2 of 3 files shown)


Alexhuszagh:
If this looks good, and passes the tests, please don't merge it (yet) since I still have to update the README.

Alexhuszagh:
There's so many compatible targets it makes no sense to do a partial test.

bors try

Alexhuszagh:
Failed due to an unrelated [error](https://github.com/cross-rs/cross/runs/6914226589?check_suite_focus=true#step:8:44) (TLS handshake timeout while fetching from the registry), but all the MUSL images and a few others seem to work, which means it should work for everything.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- #807 - update Qemu to 6.1.0 on images using Ubuntu 18.04+.
```
Emilgardis, line 32: could mention the reason for why 18.04, so maybe
```suggestion
- #807 - update Qemu to 6.1.0 on images using Ubuntu 18.04+ with python3.6+.
```

Emilgardis (approved):

Emilgardis:
r=me when readme updated, feel free to steal the ci and use it to parallel build and grab the versions to make sure it's correct, we should maybe even do that unconditionally for no other reason than curiosity now that we have a great way to do it

Alexhuszagh:
bors r=Emilgardis

PR #808, opened 2022-06-16
Update `rustc` and `cargo` commands to use envvars.

Alexhuszagh:
`rustc` and `cargo` will now use `RUSTC` and `CARGO` environment variables, respectively, if present.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -162,7 +162,7 @@ pub fn parse_docker_opts(value: &str) -> Result<Vec<String>> {
     shell_words::split(value).wrap_err_with(|| format!("could not parse docker opts of {}", value))
 }
 
-pub(crate) fn cargo_cmd(uses_xargo: bool) -> SafeCommand {
+pub(crate) fn cargo_safe_command(uses_xargo: bool) -> SafeCommand {
     if uses_xargo {
         SafeCommand::new("xargo")
     } else {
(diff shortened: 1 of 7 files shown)


src/docker/shared.rs
```diff
@@ -162,11 +162,11 @@ pub fn parse_docker_opts(value: &str) -> Result<Vec<String>> {
     shell_words::split(value).wrap_err_with(|| format!("could not parse docker opts of {}", value))
 }
 
-pub(crate) fn cargo_cmd(uses_xargo: bool) -> SafeCommand {
+pub(crate) fn cargo_safe_command(uses_xargo: bool) -> SafeCommand {
     if uses_xargo {
         SafeCommand::new("xargo")
     } else {
-        SafeCommand::new("cargo")
+        SafeCommand::new(env_program("CARGO", "cargo"))
```
Emilgardis, line 169: This is wrong, this struct should probably be renamed to `ContainedCommand`

It's used to assemble the command to use inside the container

we could introduce an environment variable `CROSS_CARGO_BIN` that sets this if wanted

Emilgardis (changes requested):

src/docker/shared.rs
Alexhuszagh: I think it's fine to skip that, since we need to mount a specific cargo right now.

Emilgardis:
r=me when commit message fixed

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
bors r=Emilgardis

PR #811, opened 2022-06-16
Bug fixes for listing/removing images.

Alexhuszagh:
Does nothing if there are no images to remove, and now correctly lists the old rustembedded images.

### DIFF
--- a/src/bin/commands/images.rs
+++ b/src/bin/commands/images.rs
@@ -4,8 +4,8 @@ use cross::CommandExt;
 // known image prefixes, with their registry
 // the docker.io registry can also be implicit
 const GHCR_IO: &str = cross::docker::CROSS_IMAGE;
-const RUST_EMBEDDED: &str = "rustembedded/cross:";
-const DOCKER_IO: &str = "docker.io/rustembedded/cross:";
+const RUST_EMBEDDED: &str = "rustembedded/cross";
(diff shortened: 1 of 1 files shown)


src/bin/commands/images.rs
```diff
@@ -185,7 +195,9 @@ fn remove_images(
         command.arg("--force");
     }
     command.args(images);
-    if execute {
+    if images.is_empty() {
+        Ok(())
+    } else if execute {
         command.run(verbose, false).map_err(Into::into)
     } else {
         println!("{:?}", command);
```
Emilgardis, line 203: this can use the new `command_pretty()` fn

Alexhuszagh:
bors r-
bors r=Emilgardis

src/bin/commands/images.rs
Emilgardis: i don't see this being done, but it's fine

Alexhuszagh: `Command::run` invokes `Command::run_and_get_status` which invokes `Command::print_verbose` which then called `Command::command_pretty`.

Alexhuszagh: Oh.... Oh.... Nevermind. I'll cancel this.

Alexhuszagh:
bors r=Emilgardis

PR #814, opened 2022-06-17
export new path instead of inlining it

Emilgardis:


Emilgardis:
closing this, this doesn't solve much sadly, will continue to investigate.

PR #818, opened 2022-06-18
change target info to json and display them neatly

Emilgardis:
this is not a serious pr, just an attempt that maybe can become something

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,51 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Alexhuszagh:
I think this is a good idea, since we can always generate the table on the Rust side from the JSON output, which is what this already does, and we can plug it into CI or something if needed in the future (maybe a good idea if we're planning a release, to ensure the README is correct).

Emilgardis:
> Do we need `rayon` though? It's a very heavy dependency (a good one, but might increase compile times by a lot).

Yeah, I placed it behind a feature to try mitigate that cost, but I thing it's impossible to not build it since we use --all-features

Used rayon because it took a long time to execute everything on my machine, I guess there is a way to do it without rayon but this was the simplest thing, can definitely remove it

PR #820, opened 2022-06-18
Add documentation for CLI for build-docker-image.

Alexhuszagh:


xtask/src/build_docker_image.rs
```diff
@@ -22,28 +23,35 @@ pub struct BuildDockerImage {
     /// Provide verbose diagnostic output.
     #[clap(short, long)]
     verbose: bool,
+    /// Print but do not execute the build commands.
     #[clap(long)]
     dry_run: bool,
+    /// Must be enabled to push is set and not on Github Actions.
```
Emilgardis, line 29:
```suggestion
    /// Force a push when `--push` is set, but not `--tag`
```

Emilgardis (changes requested):

xtask/src/build_docker_image.rs
Emilgardis: actually, should maybe hide this flag

xtask/src/build_docker_image.rs
```diff
@@ -22,28 +23,34 @@ pub struct BuildDockerImage {
     /// Provide verbose diagnostic output.
     #[clap(short, long)]
     verbose: bool,
+    /// Print but do not execute the build commands.
     #[clap(long)]
     dry_run: bool,
-    #[clap(long)]
+    #[clap(long, hide = true)]
```
Emilgardis, line 29: can keep the doc comment though, doesn't hurt

Emilgardis (approved):
lgtm!

bors r+

Issue #426, opened 2020-05-22
Can't target Raspberry Pi 1 and Pi Zero


BrianHoldsworth:
This issue is related to #391 and #414 

I am trying to build for the Pi Zero W using target=arm-unknown-linux-gnueabihf. I think that's the right target for RPi, but there is no way to distinguish between building for ARMv6 (RPi 1 and Zero) vs. ARMv7 (RPi 2). The resulting binary is ARMv7, which works on RPi2, but crashes with a segmentation fault on RPi Zero.

I add the following in my simple Rust project's `Cross.toml` file to avoid using the older "0.2" version docker image that was upgraded for #391 
```
[target.arm-unknown-linux-gnueabihf]
image = "rustembedded/cross:arm-unknown-linux-gnueabihf"
```
AND I can confirm using `docker images` after running `cross build` that I'm not using the older `arm-unknown-linux-gnueabihf-0.2.0`
```
REPOSITORY           TAG                           IMAGE ID            CREATED             SIZE
rustembedded/cross   arm-unknown-linux-gnueabihf   ba2dc7364b10        5 days ago          2.2GB
```
It appears to me that `cross build` is creating an ARMv7 binary, and this causes the Seqmentation Fault on the Pi Zero, which is an ARMv6 CPU.
```
$ !readelf
readelf --arch-specific target/arm-unknown-linux-gnueabihf/debug/rpi-test
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "7-A"
  Tag_CPU_arch: v7
  Tag_CPU_arch_profile: Application
  Tag_ARM_ISA_use: Yes
  Tag_THUMB_ISA_use: Thumb-2
  Tag_FP_arch: VFPv3-D16
  Tag_ABI_PCS_GOT_use: GOT-indirect
  Tag_ABI_PCS_wchar_t: 4
  Tag_ABI_FP_rounding: Needed
  Tag_ABI_FP_denormal: Needed
  Tag_ABI_FP_exceptions: Needed
  Tag_ABI_FP_number_model: IEEE 754
  Tag_ABI_align_needed: 8-byte
  Tag_ABI_enum_size: int
  Tag_ABI_VFP_args: VFP registers
  Tag_CPU_unaligned_access: v6
  Tag_ABI_FP_16bit_format: IEEE 754
```

schlamar:
[...]

PR #822, opened 2022-06-18
Read target architecture from cross-compiled binaries.

### LINKED ISSUE #821: Read Target Architecture Information in XTask
Read the target architecture, as if via `readelf -A` for a compiled binary from the target, without producing any build artifacts on the host, and print the output as JSON so that the target info can be verified. The `Tag_` prefixes should also be removed. This could be useful to prevent issues like #426 when adding new targets.

The output for `arm-unknown-linux-gnueabihf` would therefore look like:

```json
{
    "CPU_name": "6",
    "CPU_arch": "v6",
    "ARM_ISA_use": "Yes",
    "THUMB_ISA_use": "Thumb-1",
    "FP_arch": "VFPv2",
    "ABI_PCS_GOT_use": "GOT-indirect",
    "ABI_PCS_wchar_t": "4",
    "ABI_FP_rounding": "Needed",
    "ABI_FP_denormal": "Needed",
    "ABI_FP_exceptions": "Needed",
    "ABI_FP_number_model": "IEEE 754",
    "ABI_align_needed": "8-byte",
    "ABI_enum_size": "int",
    "ABI_VFP_args": "VFP registers",
    "CPU_unaligned_access": "v6",
    "ABI_FP_16bit_format": "IEEE 754"
}
```

Alexhuszagh:
Use `readelf -A` on generated binaries to read the target architecture information and output the data as JSON, simplifying verifying that new targets produce code for the desired architecture.

Closes #821.
Related to #426.


Alexhuszagh:
I also removed the `-it` flag from `target-info`, since this can cause a wait status to generate, and it isn't necessary for the command to function.

xtask/src/readelf.rs
```diff
@@ -0,0 +1,129 @@
...
+    command.args(&["bash", "-c", READELF_SCRIPT]);
+
+    if !verbose {
+        // capture stderr to avoid polluting table
+        command.stderr(Stdio::null());
+    }
+    let stdout = command.run_and_get_stdout(verbose)?;
+    let trimmed = stdout.trim();
+    if trimmed.is_empty() {
+        eprintln!("no target information: target is likely same architecture as host.");
```
Emilgardis, line 68:
```suggestion
    let out = command.run_and_get_output(verbose)?;
    let stdout = out.stdout()?;
    if stdout.trim().is_empty() {
        let stderr = out.stderr()?;
        if !stderr.trim().is_empty() {
            if stderr.contains("target may not be installed") {
                eprintln!("warning: target {target} is not installed");
                return Ok(());
            } else if !stderr.contains("Finished dev") {
                eprintln!("{}", stderr.header("stderr"));
                return Ok(());
            }
        }
        eprintln!("no target information: target `{target}` is likely same architecture as host.");
        if verbose {
            eprintln!("{}", stderr.header("stderr"));
        }
```

xtask/src/readelf.sh
```diff
@@ -0,0 +1,20 @@
...
+
+main() {
+    local td
+    td="$(mktemp -d)"
+
+    pushd "${td}" >/dev/null
+    cargo init --bin hello >/dev/null
+    cd hello
+    cargo build --target "${TARGET}" "${@}" >/dev/null
+    readelf -A "target/${TARGET}/debug/hello"
```
Emilgardis, line 15:
```suggestion
    cargo init --bin hello 1>&2
    cd hello
    cargo build --target "${TARGET}" "${@}" 1>&2
    readelf -A "target/${TARGET}/debug/hello"
```

xtask/src/readelf.rs
Emilgardis: uses `color_eyre::SectionExt` to output it nicely

xtask/src/util.rs
```diff
@@ -68,6 +72,30 @@ pub fn get_matrix() -> &'static Vec<Matrix> {
...
+pub fn pull_image(engine: &Path, image: &str, verbose: bool) -> cross::Result<()> {
+    let mut command = Command::new(engine);
+    command.arg("pull");
+    command.arg(image);
+    if !verbose {
+        // capture output to avoid polluting table
+        command.stdout(Stdio::null());
+        command.stderr(Stdio::null());
+    }
+    command.run(verbose, false).map_err(Into::into)
```
Emilgardis, line 96: I know this is old but I think this should be changed to

```suggestion
    let out = command.run_and_get_output(verbose)?;
    command.status_result(verbose, out.status, Some(&out))?;
    Ok(())
```

Emilgardis (changes requested):
looks good!

why not run `readelf` on the host?

Alexhuszagh:
This actually needs to install the target architecture via rustup if it's not present.

Emilgardis:
yup, I don't know if maybe it makes sense to do it only when wanted, eg `--download-targets` or something

Alexhuszagh:
Actually this might be more difficult than I thought. MIPS is giving me some very weird data. Probably will close this. You can submit the other changes as a separate PR.

PR #823, opened 2022-06-18
Minor bug fixes for target-info.

Alexhuszagh:
These manifest in the `increment_versions` branch, but are more general since they could theoretically happen with version upgrades on the current images.


Emilgardis (approved):
bors r+

PR #824, opened 2022-06-19
Minor xtask restructuring.

Alexhuszagh:
Remove the `-it` flag from `target-info`, which can the command to not complete. Also changes `pull_image` to use `run_and_get_output` rather than `run`, which is how we should capture the output. Finally, it moves `format_repo` and `pull_image` to `util`, so they can be used by other commands.

This applies a few of the changes mentioned in #822.


Emilgardis (approved):
bors r+

PR #826, opened 2022-06-19
cleanup

Emilgardis:


deny.toml
```diff
@@ -0,0 +1,39 @@
...
+multiple-versions = "deny"
+deny = []
+
+[sources]
+unknown-registry = "deny"
+unknown-git = "deny"
+allow-git = []
+
+[licenses]
+unlicensed = "deny"
```
Alexhuszagh, line 27: We should allow unlicensed code: it's public domain so it's compatible with any license.

deny.toml
```diff
@@ -0,0 +1,39 @@
...
+allow-osi-fsf-free = "neither"
+copyleft = "deny"
+confidence-threshold = 0.93
+allow = ["Apache-2.0", "MIT"]
+
+[[licenses.exceptions]]
+allow = ["CC0-1.0"]
```
Alexhuszagh, line 34: Same also with CC0 in general: we should probably allow that universally.

deny.toml
Emilgardis: I disagree. All crates should specify a license. if they don't specify, copyright and licensing is in a mess, not knowing how to apply.

Instead they should use something like `WTFPL` or `UNLICENSE`

Alexhuszagh: Oh, I assumed that was unlicense. If it's lacking a license, then yes, we should deny it.

deny.toml
Emilgardis: Hmm, without this it warns with

```
warning[L006]: license was not encountered
   ┌─ /home/emil/workspace/cross/deny.toml:31:31
   │
31 │ allow = ["Apache-2.0", "MIT", "CC0-1.0"]
   │                               ^^^^^^^^^ unmatched license allowance

advisories ok, bans ok, licenses ok, sources ok
```

it doesn't make it fail, so I'll change it eitherway

Emilgardis: This is because it checks each target

Alexhuszagh: > This is because it checks each target

Ah ok, maybe then we need something different. I might have to play around with it a bit.

src/docker/custom.rs
```diff
@@ -114,12 +114,14 @@ impl<'a> Dockerfile<'a> {
...
-                        .get(..5)
-                        .expect("sha1 is expected to be at least 5 characters long"),
+                path_hash = format!(
+                    "{}",
+                    const_sha1::sha1(&const_sha1::ConstBuffer::from_slice(
+                        metadata.workspace_root.to_string_lossy().as_bytes()
+                    ))
+                )
+                .get(..5)
+                .expect("md5 is expected to be at least 5 characters long"),
```
Alexhuszagh, line 124: This isn't MD5 anymore: it's still SHA1.

LICENSE-MIT
```diff
@@ -1,3 +1,4 @@
+Copyright (c) 2017-2022 by the respective authors
```
Emilgardis, line 1: added this to try to be as inclusive as possible, maybe there's a better way but I don't know any.

Alexhuszagh: I think that works.

Alexhuszagh (approved):

PR #827, opened 2022-06-19
Improve release infrastructure

Emilgardis:


Cargo.toml
```diff
@@ -8,6 +8,17 @@ name = "cross"
 repository = "https://github.com/cross-rs/cross"
 version = "0.2.1"
 edition = "2021"
+include = [
+  "src/**/*",
+  "docker/Dockerfile.*",
+  "docker/*.sh",
```
Emilgardis, line 14: this isn't really needed, only the Dockerfiles need to be included, we should fix that later, as currently Dockerfile.x86_64-unknown-linux-gnu.centos is included as a builtin as well.

Emilgardis (commented):
I've tested the publish action fairly well I think.

https://github.com/Emilgardis/emilgardis-test-crate/releases

Emilgardis:
I think one nice thing we could do is detect if the merge edits the version number, and then on successful ci push a tag for that release

We could move the actual cargo publish to a job in an environment, and require approval from maintainers before executionz this job would be triggered on if its a push.

Either way, we should fail the whole workflow early if tag != version specified in Cargo.toml

Cargo.toml
Alexhuszagh: Do we even need the Dockerfiles? People aren't building docker images from the registry, and if they were, the scripts would be needed. We don't reference them in the documentation currently either.

Alexhuszagh:
> I think one nice thing we could do is detect if the merge edits the version number, and then on successful ci push a tag for that release
> 
> We could move the actual cargo publish to a job in an environment, and require approval from maintainers before executionz this job would be triggered on if its a push.
> 
> Either way, we should fail the whole workflow early if tag != version specified in Cargo.toml

I think we can do that in a separate PR, and I can look pretty carefully and test your work, but you know Github Actions better than I do.

PR #828, opened 2022-06-21
Avoid using display on paths.

Alexhuszagh:
Adds a helper trait `ToUtf8` which converts the path to UTF-8 if possible, and if not, creates a pretty debugging representation of the path. We can then change `display()` to `to_utf8()?` and be completely correct in all cases.

On POSIX systems, this doesn't matter since paths must be UTF-8 anyway, but on Windows some paths can be WTF-8 (see rust-lang/rust#12056). Either way, this can avoid creating a copy in cases, is always correct, and is more idiomatic about what we're doing. We might not be able to handle non-UTF-8 paths currently (like ISO-8859-1 paths, which are historically very common).

So, this doesn't change ergonomics: the resulting code is as compact and correct. It also requires less copies in most cases, which should be a good thing. But most importantly, if we're mounting data we can't silently fail or produce incorrect data. If something was lossily generated, we probably shouldn't try to mount with a replacement character, and also print more information than there was just an invalid Unicode character.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Internal
 
+- #828 - assume paths are Unicode and provide better error messages for path encoding errors.
 - #787 - add installer for git hooks.
 - #786, #791 - Migrate build script to rust: `cargo build-docker-image $TARGET`
 - #730 - make FreeBSD builds more resilient.
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,3 @@
+disallowed-methods = [
+    { path = "std::path::Path::display", reason = "incorrect handling of non-Unicode paths, use path.to_utf8() or debug (`{path:?}`) instead" },
+]
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 
 use crate::docker::Engine;
 use crate::{config::Config, docker, CargoMetadata, Target};
-use crate::{errors::*, file, CommandExt};
+use crate::{errors::*, file, CommandExt, ToUtf8};
 
 use super::{image_name, parse_docker_opts};
 
@@ -47,7 +47,7 @@ impl<'a> Dockerfile<'a> {
             &format!(
                 "{}.workspace_root={}",
                 crate::CROSS_LABEL_DOMAIN,
-                metadata.workspace_root.display()
+                metadata.workspace_root.to_utf8()?
             ),
         ]);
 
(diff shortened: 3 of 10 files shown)


Alexhuszagh:
I'll be adding the other `as_posix` helper in this PR, but a quick explanation on a Linux system:

Say I create a file with an `ISO-8859-1` filename, ensuring this is done in Python:

```python
>>> with open("Überraschung".encode("ISO-8859-15"), "w") as f:
...     f.write("Hello")
... 
5
>>> import os; os.listdir(os.getcwd())
['\udcdcberraschung']
```

We can confirm this is correct in Bash by checking the output in UTF-8 (only works if the terminal emulator also expects UTF-8 encoded output, which it **should**):

```bash
$ LANG=en_us.UTF-8 ls
''$'\334''berraschung
```

So now I generate my Rust code:

```rust
use std::{fs, io, env};

pub fn main() -> io::Result<()> {
    let cwd = env::current_dir()?;
    for entry in fs::read_dir(cwd)? {
        let path = entry?.path();
        println!("debug={path:?}");
        println!("display={}", path.display());
    }

    Ok(())
}
```

And this outputs:

```
./main 
debug="/home/ahuszagh/Desktop/tmp/\xDCberraschung"
display=/home/ahuszagh/Desktop/tmp/�berraschung
```

Or, our output isn't exactly correct: we shouldn't be using `display` either for error messages or mounting.

Emilgardis:
does it make sense to add `Path::display` to clippy lint [`disallowed_methods`](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods)

Alexhuszagh:
> does it make sense to add `Path::display` to clippy lint [`disallowed_methods`](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods)

Yes I think so. It is incorrect for our uses, even if it's a good method for more general cases.

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
Just an additional comment in case this ever needs to be discussed again, as unlikely as that is: we need both the host and the container locale to be the same, I believe, since we need to provide bind mount mappings from the host to the container that are identical. Code like this (I believe) cannot exist if the path is not valid Unicode, since our container expects this to be a path that can be represented in UTF-8:

https://github.com/cross-rs/cross/blob/29370328523fbf86c57e6436598cfdaf04cd9eff/src/docker/shared.rs#L173-L186

However, nothing here stops you from providing non-UTF-8 filenames or directories as part of your crate data or external mounts to `cross`, just they cannot be the mount points. For example, if I have `/path/to/dir/{{ISO-8859-15}}`, I can provide a mount to `/path/to/dir`, and this should work fine, allowing me to access my files with non-UTF-8 filenames without issue (common if you need to process data extracted from a legacy source onto a modern system).

Alexhuszagh:
bors r-

This isn't a bug, but it does lead to `//` being duplicated for the path immediately after root. I'll add tests too for this.

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #830, opened 2022-06-21
add `strip = true` to release profile

Emilgardis:
also sets cross as default run binary

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
 
       - name: Install cross
         if: matrix.deploy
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -220,7 +220,7 @@ jobs:
 
       - name: Install cross
         if: matrix.deploy
-        run: cargo install --path . --force
+        run: cargo install --path . --force --debug
```
Emilgardis, line 223: fixed this also, don't think we need it in release mode do we?

Emilgardis:
oh, strip is 1.59...

PR #831, opened 2022-06-21
Fixes regression in #828 with absolute path.

Alexhuszagh:
Prior to #828, we added a '/' separator before component in our makeshift `as_posix`, so the path would always become absolute, even though the actual path was relative. `as_posix` therefore handled this as a relative path, so using `project` and not `/project` as the working directory for the cross command caused commands to fail.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -242,7 +242,7 @@ pub(crate) fn docker_cwd(
     } else {
         // We do this to avoid clashes with path separators. Windows uses `\` as a path separator on Path::join
         let cwd = &cwd;
-        let working_dir = Path::new("project").join(cwd.strip_prefix(&metadata.workspace_root)?);
+        let working_dir = Path::new("/project").join(cwd.strip_prefix(&metadata.workspace_root)?);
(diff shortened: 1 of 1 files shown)


Emilgardis (approved):
bors r+

those windows and mac os tests would've been great to have to catch simple stuff like this.

thanks!

Alexhuszagh:
Specifically, this was the code that converted everything effectively to an absolute path, which I failed to consider in my PR:

https://github.com/cross-rs/cross/pull/828/files#diff-b0f71fc97387d26459b3d22f08eaae7b431074839cce7cb6aacf3cd154db9800L246-L252

```rust
let cwd = &cwd;
let working_dir = Path::new("project").join(cwd.strip_prefix(&metadata.workspace_root)?);
// No [T].join for OsStr
docker.args(&["-w", &working_dir.as_posix()?]);
let mut mount_wd = std::ffi::OsString::new();
for part in working_dir.iter() {
    mount_wd.push("/");
    mount_wd.push(part);
}
```

So this fixes that regression (whoops).

PR #834, opened 2022-06-22
only check msrv in clippy checks

Emilgardis:


Emilgardis:
bors r-

actually I missed something...

Emilgardis:
bors r=Alexhuszagh

Emilgardis:
timeout, bumped from 5 to 10 minutes.

It's only long now because we have no cache

Emilgardis:
bors r=Alexhuszagh

PR #836, opened 2022-06-22
Add a cache dir tag when creating a target directory.

### LINKED ISSUE #835: Cross Doesn't Write a `CACHEDIR.TAG`
`cargo` does not write `CACHEDIR.TAG` in the target directory if it already exists, and since we create one if the directory does not exist for permission issues, we need to write a `CACHEDIR.TAG` file if it does not exist. This particularly affects #785, since it means we copy over the entire target directory over.

The specification can be found [here](https://bford.info/cachedir/).

Alexhuszagh:
Ensure `cross` is similar in functionality to `cargo`, and this minimizes network traffic by applications that may copy the project but do not wish to copy compiled code.

Closes #835.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


Emilgardis (approved):
bors r+

PR #837, opened 2022-06-22
check tag with package version early and determine if it's the latest version

Emilgardis:


xtask/src/build_docker_image.rs
```diff
@@ -289,7 +289,7 @@ pub fn determine_image_name(
             }
             tags.push(target.image_name(repository, version));
             // Check for unstable releases, tag stable releases as `latest`
-            if version.contains('-') {
+            if !version.contains('-') {
                 // TODO: Don't tag if version is older than currently released version.
                 tags.push(target.image_name(repository, "latest"))
             }
```
Emilgardis, line 295: don't know how to best solve the issue of this taging backports as latest.

Should we support backports? Do we need to?

Alexhuszagh: We might need to if we need to say, fix the `0.2` branch and we have released `0.3`, I believe.

Emilgardis: this is implemented now

xtask/src/build_docker_image.rs
```diff
@@ -289,8 +294,7 @@ pub fn determine_image_name(
...
             tags.push(target.image_name(repository, version));
             // Check for unstable releases, tag stable releases as `latest`
-            if version.contains('-') {
-                // TODO: Don't tag if version is older than currently released version.
+            if is_latest {
```
Emilgardis, line 297: this removes a bug, prereleases would be tagged latest, stable wouldn't

xtask/src/build_docker_image.rs
```diff
@@ -11,6 +11,8 @@ pub struct BuildDockerImage {
     ref_type: Option<String>,
     #[clap(long, hide = true, env = "GITHUB_REF_NAME")]
     ref_name: Option<String>,
+    #[clap(long, hide = true, env = "IS_LATEST")]
+    is_latest: Option<bool>,
```
Alexhuszagh, line 15: Maybe a better idea would be to make this a `bool` and have it be called `latest`, and access the `LATEST` envvar? `--latest` or `--set-latest` would describe an action seems better.

Emilgardis: fixed, kept the name `is_latest` but flag is --latest

Alexhuszagh: FYI: you've changed the envvar name but I still think `is_latest` for the CLI should be renamed to `latest`. Going to test it out shortly but looks good at first glance.

Emilgardis: it is renamed, `long = "latest"`

Alexhuszagh (approved):

Emilgardis:
bors r=Alexhuszagh

PR #841, opened 2022-06-23
Update external toolchains to support MSVC and Apple targets.

Alexhuszagh:
Adds support for building custom MSVC and Apple Darwin targets, and references these images in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,7 @@ terminate.
 due to a [compiler bug](https://github.com/rust-lang/rust/issues/85821), you will have to build them yourself for now.-->
 
 Additional Dockerfiles for other targets can be found in [cross-toolchains](https://github.com/cross-rs/cross-toolchains).
+These include MSVC targets, which for legal reasons we cannot ship pre-built images of.
 
 ## Debugging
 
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
Since this doesn't change anything in the repository, I didn't update the changelog. I've also confirmed all the images compile, that the x86 targets have valid runners, that the ARM64 image compiles correctly, and that the ARM image (which doesn't support `std` currently) compiles a `no_std` library.

Emilgardis (approved):

Alexhuszagh:
I'm going to hold off on this for a short period until I get macOS working, improve lowercase handling slightly, and add to the documentation that by default they should update the `cross-toolchains` repository to the latest version by default, so we only need to increment the submodule when very important targets are added.

README.md
```diff
@@ -393,6 +393,7 @@ terminate.
 due to a [compiler bug](https://github.com/rust-lang/rust/issues/85821), you will have to build them yourself for now.-->
 
 Additional Dockerfiles for other targets can be found in [cross-toolchains](https://github.com/cross-rs/cross-toolchains).
+These include MSVC and Apple Darwin targets, which for legal reasons we cannot ship pre-built images of.
```
Emilgardis, line 396:
```suggestion
These include MSVC and Apple Darwin targets, which we cannot ship pre-built images of.
```

Alexhuszagh:
bors r=Emilgardis

PR #843, opened 2022-06-23
list/delete custom cross images as well

Emilgardis:


src/bin/commands/images.rs
```diff
@@ -125,19 +143,27 @@ fn get_cross_images(
...
         .lines()
         .map(parse_image)
         .filter(|image| is_cross_image(&image.repository))
         .filter(|image| local || !is_local_image(&image.tag))
         .collect();
-    images.sort();
 
-    Ok(images)
+    let stdout = cross::docker::subcommand(engine, "images")
+        .args(&["--format", r#"{{.Repository}}:{{.Tag}} {{.ID}}"#])
```
Alexhuszagh, line 157: I don't think this needs an escape sequence, does it?

Emilgardis: oh yeah, that's a remenant of me trying to hack the go formatter to get the labels

sadly its not feasible, you can get them, but I don't know how well supported that is

`{{json .}}` i think was how I got them

src/bin/commands/images.rs
```diff
@@ -178,14 +236,14 @@ pub fn list_images(
 ) -> cross::Result<()> {
     get_cross_images(engine, verbose, true)?
         .iter()
-        .for_each(|line| println!("{}", line.name()));
+        .for_each(|image| println!("{}", image));
```
Emilgardis, line 239: this fixes an issue with images with only an id

src/bin/commands/images.rs
```diff
@@ -160,16 +185,44 @@ fn rustembedded_target(tag: &str) -> String {
...
+            .cloned()
+        {
+            return Ok(target);
+        }
+    }
+    let mut command = cross::docker::subcommand(engine, "inspect");
+    command.args(&[
+        "--format",
+        &format!(
+            r#"{{index .Config.Labels "{}.for-cross-target"}}"#,
```
Alexhuszagh, line 216: Does this work if more than one custom image with that label exists, and falls through to here?

Emilgardis: oh you're right, an argument is missing

```
.arg(image.id)
```

Emilgardis: I had multiple branches with this change, I picked one that was not fully complete :D, this is now resolved

src/bin/commands/images.rs
```diff
@@ -160,16 +188,46 @@ fn rustembedded_target(tag: &str) -> String {
...
 
-fn get_image_target(image: &Image) -> cross::Result<String> {
-    if let Some(stripped) = image.repository.strip_prefix(GHCR_IO) {
-        Ok(stripped.to_string())
+fn get_image_target(
+    engine: &cross::docker::Engine,
+    image: &Image,
+    target_list: &TargetList,
+) -> cross::Result<String> {
+    if let Some(stripped) = image.repository.strip_prefix(&format!("{GHCR_IO}/")) {
```
Emilgardis, line 196: without this, targets become `/<target>`, so that they dont match

src/bin/commands/images.rs
```diff
@@ -124,20 +141,31 @@ fn get_cross_images(
...
+        .run_and_get_stdout(verbose)?;
+    let ids: Vec<_> = images.iter().map(|i| i.id.to_string()).collect();
+    images.extend(
+        stdout
+            .lines()
+            .map(parse_image)
+            .filter(|i| !ids.iter().any(|id| id == &i.id))
+            .filter(|image| is_cross_image(&image.repository))
+            .filter(|image| local || !is_local_image(&image.tag)),
+    );
```
Emilgardis, line 166: skips parsing image meta if they've already been discovered

Alexhuszagh (approved):

PR #849, opened 2022-06-24
implement baking

Emilgardis:


Emilgardis:
with this comes a question, should we tag all these even if they will eventually come up in a manifest?

Reason for including this is to enable easier cleanup, as right now the images that are in the manifest will be untagged.

I think it's fine to tag them, I'm just not sure how to tag them. Maybe `<tag>-<arch>` ?

Emilgardis:
Note also, there's a bug on ghcr where they suggest adding a LABEL to the manifest, this is impossible to do: 
image

both images in the manifest do have descriptions

image

Emilgardis:
The proposed workflow for a multi platform image would be 

1. build the image(s) as normal, separately, specify platform if needed. make sure they are on the same builder as bake would use
2. test
3. on need to push, use --bake and specify

xtask/src/build_docker_image.rs
```diff
@@ -165,148 +202,375 @@ pub fn build_docker_image(
...
-
-        if gha || progress == "plain" {
-            docker_build.args(&["--progress", "plain"]);
+        if push {
+            if let Some("") = tag_override.as_deref() {
+                docker_bake.args(&["--set", "*.output=type=registry,push-by-digest=true"]);
+            } else {
+                docker_bake.args(&["--set", "*.output=type=registry"]);
+            }
+            //docker_bake.args(&["--set", "*.cache-to=type=inline"]);
```
Emilgardis, line 469: this should not be commented

Emilgardis:
marking as draft again, this is ready for use though, there is something I'm not liking about the naming though.

The issue is that for proper/good caching, we need unique tags for each arhitecture, otherwise the cache will use the most "upper" image in the manifest.

I think

`<target>:<version>-<arch>` makes sense, but it is a bit confusing.

`x86_64-unknown-linux-gnu:main-aarch64` for example

Emilgardis:
closing this, the code has diverged to much and I want to reimplement this.

The idea would be to append `-{arch}`, and then assemble into a manifest, I don't think using bake is needed. Instead we use a normal build, publish for the different archs, and create a manifest with the images of all the architectures

PR #851, opened 2022-06-24
run simple tests on windows and linux

Emilgardis:
solves parts of #610, depends on #852

PR #852, opened 2022-06-24
use cygpath if wsl is not available

Emilgardis:
Implements getting unix paths with `cygpath` if wsl/wslpath is not installed.

```
Error:
   0: could not get linux compatible path for `"G:\\workspace\\small dev space\\test-workspace\\workspace"`

Error:
   0: could not find cygpath.exe

Error:
   0: could not find wslpath.exe

Warning: usage of `env.volumes` requires WSL or cygpath on Windows
Suggestion: is WSL or cygpath installed on the host?
```

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -85,8 +85,8 @@ impl Directories {
         let mount_root: PathBuf;
         #[cfg(target_os = "windows")]
         {
-            // On Windows, we can not mount the directory name directly. Instead, we use wslpath to convert the path to a linux compatible path.
-            mount_root = wslpath(&host_root, verbose)?;
+            // On Windows, we can not mount the directory name directly. Instead, we use winpath to convert the path to a linux compatible path.
+            mount_root = winpath(&host_root, verbose)?;
         }
         #[cfg(not(target_os = "windows"))]
         {
@@ -95,8 +95,8 @@ impl Directories {
         let mount_cwd: PathBuf;
         #[cfg(target_os = "windows")]
         {
-            // On Windows, we can not mount the directory name directly. Instead, we use wslpath to convert the path to a linux compatible path.
-            mount_cwd = wslpath(cwd, verbose)?;
+            // On Windows, we can not mount the directory name directly. Instead, we use winpath to convert the path to a linux compatible path.
+            mount_cwd = winpath(cwd, verbose)?;
         }
         #[cfg(not(target_os = "windows"))]
         {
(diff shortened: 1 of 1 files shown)


src/docker/shared.rs
```diff
@@ -339,29 +339,61 @@ pub(crate) fn docker_mount(
...
+                .with_error(|| {
+                    <eyre::Report as std::convert::AsRef<
+                        dyn std::error::Error + Send + Sync + 'static,
+                    >>::as_ref(Box::leak(Box::new(cyg_err)))
+                })
+                .with_error(|| {
+                    <eyre::Report as std::convert::AsRef<
+                        dyn std::error::Error + Send + Sync + 'static,
+                    >>::as_ref(Box::leak(Box::new(wsl_err)))
+                })
```
Emilgardis, line 359: some nice leaks, unfortunately couldn't find a better way to do this

src/docker/shared.rs
```diff
@@ -410,7 +442,7 @@ pub(crate) fn docker_seccomp(
             #[cfg(target_os = "windows")]
             if matches!(engine_type, EngineType::Podman | EngineType::PodmanRemote) {
                 // podman weirdly expects a WSL path here, and fails otherwise
-                path = wslpath(&path, verbose)?;
+                path = winpath(&path, verbose)?;
```
Emilgardis, line 445: does this work with podman?

Alexhuszagh:
Isn't WSL2 required though? Both Docker and Podman require it on Windows.

Emilgardis:
wsl2 is not required, only wsl. doing this specifically because of the error in #851 : https://github.com/cross-rs/cross/runs/7045847730?check_suite_focus=true

Alexhuszagh:
> wsl2 is not required, only wsl. doing this specifically because of the error in #851 : https://github.com/cross-rs/cross/runs/7045847730?check_suite_focus=true

I believe WSL2 is now required, or Hyper-V (I forgot that [LinuxKit](https://github.com/linuxkit/linuxkit) was a thing). I know WSL used to be used back in 2017ish? 2016ish? back when I used it then before Docker Desktop was released, but I doubt we should rely on anyone used that anymore. We probably need a different solution than cygpath if we do, since we would want it to be compatible with Hyper-V?

Emilgardis:
Putting the discussion from matrix here:

> We probably need a different solution than cygpath if we do, since we would want it to be compatible with Hyper-V?

We don't need the paths to be compatible, only that they are expressed in a sane way that unix understands.

This is because we mount the volume paths with `-v C:\path\to\thing:/mnt/c/path/to/thing`

We could instead of using cygpath fallback to something like [`relative_path`](https://crates.io/crates/relative-path)

Alexhuszagh:
I think this won't be too complex to implement in cross itself. We shouldn't have to worry about UNC paths or I believe verbatim paths as long as we canonicalize them first.

Emilgardis:
docker -v doesn't take unc paths, don't think there are cases where we use unc paths

Alexhuszagh:
> docker -v doesn't take unc paths, don't think there are cases where we use unc paths

If that's the case, then handling drives is pretty easy.

PR #856, opened 2022-06-24
Remove WSLpath and implement WSL-style path.

### LINKED ISSUE #854: Remove WSLPath
### Checklist

### Describe your issue

Currently, we use `wsl -e wslpath -a ...` to find the WSL2 path of a file for our mount point on Windows. This poses a few issues:
- It requires external commands, which add significant overhead.
- It does not work for UNC paths.

A simple test is as follows:

```bash
wsl -e wslpath -a "\\server\share"
wslpath: \\server\share
```

_No response_

- [ ] macOS
- [x] Windows
- [ ] Linux / BSD

- [x] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [ ] podman

### cross version

latest main build

### Example

_No response_

_No response_

Alexhuszagh:
Removes the dependency on WSL2, and properly handles DOS-style paths, UNC paths, including those on localhost, even if Docker itself does not support them.

Closes #854.
Related to #665.
Supersedes #852.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -35,8 +35,8 @@ pub struct Directories {
     pub nix_store: Option<PathBuf>,
     pub host_root: PathBuf,
     // both mount fields are WSL paths on windows: they already are POSIX paths
-    pub mount_root: PathBuf,
-    pub mount_cwd: PathBuf,
+    pub mount_root: String,
+    pub mount_cwd: String,
     pub sysroot: PathBuf,
 }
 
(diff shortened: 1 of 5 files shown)


src/docker/shared.rs
```diff
@@ -323,49 +320,31 @@ pub(crate) fn docker_mount(
...
-        .arg("-a")
-        .arg(path)
-        .run_and_get_stdout(verbose)
-        .wrap_err_with(|| format!("could not get linux compatible path for `{path:?}`"))
-        .map(|s| s.trim().into())
-}
-
-#[allow(unused_variables)]
-pub(crate) fn canonicalize_mount_path(path: &Path, verbose: bool) -> Result<PathBuf> {
+pub(crate) fn canonicalize_mount_path(path: &Path) -> Result<String> {
```
Emilgardis, line 339: it'd be nice to encode this as a string struct, to ensure we handle it properly.

That can be done later.

Alexhuszagh:
Forgot `C:` won't necessarily map to the current directory if you're not on the same drive, and we have no way of telling which drive we're on in CI.

src/file.rs
```diff
@@ -166,6 +245,12 @@ mod tests {
     use super::*;
     use std::fmt::Debug;
 
+    macro_rules! p {
+        ($path:expr) => {
+            Path::new($path)
+        };
+    }
```
Emilgardis, line 252: haha nice :)

Emilgardis (approved):
Nice!

bors r+

PR #859, opened 2022-06-25
Add cargo-style output diagnostics.

### LINKED ISSUE #797: Output diagnostics in cargo-style
### Checklist

### Describe your request

cross should output it's warnings, notes and errors in the same way cargo does it

e.g something like this https://github.com/rust-lang/cargo/blob/ca4edabb28fc96fdf2a1d56fe3851831ac166f8a/src/cargo/core/shell.rs#L251-L257

This will make integration with cargo look better

Alexhuszagh:
Adds the quiet and color command-line flags, where color supports `auto`, always`, and `never`. These command-line flags are parsed to a verbosity which can be quiet, normal, or verbose.

With these, we then have the stderr message formatters:
- `fatal_usage`: print a fatal error message with the failing argument, and add a help context menu for how to use cross.
- `fatal`: print red 'error' message and exit with an error code
- `error`: print red 'error' message
- `warn`: print amber 'warning' message
- `note`: print cyan 'note' message
- `status`: print an uncolored and unprefixed 'status' message

We have the stdout message formatters:
- `print`: always print the message
- `info`: print the message as long as the verbosity is not quiet
- `debug`: only print the message if the output is not quiet

We also have a few specialized error handlers, and methods to help ensure we can have flexible error reporting in the future:
- `status_stderr`
- `status_stdout`

The command extensions now have, `print`, `info`, and `debug`, which formats the command and sends it to the shell. This allows us to avoid using `print_verbose` where we sometimes manually override the default setting.

A few of these settings aren't currently used (such as `info` and `status`, but they're a very common pattern), so we can ensure we have the necessary utilities to ensure we can properly format messages in the future. A few like `shell::print` are practically identical to `println!`, but this allows us to customize it if need be in the future.

Solves parts of #797.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - #869 - ensure cargo configuration environment variable flags are passed to the docker container.
+- #859 - added color diagnostic output and error messages.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ ctrlc = { version = "3.2.2", features = ["termination"] }
 directories = "4.0.1"
 walkdir = { version = "2", optional = true }
 tempfile = "3.3.0"
+owo-colors = { version = "3.4.0", features = ["supports-colors"] }
 
 [target.'cfg(not(windows))'.dependencies]
 nix = { version = "0.24", default-features = false, features = ["user"] }
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -1,68 +1,74 @@
+use std::fmt::Write;
 use std::path::Path;
 
-use crate::util::cargo_metadata;
+use crate::util::{cargo_metadata, gha_error, gha_output, gha_print};
 use clap::Args;
-use color_eyre::Section;
+use cross::shell::{self, MessageInfo};
 use cross::{docker, CommandExt, ToUtf8};
-use std::fmt::Write;
 
 #[derive(Args, Debug)]
 pub struct BuildDockerImage {
     #[clap(long, hide = true, env = "GITHUB_REF_TYPE")]
-    ref_type: Option<String>,
+    pub ref_type: Option<String>,
     #[clap(long, hide = true, env = "GITHUB_REF_NAME")]
     ref_name: Option<String>,
     #[clap(action, long = "latest", hide = true, env = "LATEST")]
     is_latest: bool,
     /// Specify a tag to use instead of the derived one, eg `local`
     #[clap(long)]
-    tag: Option<String>,
+    pub tag: Option<String>,
     /// Repository name for image.
     #[clap(long, default_value = docker::CROSS_IMAGE)]
-    repository: String,
+    pub repository: String,
     /// Newline separated labels
     #[clap(long, env = "LABELS")]
-    labels: Option<String>,
+    pub labels: Option<String>,
     /// Provide verbose diagnostic output.
     #[clap(short, long)]
     pub verbose: bool,
+    /// Do not print cross log messages.
+    #[clap(short, long)]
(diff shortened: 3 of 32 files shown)


Alexhuszagh:
We only manually intercept `--color` if it's invalid, or `--verbose` and `--quiet` if both are set, and let everything else be handled by cargo. Some sample differences (where `target/debug/cross` is my local branch, and `cross` is the v0.2.2 release) are:

Color
Fallback
Quiet Verbose
Warning

There's quite a dramatic difference.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ ctrlc = { version = "3.2.2", features = ["termination"] }
 directories = "4.0.1"
 walkdir = { version = "2", optional = true }
 tempfile = "3.3.0"
+termcolor = "1.1.3"
```
Emilgardis, line 50: I'd prefer if we kept using owo-colors, as we now have two color crates.

owo-colors should work nicely for this. It also has automatic `COLOR` support https://github.com/jam1garner/owo-colors#only-style-on-supported-terminals

Alexhuszagh: The only downside to `owo-colors` is I believe it does not support the `always` configuration like `cargo`, and `termcolor` also supports intelligent color support (only print on supported terminals).

Alexhuszagh: Ok updated to use `owo-colors`. It's a bit of a mess, but the best I think we can do while still supporting how Cargo does it (allowing you to specify `auto`, `never`, and `always`). So we still have `ColorChoice`:

```rust
/// Whether messages should use color output
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ColorChoice {
    /// force color output
    Always,
    /// force disable color output
    Never,
    /// intelligently guess whether to use color output
    Auto,
}
```

And then writing the style is as follows:

```rust
// generate the color style
macro_rules! write_style {
    ($stream:ident, $msg_info:expr, $message:expr $(, $style:ident)* $(,)?) => {{
        match $msg_info.color_choice {
            ColorChoice::Always => write!($stream, "{}", $message $(.$style())*),
            ColorChoice::Never => write!($stream, "{}", $message),
            ColorChoice::Auto => write!(
                $stream,
                "{}",
                $message $(.if_supports_color($stream.owo(), |text| text.$style()))*
            ),
        }?;
    }};
}
```

In order to avoid repeating ourselves extensively, we somewhat need wildcard expansion.

Alexhuszagh: Update: we might need to switch back to `termcolor`, or use the old logic of `is_atty`. We can't use `if_supports_color` due to the ISC license of `is_ci`, which we need for the `supports-color` feature.

Emilgardis (approved):
Awesome, looks great

I don't think this solves #797 however, but it's a good way to it

bors r+

xtask/src/build_docker_image.rs
```diff
@@ -250,30 +256,27 @@ pub fn build_docker_image(
...
-            println!("::set-output name=image::{}", &tags[0]);
-            println!(
-                "::set-output name=images::'{}'",
-                serde_json::to_string(&tags)?
-            );
+            gha_output("image", &tags[0]);
+            gha_output("images", &serde_json::to_string(&tags)?);
             if targets.len() > 1 {
-                println!("::endgroup::");
+                gha_print("::endgroup::");
```
Emilgardis, line 268: This changeset is wrong,

Note the `'` around images

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- #859 - added color diagnostic output and error messages.
```
Emilgardis, line 47: oops! Missed this too, wrong section for changelog, should be moved to unreleased

Emilgardis (changes requested):

Emilgardis (approved):
bors r+

PR #867, opened 2022-06-26
Fix parsing of config `build.env.passthrough` values.

Alexhuszagh:
Ensure `build.env.passthrough` values are correctly parsed, and do not return `None`, and ensure that they merge as expected with `target.(...).env.passthrough` values.


Emilgardis (approved):
bors r+

PR #873, opened 2022-06-27
Add integration tests.

Alexhuszagh:
Adds an an integration test for remote Docker support with and without persistent data volumes, and adds it to CI. In addition, we've added integration tests for custom toolchains using `cargo-bisect-rustc`.  Also changes the workflow  to ensure publish depends on fmt, clippy,  and cargo-deny. Fixes a bug when removing data volumes, since it does  not require a target.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,8 +321,69 @@ jobs:
           LATEST: ${{ needs.check.outputs.is-latest || 'false' }}
         shell: bash
 
+  # we should always have an artifact from a previous build.
+  remote:
+    needs: [shellcheck, test, check]
+    runs-on: ubuntu-latest
+    if: github.actor == 'bors[bot]'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-rust
+
+      - name: LLVM instrument coverage
+        id: remote-cov
+        uses: ./.github/actions/cargo-llvm-cov
+        with:
+          name: integration-remote
+
+      - name: Run Remote Test
+        env:
+          TARGET: aarch64-unknown-linux-gnu
+        run: ./ci/test-remote.sh
+        shell: bash
+
+  bisect:
+    needs: [shellcheck, test, check]
+    runs-on: ubuntu-latest
+    if: github.actor == 'bors[bot]'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-rust
+
+      - name: LLVM instrument coverage
+        id: bisect-cov
+        uses: ./.github/actions/cargo-llvm-cov
+        with:
+          name: integration-bisect
+
+      - name: Run Bisect Test
+        env:
+          TARGET: aarch64-unknown-linux-gnu
+        run: ./ci/test-bisect.sh
+        shell: bash
+
+  docker-in-docker:
+    needs: [shellcheck, test, check]
+    runs-on: ubuntu-latest
+    if: github.actor == 'bors[bot]'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-rust
+
+      - name: LLVM instrument coverage
+        id: docker-in-docker-cov
+        uses: ./.github/actions/cargo-llvm-cov
+        with:
+          name: integration-docker-in-docker
+
+      - name: Run Docker-in-Docker Test
+        env:
+          TARGET: aarch64-unknown-linux-gnu
+        run: ./ci/test-docker-in-docker.sh
+        shell: bash
+
   publish:
-    needs: [build, check]
+    needs: [build, check, fmt, clippy, cargo-deny, remote, bisect, docker-in-docker]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -763,6 +763,162 @@ mod tests {
         }
     }
 
+    mod directories {
+        use super::*;
+        use crate::cargo::cargo_metadata_with_args;
+        use crate::temp;
+
+        fn unset_env() -> Vec<(&'static str, Option<String>)> {
+            let mut result = vec![];
+            let envvars = ["CARGO_HOME", "XARGO_HOME", "NIX_STORE"];
+            for var in envvars {
+                result.push((var, env::var(var).ok()));
+                env::remove_var(var);
+            }
+
+            result
+        }
+
+        fn reset_env(vars: Vec<(&'static str, Option<String>)>) {
+            for (var, value) in vars {
+                if let Some(value) = value {
+                    env::set_var(var, value);
+                }
+            }
+        }
+
+        fn create_engine(msg_info: MessageInfo) -> Result<Engine> {
+            Engine::from_path(get_container_engine()?, Some(false), msg_info)
+        }
+
+        fn cargo_metadata(subdir: bool, msg_info: MessageInfo) -> Result<CargoMetadata> {
+            let mut metadata = cargo_metadata_with_args(
+                Some(Path::new(env!("CARGO_MANIFEST_DIR"))),
+                None,
+                msg_info,
+            )?
+            .ok_or_else(|| eyre::eyre!("could not find cross workspace"))?;
+
+            let root = match subdir {
+                true => get_cwd()?.join("member"),
+                false => get_cwd()?.parent().unwrap().to_path_buf(),
+            };
+            fs::create_dir_all(&root)?;
+            metadata.workspace_root = root;
+            metadata.target_directory = metadata.workspace_root.join("target");
+
+            Ok(metadata)
+        }
+
+        fn home() -> Result<PathBuf> {
+            home::home_dir().ok_or_else(|| eyre::eyre!("could not find home directory"))
+        }
+
+        fn get_cwd() -> Result<PathBuf> {
+            // we need this directory to exist for Windows
+            let path = temp::dir()?.join("Documents").join("package");
+            fs::create_dir_all(&path)?;
+            Ok(path)
+        }
+
+        fn get_sysroot() -> Result<PathBuf> {
+            Ok(home()?
+                .join(".rustup")
+                .join("toolchains")
+                .join("stable-x86_64-unknown-linux-gnu"))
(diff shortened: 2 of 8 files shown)


Alexhuszagh:
Since this would slow down our CI by quite a bit by default (`bisect` took 8 min, `remote` took 4 min), while most of our other actions take ~1m or less, I've changed it so it only runs if we use `bors`, so this works on PRs or if we try. (I'm using an invalid target here intentionally).

bors try --target x

Alexhuszagh:
bors try --target xyzabcdef

Alexhuszagh:
bors try --target xyzabcdef

Alexhuszagh:
Not sure why the CI failed, maybe it has to do with `build`?

Alexhuszagh:

Alexhuszagh:

Alexhuszagh:
Marking as a draft since #878 will affect this (specifically, the docker-in-docker tests).

Alexhuszagh:

Alexhuszagh:

Alexhuszagh:

Alexhuszagh:

Alexhuszagh:
bors try-

Alexhuszagh:

.github/workflows/ci.yml
```diff
@@ -321,8 +321,49 @@ jobs:
...
+    if: github.actor == 'bors[bot]'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-rust
+
+      - name: LLVM instrument coverage
+        id: cov
+        uses: ./.github/actions/cargo-llvm-cov
+        with:
+          name: cross-${{matrix.pretty}}
```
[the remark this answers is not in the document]
Emilgardis: should be `integration-<reason>` I think

Emilgardis:

Emilgardis:
Nice, coverage works now!

https://app.codecov.io/gh/cross-rs/cross/compare/873?src=pr&el=h1

Alexhuszagh:
We should re-add the docker-in-docker tests here, so we can run the test suite and ensure the volume finder works, and test with custom mounts (so #893 never regresses). We can have a separate test for the cross image itself, I believe.

Alexhuszagh:
Currently waiting on a patch for #901.

Emilgardis:

Alexhuszagh:
Mistakenly had an assumption that a container engine would be available for the tests, which isn't a requirement, since the docker-in-docker test needs one to find the proper mount paths. I've gated that test to only run on Linux, if the container engine and hostname are set, and the container engine can inspect the hostname.

src/docker/shared.rs
```diff
@@ -763,6 +763,164 @@ mod tests {
...
+
+            reset_env(vars);
+            Ok(())
+        }
+
+        #[test]
+        fn test_docker_in_docker() -> Result<()> {
+            if !cfg!(target_os = "linux") {
+                return Ok(());
+            }
```
Emilgardis, line 882:
```suggestion
        #[test]
        #[cfg_attr(not(target_os = "linux"), ignore)]
        fn test_docker_in_docker() -> Result<()> {
```

src/docker/shared.rs
```diff
@@ -763,6 +763,164 @@ mod tests {
...
+
+            let vars = unset_env();
+
+            let engine = create_engine(MessageInfo::default());
+            let hostname = env::var("HOSTNAME");
+            if engine.is_err() || hostname.is_err() {
+                reset_env(vars);
+                return Ok(());
```
Emilgardis, line 890:
```suggestion
                reset_env(vars);
                eprintln!("no hostname found");
                return Ok(());
```

src/docker/shared.rs
```diff
@@ -763,6 +763,164 @@ mod tests {
...
+            }
+            let engine = engine.unwrap();
+            let hostname = hostname.unwrap();
+            let output = subcommand(&engine, "inspect")
+                .arg(hostname)
+                .run_and_get_output(MessageInfo::default())?;
+            if !output.status.success() {
+                reset_env(vars);
+                return Ok(());
+            }
```
Emilgardis, line 900:
```suggestion
            if !output.status.success() {
                reset_env(vars);
                eprintln!("inspect failed");
                return Ok(());
            }
```

src/docker/shared.rs
```diff
@@ -763,6 +763,164 @@ mod tests {
...
+                None,
+                msg_info,
+            )?
+            .ok_or_else(|| eyre::eyre!("could not find cross workspace"))?;
+
+            metadata.workspace_root = get_workspace_root(subdir)?;
+            metadata.target_directory = metadata.workspace_root.join("target");
+
+            Ok(metadata)
+        }
```
Emilgardis, line 815: Why not: `tests::get_cargo_workspace()` here?

Emilgardis (changes requested):
I don't understand how the dind test is supposed to help. For example I don't see that it's even tested in CI right now, it just silently fails.

Alexhuszagh:

Alexhuszagh:
> I don't understand how the dind test is supposed to help. For example I don't see that it's even tested in CI right now, it just silently fails.

It should be tested. Also, the goal is to prevent a regression like in #901.

Emilgardis:
I mean this one: https://github.com/cross-rs/cross/runs/7169829781?check_suite_focus=true#step:5:46

The unit test. I'm pretty sure it's exiting early, and that because `HOSTNAME` is not defined

Emilgardis:
https://app.codecov.io/gh/cross-rs/cross/compare/873/tree/src/docker/shared.rs#D1L891 see coverage

Alexhuszagh:
> I mean this one: https://github.com/cross-rs/cross/runs/7169829781?check_suite_focus=true#step:5:46
> 
> The unit test. I'm pretty sure it's exiting early, and that because `HOSTNAME` is not defined

That's intentional, so it doesn't error if docker-in-docker isn't being used. I believe `HOSTNAME` is probably defined, but `docker inspect $HOSTNAME` fails, and so we skip the test. This test, however, shouldn't skip:

https://github.com/cross-rs/cross/runs/7169840511?check_suite_focus=true#step:5:359

Testing it locally with a panic after will fail.

```rust
#[test]
fn test_docker_in_docker() -> Result<()> {
    if !cfg!(target_os = "linux") {
        return Ok(());
    }

    let vars = unset_env();

    let engine = create_engine(MessageInfo::default());
    let hostname = env::var("HOSTNAME");
    if engine.is_err() || hostname.is_err() {
        reset_env(vars);
        return Ok(());
    }
    let engine = engine.unwrap();
    let hostname = hostname.unwrap();
    let output = subcommand(&engine, "inspect")
        .arg(hostname)
        .run_and_get_output(MessageInfo::default())?;
    if !output.status.success() {
        reset_env(vars);
        return Ok(());
    }
    panic!("got past inspect");
    
    ...

    Ok(())
}
```

src/docker/shared.rs
Alexhuszagh: Do we need this diagnostic? It's intended to return early if we're not running inside docker.

Alexhuszagh:
> https://app.codecov.io/gh/cross-rs/cross/compare/873/tree/src/docker/shared.rs#D1L891 see coverage

Ah that's unusual. Not sure why that's the case, it should be defining `HOSTNAME` inside the image. I guess the diagnostics are good, then.

Emilgardis:
Aha, I get it now!

You'll have to forward `RUSTFLAGS` inside the containered container, it's probably doing everything correct just not providing coverage.

See https://github.com/cross-rs/cross/runs/7169840511?check_suite_focus=true#step:19:22 where there's no profraw files and https://github.com/cross-rs/cross/runs/7169840371?check_suite_focus=true#step:10:7 where there is

src/docker/shared.rs
Alexhuszagh: I changed it to mention that either the container engine or hostname isn't defined. It's not very likely `HOSTNAME` won't be defined.

Alexhuszagh:

Emilgardis:
hmm, still not working

```
      Running `CARGO_INCREMENTAL=0 /home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-profdata merge -sparse -o /home/runner/work/cross/cross/target/cross.profdata`
added 18 packages, and audited 19 packages in 1s
2 packages are looking for funding
  run `npm fund` for details
found 0 vulnerabilities
added 3 packages, changed 3 packages, and audited 22 packages in 360ms
3 packages are looking for funding
  run `npm fund` for details
found 0 vulnerabilities
error: no input files specified. See llvm-profdata merge -help
```

I don't know why.

Alexhuszagh:
I have no idea either, I might need to also forward all of these as well:
- `LLVM_PROFILE_FILE`
- `CARGO_INCREMENTAL`
- `RUST_TEST_THREADS`

Emilgardis:
oh.. ofc, need to forward `LLVM_PROFILE_FILE` too...

Emilgardis:
Incremental would be good to disable, as we don't use the incremental data

Alexhuszagh:
> Incremental would be good to disable, as we don't use the incremental data

I've just forwarded everything, since if they're provided by the action, there's probably a good reason for them.

Alexhuszagh:
Code coverage still isn't working. At least the code tests to work on my machine.

Emilgardis:
We need to make the target folder available at the same path/the path in LLVM_PROFILE, easiest solution is to do `-v /home/runner/work/cross/cross/target/:/home/runner/work/cross/cross/target/`

Alexhuszagh:

Alexhuszagh:
Nothing is working so I'm going to throw in a panic and see if it just "works".

Alexhuszagh:
Let's see if it panics:

Emilgardis:
It is working, no? https://github.com/cross-rs/cross/runs/7170509736

Alexhuszagh:
> It is working, no? https://github.com/cross-rs/cross/runs/7170509736

Wait it is, maybe the code coverage I got was outdated. I'll squash the commits into one.

Emilgardis (approved):
bors r+

awesome

PR #877, opened 2022-06-27
Switch crosstool-ng configure scripts to xtask.

Alexhuszagh:
Rewrite the bash build scripts to Rust, allowing a better CLI interface, easier maintenance, and better help messages via clap.


Alexhuszagh:
This is related to the discussion [here](https://github.com/cross-rs/cross/discussions/875#discussioncomment-3031380).

xtask/src/crosstool.rs
```diff
@@ -0,0 +1,227 @@
...
+        let basename = file.file_name().to_utf8()?.to_string();
+        if let Some(target) = basename.strip_suffix(".config.in") {
+            targets.push(target.to_string());
+        }
+    }
+
+    Ok(targets)
+}
+
+fn configure_target(
```
Emilgardis, line 71: I feel like all the if statements here on major and minor can be combined into a `match (minor, major)` with either range patterns or if guards.

Emilgardis (approved):
I dislike the repetitive if blocks, but that just leaves more room for improvement, and if it works it works. LGTM!

bors r+

PR #878, opened 2022-06-27
Add dockerfile for cross.

### LINKED ISSUE #468: Images with cargo and rustc built-in?
Great docker images are provided. When used with cross tool it works great. However, cross can not be invoked in any of the CI systems running builds inside a container. I have found what inheriting the images from "rust" base image instead or ubuntu/debian addresses the issue and the images can be used with cross and in CI too.

Is it possible to modify the images so they are inherited from "rust" image, or alternatively, where rustup/cargo/rustc are installed as a last step? If you would like to keep the current images smallish, is it possible to provide an additional set of the same images where rustup/cargo are installed too?

If these images were provided, I believe rust developers would consolidate and use these, instead of working around the issue creating private not-well-maintained derivatives from your images.

I am keen to submit pull request if you are ok with any of the above options.

Makes sense?

Alexhuszagh:
This uses the official Docker docker-in-docker image for the latest stable, and installs the latest stable Rust version internally. It uses an entrypoint to ensure cargo is on the path, and that cross is installed from the latest git (locked), which when tagged will work well with our releases. It exports the environment variable `CROSS_CONTAINER_IN_CONTAINER` so everything should work as expected.

Closes #468.
Replaces #667.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,7 @@ jobs:
             - { target: thumbv7em-none-eabi,              os: ubuntu-latest,                    std: 1 }
             - { target: thumbv7em-none-eabihf,            os: ubuntu-latest,                    std: 1 }
             - { target: thumbv7m-none-eabi,               os: ubuntu-latest,                    std: 1 }
+            - { target: cross,                            os: ubuntu-latest }
 
   build:
     name: target (${{ matrix.pretty }},${{ matrix.os }})
@@ -283,7 +284,7 @@ jobs:
           IMAGE: ${{ steps.build-docker-image.outputs.image }}
         shell: bash
       - name: Test Image
-        if: steps.prepare-meta.outputs.has-image
+        if: steps.prepare-meta.outputs.has-image && steps.prepare-meta.outputs.test-variant == 'default'
         run: ./ci/test.sh
         env:
           TARGET: ${{ matrix.target }}
@@ -299,6 +300,13 @@ jobs:
         with:
           target: ${{ matrix.target }}
           image: ${{ steps.build-docker-image.outputs.image }}
+
+      - name: Test Cross Image
+        if: steps.prepare-meta.outputs.has-image && steps.prepare-meta.outputs.test-variant == 'cross'
+        run: ./ci/test-cross-image.sh
+        env:
+          TARGET: 'aarch64-unknown-linux-gnu'
+        shell: bash
 
       - name: Login to GitHub Container Registry
         if: steps.prepare-meta.outputs.has-image
--- a/ci/test-cross-image.sh
+++ b/ci/test-cross-image.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2086
+
+set -x
+set -eo pipefail
+
+if [[ -z "${TARGET}" ]]; then
+    export TARGET="aarch64-unknown-linux-gnu"
+fi
+if [[ -z "${CROSS_TARGET_CROSS_IMAGE}" ]]; then
+    CROSS_TARGET_CROSS_IMAGE="ghcr.io/cross-rs/cross:main"
+fi
+
+main() {
+    docker run --rm -e TARGET \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        "${CROSS_TARGET_CROSS_IMAGE}" sh -c '
+#!/usr/bin/env sh
+td="$(mktemp -d)"
+git clone --depth 1 https://github.com/cross-rs/rust-cpp-hello-word "${td}"
+cd "${td}"
+cross run --target "${TARGET}"
+'
+}
+
+main "${@}"
--- a/docker/Dockerfile.cross
+++ b/docker/Dockerfile.cross
@@ -1,0 +1,22 @@
+FROM ubuntu:20.04 as rust
+ARG DEBIAN_FRONTEND=noninteractive
+COPY docker/lib.sh docker/cross.sh /
+COPY ./ /project
+RUN /cross.sh /project
+
+# we build our images in 2 steps, to ensure we have a compact
+# image, since we want to add our current subdirectory
+FROM ubuntu:20.04
+COPY --from=rust /root/.cargo /root/.cargo
+COPY --from=rust /root/.rustup /root/.rustup
+
+# need some basic devtools, and requirements for docker
+RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
+    ca-certificates \
+    curl \
+    git
+
+RUN curl -fsSL https://get.docker.com | sh
+
+ENV CROSS_CONTAINER_IN_CONTAINER=1 \
+    PATH=/root/.cargo/bin:$PATH
--- a/docker/cross.sh
+++ b/docker/cross.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1090,SC1091
+
+set -x
+set -euo pipefail
+
+. lib.sh
+
+main() {
+    local project_dir="${1}"
+
+    install_packages ca-certificates curl gcc libc6-dev
+
+    cd "${project_dir}"
+    curl --proto "=https" --tlsv1.2 --retry 3 -sSfL https://sh.rustup.rs | sh -s -- -y
+    source "${HOME}"/.cargo/env
+    cargo install --path . --locked
+
+    purge_packages
+
+    rm -rf "${0}"
+}
+
+main "${@}"
--- a/xtask/src/ci.rs
+++ b/xtask/src/ci.rs
@@ -40,10 +40,11 @@ pub fn ci(args: CiJob, metadata: CargoMetadata) -> cross::Result<()> {
             // Set labels
             let mut labels = vec![];
 
-            labels.push(format!(
-                "org.opencontainers.image.title=cross (for {})",
-                target.triplet
-            ));
+            let image_title = match target.triplet.as_ref() {
+                "cross" => target.triplet.to_string(),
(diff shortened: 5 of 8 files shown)


Alexhuszagh:
For some reason, the tags aren't working:
https://github.com/cross-rs/cross/runs/7082679487?check_suite_focus=true#step:5:13

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try-
bors try --target cross

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try-
bors try --target cross

Alexhuszagh:
bors try --target cross

Alexhuszagh:
I managed to get this to work by effectively skipping the test if it was a Docker image for `cross`, but that's definitely not the solution here. Any suggestions would be great, since my idea was to skip testing if the `matrix.target == 'cross'`, but I didn't have any luck with:
- `${{ matrix.target != 'cross' }}`
- `${{ matrix.target }} != 'cross'`

Emilgardis (changes requested):
Does this entry point approach not incur quite a bit of startup?

Also, this doesnt test the code in the branch or even local base, but current main

Emilgardis:
> I managed to get this to work by effectively skipping the test if it was a Docker image for `cross`, but that's definitely not the solution here. Any suggestions would be great, since my idea was to skip testing if the `matrix.target == 'cross'`, but I didn't have any luck with:
> 
>     * `${{ matrix.target != 'cross' }}`
> 
>     * `${{ matrix.target }} != 'cross'`

Instead of doing that, better approach I think would be to make something similar to https://github.com/cross-rs/cross/blob/eb2b8ad4aa521953fdf36decc99508c332df9dd0/xtask/src/ci.rs#L70, and gate the test on that

The first alternative is correct also, however in if statements you dont need `${{}}`

Alexhuszagh:
Yeah I probably need another test script or something. Also, it's probably a good idea to move this to Ubuntu, since all our other images are. I think having a single Alpine image probably isn't worth it.

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try-

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try-
bors try --target cross

Alexhuszagh:
bors try --target cross

ci/test-cross-image.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
Alexhuszagh, line 1: I've added this so we can test docker-in-docker scenarios where the `CROSS_CONTAINER_IN_CONTAINER` envvar has already been set.

docker/Dockerfile.cross
```diff
@@ -0,0 +1,22 @@
+# we build our images in 2 steps, to ensure we have a compact
+# image, since we want to add our current subdirectory
+
+FROM ubuntu:20.04 as rust
+ARG DEBIAN_FRONTEND=noninteractive
+COPY lib.sh cross.sh /
+RUN /cross.sh
+
+FROM ubuntu:20.04
```
Alexhuszagh, line 9: I've decided to Ubuntu as a base so it's consistent with all our other images.

.github/workflows/ci.yml
```diff
@@ -270,7 +271,7 @@ jobs:
           IMAGE: ${{ steps.build-docker-image.outputs.image }}
         shell: bash
       - name: Test Image
-        if: steps.prepare-meta.outputs.has-image && (github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/trying')
+        if: (steps.prepare-meta.outputs.has-image && (github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/trying')) && matrix.target != 'cross'
```
Alexhuszagh, line 274: I think this is the right approach, but feedback would be nice. I want separate logic for our cross image and everything else, so I've split the tests into 2: one for cross and one for everything else.

ci/test-cross-image.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -x
+set -eo pipefail
+
+if [[ -z "${TARGET}" ]]; then
+    export TARGET="aarch64-unknown-linux-gnu"
+fi
+if [[ -z "${CROSS_TARGET_CROSS_IMAGE}" ]]; then
+    CROSS_TARGET_CROSS_IMAGE="ghcr.io/cross-rs/cross:main"
+fi
```
Alexhuszagh, line 12: This way we use the CI default if present, but have a reasonable target if testing our CI configurations locally.

ci/test-cross-image.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2086
+
+set -x
+set -eo pipefail
+
+if [[ -z "${TARGET}" ]]; then
+    export TARGET="aarch64-unknown-linux-gnu"
```
Alexhuszagh, line 8: We should always have this image, so it's a sensible target.

.github/workflows/ci.yml
```diff
@@ -287,6 +288,18 @@ jobs:
...
+        if: (steps.prepare-meta.outputs.has-image && (github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/trying')) && matrix.target == 'cross'
+        run: ./ci/test-cross-image.sh
+        env:
+          TARGET: 'aarch64-unknown-linux-gnu'
+        shell: bash
+      - uses: ./.github/actions/cargo-install-upload-artifacts
+        if: matrix.deploy
+        with:
+          target: ${{ matrix.target }}
+          image: ${{ steps.build-docker-image.outputs.image }}
```
Emilgardis, line 301: again? See line L285-L289. Those lines should be removed if this is kept

docker/Dockerfile.cross
```diff
@@ -0,0 +1,22 @@
+# we build our images in 2 steps, to ensure we have a compact
+# image, since we want to add our current subdirectory
```
Emilgardis, line 2: This seems like it's a header, but it's more a comment on internals, needs a proper header I think

.github/workflows/ci.yml
Alexhuszagh: Ah my mistake, is this only keeping the artifacts for cross itself? I assumed it was the artifacts for the Docker image, but that's probably my mistake.

docker/Dockerfile.cross
Emilgardis: Sounds good

Alexhuszagh:
bors try --target cross

.github/workflows/ci.yml
Alexhuszagh: I removed these lines from `Test Cross Image`. I also didn't add LLVM coverage to this PR, since it makes little sense (everything is running inside a container and with a different target directory/etc).

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try --target cross

Alexhuszagh:
bors try-
bors try --target cross

xtask/src/ci.rs
```diff
@@ -69,6 +69,9 @@ pub fn ci(args: CiJob, metadata: CargoMetadata) -> cross::Result<()> {
             if target.has_ci_image() {
                 gha_output("has-image", "true")
             }
+            if target.is_default_test_image() {
```
Alexhuszagh, line 72: I've added this so when `cross` and `zig` images are added, we can just exclude them with a single `matches!` statement later.

Alexhuszagh:
bors try --target cross

xtask/src/ci.rs
Alexhuszagh: I've also implemented the same change mentioned [here](https://github.com/cross-rs/cross/pull/880#pullrequestreview-1025458032) for the zig image.

Alexhuszagh:
bors try --target cross

docker/cross.sh
```diff
@@ -0,0 +1,23 @@
...
+. lib.sh
+
+main() {
+    install_packages ca-certificates curl gcc libc6-dev
+
+    # since we can't share outside the current docker context,
+    # we need to install from git, rather than from path
+    curl --proto "=https" --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+    source "$HOME"/.cargo/env
+    cargo install cross --git https://github.com/cross-rs/cross --locked
```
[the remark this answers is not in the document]
Alexhuszagh: I somewhat avoided that, because it would require the script to be sufficiently distant from `docker buildx`, but I guess we could do it. It would require a release build of `cross` inside `build_docker_image`, and then make the binaries (both `cross` and `cross-util`) available to the Dockerfile context (either by making it the project directory, or via copying), and then the build can start. I wasn't sure if that was the best approach.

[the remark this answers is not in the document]
Alexhuszagh: We could, somewhat easier, make the parent directory the Dockerfile context just for this image, build it inside the container via a script via a temporary layer, and then copy the built binary `using --from`.

Alexhuszagh:
I changed the Dockerfile context to the workspace root, and then use that in the build process, so we can copy over the entire project over from the current source code, to ensure the image reflects the current project state.

bors try --target cross

Emilgardis (commented):
LGTM!

One thing missing is changing the label for title, it should probably be simply "cross" or similar

Alexhuszagh:
bors try --target cross

Emilgardis (approved):
lgtm!

PR #879, opened 2022-06-28
always run tests

Emilgardis:


Emilgardis:
CodeCov did not expect the default branch to be named `main`, so the yml did not get picked up for flags. This intentionally touches the `codecov.yml` file so that it get's recognized, but also makes it so that tests are always run, even on main, so that the project doesn't only get tested in staging, but on merge as well (so that coverage get's recorded on the main branch and not just in staging).


There could be a way to use the same artifacts as the sucessfull staging build did and just upload them, but I'm not sure how to do that without going with something like S3 or artifactory (there's no official api to download "any" artifact except for current workflow artifacts)

Emilgardis:
There is also an argument to this that could be used.

When we upload the artifacts, we've already signaled to bors that the run was successfull and to merge, we could possibly set the `--branch` to `main` on staging builds

Alexhuszagh:
Is there a particular reason why this wasn't done before, potentially CI times?

Alexhuszagh (approved):
Seems like it was introduced in d1959de, so it seems that's just how it was always done. Should make the rest of the CI config logic for if/then tests easier as well. Since the tests are never that long, looks like an improvement in every case.

PR #880, opened 2022-06-29
Add a `zig cc`-based image.

### LINKED ISSUE #860: Provide a `zig cc`-based image.
### Checklist

### Describe your request

Inspired by the suggestion on [reddit](https://www.reddit.com/r/rust/comments/vk2xfc/comment/idnrw3v/?utm_source=reddit&utm_medium=web2x&context=3), which would then allow users to favor using `zig cc` or similar, which would allow use of custom glibc versions, etc.

Effectively, we'd have a config option like:

```toml
[build]
use-zig = true
```

And then a composite image for all targets `zig cc` supports, and it would ensure we set the appropriate `CC_$target` and `CXX_$target` environment variables to `zig cc` and `zig c++` appropriately.

It would allow passing in custom glibc versions for `*-linux-gnu` targets, if done correctly, and could reduce image sizes by a lot. The only caveat would be is issues with `zig` respecting all C, C++ and link flags, which I know have been issues in the past. Making it opt-in, however, solves all of these.

Alexhuszagh:
Uses cargo-zigbuild as a backend, and adds configuration options for zig under `[build.zig]` and `[target.(...).zig]`. If enabled, and an image override is not provided, `cross` will always use the `zig` image.

It supports custom `glibc` versions by parsing the libc portion of the target, and extracting a the libc version if present. The target, if built-in, is then the triple/libc pair, otherwise, it's just the triple.

The image does not provide runners, `bindgen` Clang args, or `pkg-config` paths, since `zig cc` does not provide the dynamic library loader (`ld-linux*.so`) required, meaning none of the binaries can be run. For `bindgen`, `zig cc` has an unusual directory structure, so there is no traditional sysroot with `usr`, `lib`, and `include` subdirectories. Finally, since we don't have system packages we can work with, exporting a `pkg-config` path makes little sense.

Closes #860.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,7 @@ jobs:
             - { target: thumbv7em-none-eabihf,            os: ubuntu-latest,                    std: 1 }
             - { target: thumbv7m-none-eabi,               os: ubuntu-latest,                    std: 1 }
             - { target: cross,                            os: ubuntu-latest }
+            - { target: zig,                              os: ubuntu-latest }
 
   build:
     name: target (${{ matrix.pretty }},${{ matrix.os }})
@@ -247,7 +248,7 @@ jobs:
       - name: LLVM instrument coverage
         id: cov
         uses: ./.github/actions/cargo-llvm-cov
-        if: steps.prepare-meta.outputs.has-image
+        if: steps.prepare-meta.outputs.has-image  && steps.prepare-meta.outputs.test-variant != 'zig'
         with:
           name: cross-${{matrix.pretty}}
 
@@ -301,6 +302,11 @@ jobs:
         with:
           target: ${{ matrix.target }}
           image: ${{ steps.build-docker-image.outputs.image }}
+
+      - name: Test Zig Image
+        if: steps.prepare-meta.outputs.has-image && steps.prepare-meta.outputs.test-variant == 'zig'
+        run: ./ci/test-zig-image.sh
+        shell: bash
 
       - name: Test Cross Image
         if: steps.prepare-meta.outputs.has-image && steps.prepare-meta.outputs.test-variant == 'cross'
--- a/ci/test-zig-image.sh
+++ b/ci/test-zig-image.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2086,SC1091,SC1090
+
+set -x
+set -eo pipefail
+
+# NOTE: "${@}" is an unbound variable for bash 3.2, which is the
+# installed version on macOS. likewise, "${var[@]}" is an unbound
+# error if var is an empty array.
+
+ci_dir=$(dirname "${BASH_SOURCE[0]}")
+ci_dir=$(realpath "${ci_dir}")
+. "${ci_dir}"/shared.sh
+
+# zig cc is very slow: only use a few targets.
+TARGETS=(
+    "aarch64-unknown-linux-gnu"
+    "aarch64-unknown-linux-musl"
+    "i586-unknown-linux-gnu"
+    "i586-unknown-linux-musl"
+)
+
+# on CI, it sets `CROSS_TARGET_ZIG_IMAGE` rather than `CROSS_TARGET_ZIG_IMAGE`
+if [[ -n "${CROSS_TARGET_ZIG_IMAGE}" ]]; then
+    export CROSS_BUILD_ZIG_IMAGE="${CROSS_TARGET_ZIG_IMAGE}"
+    unset CROSS_TARGET_ZIG_IMAGE
+fi
+
+main() {
+    export CROSS_BUILD_ZIG=1
+
+    local td=
+    local target=
+
+    retry cargo fetch
+    cargo build
+    export CROSS="${PROJECT_HOME}/target/debug/cross"
+
+    td="$(mktemp -d)"
+    git clone --depth 1 https://github.com/cross-rs/rust-cpp-hello-word "${td}"
+    pushd "${td}"
+
+    for target in "${TARGETS[@]}"; do
+        CROSS_CONTAINER_ENGINE="${CROSS_ENGINE}" "${CROSS}" build --target "${target}" --verbose
+        # note: ensure #724 doesn't replicate during CI.
+        # https://github.com/cross-rs/cross/issues/724
+        cargo clean
+    done
+
+    popd
+    rm -rf "${td}"
+}
+
+main "${@}"
--- a/docker/zig.sh
+++ b/docker/zig.sh
@@ -1,0 +1,197 @@
+#!/usr/bin/env bash
+
+set -x
+set -eo pipefail
+
+# shellcheck disable=SC1091
+. lib.sh
+
+main() {
+    local platform="${1}"
+    install_packages ca-certificates curl xz-utils
+
+    install_zig "${platform}"
+    install_zigbuild "${platform}"
+
+    purge_packages
+    rm "${0}"
+}
+
+install_zig() {
+    local platform="${1}"
+    local version="0.9.1"
+    local dst="/opt/zig"
+    local arch=
+    local os=
+    local triple=
+
+    case "${platform}" in
+        'linux/386')
+            arch="i386"
+            os="linux"
+            ;;
+        'linux/amd64')
+            arch="x86_64"
+            os="linux"
+            ;;
+        'linux/arm64')
+            arch="aarch64"
+            os="linux"
+            ;;
+        'linux/riscv64')
+            arch="riscv64"
+            os="linux"
+            ;;
+        'linux/ppc64le')
+            triple="powerpc64le-linux-gnu"
+            ;;
+        'linux/s390x')
+            triple="s390x-linux-gnu"
+            ;;
+        'darwin/amd64')
+            arch="x86_64"
+            os="macos"
+            ;;
+        'darwin/arm64')
+            arch="aarch64"
+            os="macos"
+            ;;
+        # NOTE: explicitly don't support linux/arm/v6
+        *)
+            echo "Unsupported target platform '${platform}'" 1>&2
+            exit 1
+            ;;
+    esac
+
+    if [[ -n "${arch}" ]]; then
+        install_zig_tarball "${arch}" "${os}" "${version}" "${dst}"
+    else
+        install_zig_source "${triple}" "${version}" "${dst}"
+    fi
+}
+
+install_zig_tarball() {
+    local arch="${1}"
+    local os="${2}"
+    local version="${3}"
+    local dst="${4}"
+    local filename="zig-${os}-${arch}-${version}.tar.xz"
+
+    local td
+    td="$(mktemp -d)"
+
+    pushd "${td}"
+
+    curl --retry 3 -sSfL "https://ziglang.org/download/${version}/${filename}" -O
+    mkdir -p "${dst}"
+    tar --strip-components=1 -xJf "${filename}" --directory "${dst}"
+
+    popd
+
+    rm -rf "${td}"
+}
+
+install_zig_source() {
+    local triple="${1}"
+    local version="${2}"
+    local dst="${3}"
+    local filename="zig-bootstrap-${version}.tar.xz"
+
+    local td
+    td="$(mktemp -d)"
+
+    pushd "${td}"
+
+    curl --retry 3 -sSfL "https://ziglang.org/download/${version}/${filename}" -O
+    mkdir zig
+    tar --strip-components=1 -xJf "${filename}" --directory zig
+
+    pushd zig
+    install_packages python3 make g++
+    ./build -j5 "${triple}" native
+    mv "out/zig-${triple}-native" /opt/zig
+
+    popd
+    popd
+
+    rm -rf "${td}"
+}
+
+install_zigbuild() {
+    local platform="${1}"
+    local version=0.11.0
+    local dst="/usr/local"
+    local triple=
+
+    # we don't know if `linux/arm/v7` is hard-float,
+    # and we don't know the the zigbuild `apple-darwin`
+    # target doesn't manually specify the architecture.
+    case "${platform}" in
+        'linux/386')
+            triple="i686-unknown-linux-musl"
+            ;;
+        'linux/amd64')
+            triple="x86_64-unknown-linux-musl"
+            ;;
+        'linux/arm64')
+            triple="aarch64-unknown-linux-musl"
+            ;;
+        *)
+            ;;
+    esac
+
+    if [[ -n "${triple}" ]]; then
+        install_zigbuild_tarball "${triple}" "${version}" "${dst}"
+    else
+        install_zigbuild_source "${version}" "${dst}"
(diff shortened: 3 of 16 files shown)


Alexhuszagh:
I still need to figure out the best way to add it to CI (it's currently not built as part of the matrix), since it can target numerous other architectures. It might be good to just change the exported triple from `zig` to a well-known target like `aarch64-unknown-linux-gnu`, which supports C++, etc.

src/lib.rs
```diff
@@ -150,14 +150,20 @@ impl<'a> From<&'a str> for Host {
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
 #[serde(from = "String")]
 pub enum Target {
-    BuiltIn { triple: String },
-    Custom { triple: String },
+    BuiltIn {
+        triple: String,
+        libc: Option<String>,
```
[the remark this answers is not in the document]
Alexhuszagh: > this seems wierd, think there's a better way to encapsulate it

How? We split the triple for built-in images, which are the ones where we can add targets for, into the actual triple and the libc version, if present. The serialized triple is the same for the docker command, so it correctly has the libc version when passed to `zigbuild`, but for rustup and adding targets, etc., it uses the regular triple.

Alexhuszagh:
bors try --target zig

Alexhuszagh:
bors try-
bors try --target zig

src/lib.rs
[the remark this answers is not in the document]
Emilgardis: The reason I think it's wierd is because it's not real syntax compatible with cargo.

I don't think the version should be a property of the target, as it's not part of the target string outside zig.

If we use this way, there's no good way to specify glibc without always specifying the version in `--target`.

Instead, I think the version specifier should be a property in the config, possibly

```toml
[target.$target.zig]
version = "2.17"
```

and `Target` should not have the version encoded.

the big thing I'm seeing is that 

1. --target $target.$glibc is not a real target according to rustc/cargo
2. you'll have to always specify the version number when using `*.zig` to be correct
3. the glibc version becomes promoted, i.e always forwarded, even if it's only used in one case.

Alexhuszagh:
bors try --target zig

Alexhuszagh:
I've changed the format entirely.

`cross ... --target` now only accepts valid Rust triples, and the config format is as follows:

```toml
[target.x86_64-unknown-linux-gnu.zig]
enable = true       # enable use of the zig image
version = "2.17"    # glibc version to use
image = "ghcr.io/cross-rs/zig:local"    # custom image to use
```

If provided as a bool, it will use the default glibc version:

```toml
# equivalent to { enable = true }
zig = true
```

If provided as a string, `zig` will be automatically enabled:

```toml
# equivalent to { enable = true, version = "2.17" }
zig = "2.17"
```

A custom zig image is so users can provide their own, general-purpose zig image if needed. These values can also be provided via environment variables, such as `CROSS_BUILD_ZIG`, `CROSS_BUILD_ZIG_VERSION`, and `CROSS_BUILD_ZIG_IMAGE`, or similar values for specific targets.

src/lib.rs
```diff
@@ -169,8 +169,8 @@ impl Target {
 
     fn triple(&self) -> &str {
         match *self {
-            Target::BuiltIn { ref triple } => triple,
-            Target::Custom { ref triple } => triple,
+            Target::BuiltIn { ref triple, .. } => triple,
+            Target::Custom { ref triple, .. } => triple,
```
Emilgardis, line 173: future proofing?

src/lib.rs
```diff
@@ -353,11 +353,42 @@ impl From<String> for Target {
 
 impl Serialize for Target {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_str(&self.to_string())
```
Emilgardis, line 356: nice

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- #880 - added a zig-based image, allowing multiple targets to be built from the same image.
```
Emilgardis, line 12: mention cargo-zigbuild

ci/test-zig-image.sh
```diff
@@ -0,0 +1,54 @@
...
+    cargo build
+    export CROSS="${project_home}/target/debug/cross"
+
+    td="$(mktemp -d)"
+    git clone --depth 1 https://github.com/cross-rs/rust-cpp-hello-word "${td}"
+    pushd "${td}"
+
+    for target in "${TARGETS[@]}"; do
+        "${CROSS}" build --target "${target}"
+        # note: ensure #724 doesn't replicate during CI.
```
Emilgardis, line 46: should probably be an url

Emilgardis (changes requested):

src/lib.rs
Alexhuszagh: Uh whoops that was from when it used `libc` previously. This needs to be removed.

src/docker/shared.rs
```diff
@@ -532,7 +531,15 @@ pub(crate) fn image_name(config: &Config, target: &Target) -> Result<String> {
...
-    Ok(format!("{CROSS_IMAGE}/{target}:{version}"))
+    if uses_zig {
+        if let Some(image) = config.zig_image(target)? {
+            Ok(image)
+        } else {
+            Ok(format!("{CROSS_IMAGE}/zig:{version}"))
+        }
+    } else {
+        Ok(format!("{CROSS_IMAGE}/{target}:{version}"))
+    }
```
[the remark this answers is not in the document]
Alexhuszagh: Because `zig.enable` might not be active even if `zig` has an image. It makes sense if `build.zig.image` is set, but `target.(...).zig.enable` is deactivated. But there's probably a better way to do this, probably just getting the target name.

docker/zig.sh
```diff
@@ -0,0 +1,62 @@
...
+    export RUSTUP_HOME="${td}/rustup"
+    export CARGO_HOME="${td}/cargo"
+
+    curl --retry 3 -sSfL https://sh.rustup.rs -o rustup-init.sh
+    sh rustup-init.sh -y --no-modify-path --profile minimal
+
+    PATH="${CARGO_HOME}/bin:${PATH}" \
+        cargo install cargo-zigbuild \
+        --version "${version}" \
+        --root /usr/local
```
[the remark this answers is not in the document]
Alexhuszagh: I think building is the right approach, or else we have to:
1. Store the host architecture
2. Convert it to the correct binary

They only offer `*-musl` targets, so although that works fine with our `*-linux-gnu` images, it might get annoying, and we have to consider we won't support on `x86_64-unknown-linux-gnu` base images anymore, necessarily.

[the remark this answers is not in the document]
Emilgardis: aha! didn't look too closely

Alexhuszagh:
Actually we should provide an ARCH here. I'll mark this as a draft since we need to download `zig` from a binary anyway.

Alexhuszagh:
bors try --target zig

Ok I've changed the build in `zig.sh` to fetch tarballs when possible, or build from source when we can't. This way, at least we should support numerous different target platforms There's a good chance the bors command fails, but weirdly it works on my machine so we'll see....

Alexhuszagh:
Oh this is why it's failing:

```bash
$ /usr/bin/docker run --userns host -e 'PKG_CONFIG_ALLOW_CROSS=1' -e 'XARGO_HOME=/xargo' -e 'CARGO_HOME=/cargo' -e 'CARGO_TARGET_DIR=/target' -e 'CROSS_RUNNER=' -e CARGO_HTTP_CHECK_REVOKE -e RUSTFLAGS -e CARGO_NET_RETRY -e CARGO_INCREMENTAL -e 'USER=runner' --rm --user 1001:121 -v /home/runner/.xargo:/xargo:Z -v /home/runner/.cargo:/cargo:Z -v /cargo/bin -v /tmp/tmp.xdLqWsDyv2:/project:Z -v /home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu:/rust:Z,ro -v /tmp/tmp.xdLqWsDyv2/target:/target:Z -w /project ghcr.io/cross-rs/zig:trying sh -c 'PATH=$PATH:/rust/bin cargo-zigbuild build --target aarch64-unknown-linux-gnu --verbose'
```

Specifically `--rm --user 1001:121`, and we have `RUN adduser --uid 1000 --disabled-password --gecos "" cross` in the Dockerfile.

Emilgardis (approved):

xtask/src/ci.rs
```diff
@@ -69,6 +69,9 @@ pub fn ci(args: CiJob, metadata: CargoMetadata) -> cross::Result<()> {
             if target.has_ci_image() {
                 gha_output("has-image", "true")
             }
+            if target.is_default_test_image() {
+                gha_output("default-test", "true")
+            }
```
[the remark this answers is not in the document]
Alexhuszagh: Sounds good, that also works if we ever move them outside the matrix for whatever reason.

Alexhuszagh:
bors try --target zig

Alexhuszagh:
bors try --target zig

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #900 - add the option to skip copying build artifacts back to host when using remote cross via `CROSS_REMOTE_SKIP_BUILD_ARTIFACTS`.
 - #891 - support custom user namespace overrides by setting the `CROSS_CONTAINER_USER_NAMESPACE` environment variable.
 - #890 - support rootless docker via the `CROSS_ROOTLESS_CONTAINER_ENGINE` environment variable.
+- #880 - added a zig-based image, allowing multiple targets to be built from the same image, using cargo-zigbuild.
```
Emilgardis, line 22: changelog in wrong place

src/docker/shared.rs
```diff
@@ -472,6 +479,10 @@ pub(crate) fn docker_envvars(
         .args(&["-e", "CARGO_HOME=/cargo"])
         .args(&["-e", "CARGO_TARGET_DIR=/target"])
         .args(&["-e", &cross_runner]);
+    if uses_zig {
+        // otherwise, zig has a permission error trying to create the cache
+        docker.args(&["-e", "XDG_CACHE_HOME=/target/.zig-cache"]);
+    }
```
Emilgardis, line 485: I think this should take the cargo variant, instead of taking a bool.

Alexhuszagh:
Temporarily remarking this as a draft: it needs 2 things right now:
1. Fix adding the libc version to `--target` inside the docker image when the target is provided.
2. Fix CI.

I've aded a `ProvidedImage` for `zig` and also added it to use `PossibleImage` for a base.

Alexhuszagh:
bors try --target zig

Alexhuszagh:
bors try-
bors try --target zig

Alexhuszagh:
bors try --target zig

Alexhuszagh:
This won't work but the log will be helpful.
bors try --target zig

Alexhuszagh:
bors try --target zig

.github/workflows/ci.yml
```diff
@@ -247,7 +248,9 @@ jobs:
       - name: LLVM instrument coverage
         id: cov
         uses: ./.github/actions/cargo-llvm-cov
-        if: steps.prepare-meta.outputs.has-image
+        # instrument coverage breaks the zig image tests, which is
```
Alexhuszagh, line 251: It seems the issue was the LLVM code coverage, which caused the internal toolchain to fail.

Emilgardis: comment is unneccessary, can be removed

src/docker/custom.rs
```diff
@@ -68,6 +68,7 @@ impl<'a> Dockerfile<'a> {
         options: &DockerOptions,
         paths: &DockerPaths,
         build_args: impl IntoIterator<Item = (impl AsRef<str>, impl AsRef<str>)>,
+        uses_zig: bool,
         msg_info: &mut MessageInfo,
     ) -> Result<String> {
         let mut docker_build = docker::subcommand(&options.engine, "buildx");
```
Emilgardis, line 74: `DockerOptions` has the `CargoVariant`

Emilgardis (changes requested):

src/lib.rs
```diff
@@ -479,10 +521,12 @@ pub fn run(
             .unwrap_or_else(|| Target::from(host.triple(), &target_list));
         config.confusable_target(&target, msg_info)?;
 
+        let uses_zig = config.zig(&target).unwrap_or(false);
+        let zig_version = config.zig_version(&target)?;
         // Get the image we're supposed to base all our next actions on.
         // The image we actually run in might get changed with
         // `target.{{TARGET}}.dockerfile` or `target.{{TARGET}}.pre-build`
-        let image = match docker::get_image(&config, &target) {
+        let image = match docker::get_image(&config, &target, uses_zig) {
```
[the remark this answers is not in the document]
Alexhuszagh: I mean, the image needs to be dependent on zig, because if it's present, we use a different image. Also, we can't have a manual image for a given target and zig support, because we can't know outside the function without just redoing the logic if the returned image was from the config/environment or just the default from the target.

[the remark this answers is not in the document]
Alexhuszagh: I think keeping pre-build hooks is fine: it doesn't have any real issues with `zig cc` itself: it has a toolchain itself installed, can dynamically to libraries installed by a package manager, etc.

src/lib.rs
```diff
@@ -616,8 +662,24 @@ To override the toolchain mounted in the image, set `target.{}.image.toolchain =
...
+                        if let Some(triple) = args_iter.next() {
+                            filtered_args.push(add_libc(&triple));
+                        }
+                    } else if let Some(stripped) = arg.strip_prefix("--target=") {
+                        filtered_args.push(format!("--target={}", add_libc(stripped)));
+                    } else {
+                        filtered_args.push(arg);
+                    }
+                }
+                filtered_args
```
Emilgardis, line 682: do we need to add the zig specific `*.ver` here?

Emilgardis (changes requested):

src/lib.rs
Alexhuszagh: Yes, because otherwise we have to add it much, much later, and this doesn't change the `target`, or any other variables except those in the filtered args.

We could technically do it much later, but then we'd have to modify it for both remote and local, and we'd have to re-handle the logic of filtering the args. If we do it earlier, during CLI parsing, then we get the wrong target triple.

Alexhuszagh: I guess technically we could do it when adding it to the CLI arguments, but that feels iffy because it means modifying the logic, since we'd need to have already parsed `Cross.toml` which moves that significantly further up, and we'd have to then pass the Zig glibc version to the target callback. Which, IMO, feels odd.

Emilgardis (approved):
awesome!

Alexhuszagh:
bors r=Emilgardis

PR #882, opened 2022-06-29
Minor bug fixes.

Alexhuszagh:
Removes the requirement of an unused target from `cross-util volumes remove` and fixes the Ubuntu base testing.


Emilgardis (commented):
technically this is a breaking change due to removed required cli flag

Alexhuszagh:
> technically this is a breaking change due to removed required cli flag

True, should we just make it optional then and ignore it?

Emilgardis (approved):
bors r+

PR #890, opened 2022-07-01
Support rootless docker.

### LINKED ISSUE #889: Support Rootless Docker
### Checklist

### Describe your request

Currently, if using rootless docker, `cross` fails to run. This is because we automatically add `--user 1000:1000` permissions (or the current UID/GID) when running Docker. This is an issue, because there is now rootless docker:

First, install rootless docker:

```bash
$  dockerd-rootless-setuptool.sh install  # this may require a --force if the rootful docker is available.
```

Then, use the rootless context and try to touch a file:

```bash
$ docker context use rootless 
rootless
Current context is now "rootless"
$ docker run --user 1000:1000 -it --rm -v "$PWD":/project -w /project ubuntu:20.04 bash
groups: cannot find name for group ID 1000
$ touch a
touch: cannot touch 'a': Permission denied
```

This can be solved by allowing an environment variable to override our default, good assumptions of whether the container engine is rootful or not.

Currently, detecting rootful/rootless mode is quite difficult, or expensive computationally, and the defaults are quite good:
- Podman always runs rootless
- Docker mostly runs rootful

Therefore, just making these assumptions generally works. However, this is an issue if rootless docker exists, or we have another container engine that runs as root. In short, we need to be able to override setting `--user 1000:1000` permissions. This likely could best be done with a `CROSS_ROOTLESS_CONTAINER_ENGINE`, which is an `Option<bool>`, parsed via `bool_from_envvar`. If it's not present, use the sensible default. If it is present, force the presence or absence of lower user permissions.

This also should simplify supporting new container engines, since we can handle those with varying behavior quite easily, without any code changes, until we can provide reasonable defaults for them, such as in #588.

Alexhuszagh:
Adds support for rootless docker, and manually overriding rootless/rootful container engines through the `CROSS_ROOTLESS_CONTAINER_ENGINE` environment variable. If not set, it will use the default mode for the container engine (rootful for docker, rootless for everything else).

```bash
# use the defaults
cross run ...
# auto-select if using rootless (the default)
CROSS_ROOTLESS_CONTAINER_ENGINE=auto cross run ...
# always use rootful mode
CROSS_ROOTLESS_CONTAINER_ENGINE=0 cross run ...
# always use rootless mode
CROSS_ROOTLESS_CONTAINER_ENGINE=1 cross run ...
```

Closes #889.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- #890 - support rootless docker via the `CROSS_ROOTLESS_CONTAINER_ENGINE` environment variable.
+
 ### Changed
 
 - #869 - ensure cargo configuration environment variable flags are passed to the docker container.
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -6,7 +6,7 @@ use std::{env, fs};
 use super::custom::Dockerfile;
 use super::engine::*;
 use crate::cargo::{cargo_metadata_with_args, CargoMetadata};
-use crate::config::Config;
+use crate::config::{bool_from_envvar, Config};
 use crate::errors::*;
 use crate::extensions::{CommandExt, SafeCommand};
 use crate::file::{self, write_file, PathExt, ToUtf8};
@@ -399,8 +399,17 @@ pub(crate) fn group_id() -> String {
 }
 
 pub(crate) fn docker_user_id(docker: &mut Command, engine_type: EngineType) {
-    // We need to specify the user for Docker, but not for Podman.
-    if engine_type == EngineType::Docker {
+    // by default, docker runs as root so we need to specify the user
+    // so the resulting file permissions are for the current user.
+    // since we can have rootless docker, we provide an override.
+    let is_rootless = env::var("CROSS_ROOTLESS_CONTAINER_ENGINE")
+        .ok()
+        .and_then(|s| match s.as_ref() {
+            "auto" => None,
+            b => Some(bool_from_envvar(b)),
+        })
+        .unwrap_or_else(|| engine_type != EngineType::Docker);
+    if !is_rootless {
         docker.args(&["--user", &format!("{}:{}", user_id(), group_id(),)]);
     }
 }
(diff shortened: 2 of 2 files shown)


src/docker/shared.rs
```diff
@@ -399,8 +399,13 @@ pub(crate) fn group_id() -> String {
 }
 
 pub(crate) fn docker_user_id(docker: &mut Command, engine_type: EngineType) {
-    // We need to specify the user for Docker, but not for Podman.
-    if engine_type == EngineType::Docker {
+    // by default, docker runs as root so we need to specify the user
+    // so the resulting file permissions are for the current user.
+    // since we can have rootless docker, we provide an override.
+    let is_rootless = env::var("CROSS_ROOTLESS_CONTAINER_ENGINE")
+        .map(|s| bool_from_envvar(&s))
```
Emilgardis, line 406: I feel like this should take `1`, `0` and `auto` just for completness

src/docker/shared.rs
```diff
@@ -648,6 +657,50 @@ pub fn path_hash(path: &Path) -> Result<String> {
...
+        env::set_var(var, "auto");
+        test(EngineType::Docker, &rootful);
+        test(EngineType::Podman, &rootless);
+        test(EngineType::PodmanRemote, &rootless);
+        test(EngineType::Other, &rootless);
+
+        match old {
+            Ok(v) => env::set_var(var, v),
+            Err(_) => env::remove_var(var),
+        }
```
Emilgardis, line 702: vars are not shared in test context

Emilgardis (changes requested):

src/docker/shared.rs
Emilgardis: Sorry!! I think this is false actually. Dismiss this :)

Alexhuszagh: Whoops, Ok I'll re-apply this.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+- #890 - support rootless docker via the `CROSS_ROOTLESS_CONTAINER_ENGINE` environment variable.
```
Emilgardis, line 11: newline!

Emilgardis (approved):
lgtm

one thing I just realised is that we have `config::Environment`, we should probably convert all our special case stuff to be in there sometime. This is fine though

Alexhuszagh:
> one thing I just realised is that we have `config::Environment`, we should probably convert all our special case stuff to be in there sometime. This is fine though

That would also be good, since we could move `bool_from_envvar` and other logic there as well, and then provide `bool_or_default` or something. Something like:

```rust
fn bool_or_default(var: &str) -> bool {
    env::var(var)
        .map(|s| bool_from_envvar(&s))
        .unwrap_or_default()
}
```

And this would cover basically every case.

Alexhuszagh:
bors r=Emilgardis

Issue #888, opened 2022-07-01
Support lima


messense:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar request
- [ ] This feature could be solved with a [custom docker image](https://github.com/cross-rs/cross#custom-docker-images) (optional)

### Describe your request

https://github.com/lima-vm/lima

Linux virtual machines, typically on macOS, for running containerd.

It can be used as a replacement of Docker Desktop for Mac.

### Describe why this would be a good inclusion for `cross`

`lima nerdctl` is mostly compatible with `docker`, so it can be used to replace `docker` cli.

The benefits for end user is that [`Lima is free software (Apache License 2.0), while Docker for Mac is not`](https://github.com/lima-vm/lima#advantages-compared-to-docker-for-mac).

Alexhuszagh:
Do you know if Lima runs rootless or not? We probably also need to detect rootless Docker as well, on a separate note. It seems like there might not be much work as long as the `nerdctl.lima` wrapper is used.

messense:
The default template runs rootless I think.

https://github.com/lima-vm/lima/blob/a250961e868c372a20ee4170f69d63e39359a51a/examples/default.yaml#L145

Alexhuszagh:
This might be an issue (see same for #756). I'm currently running on a x86_64 Linux host, so I'll have to test on macOS at well but I'm not very optimistic.

### Attempts

**Native Lima**

```bash
$ limactl start default
$ nerdctl.lima run -it --rm -v "$PWD":/project:Z -w /project ubuntu:20.04 bash                           
WARN[0006] unsupported volume option "Z" 
$ touch a
touch: cannot touch 'a': Read-only file system
```

**Podman Emulation**

```bash
$ limactl start --name=podman template://podman
$ podman system connection add lima "unix:///home/ahuszagh/.lima/podman/sock/podman.sock"
$ podman system connection default lima
$ podman --remote run quay.io/podman/hello   # this works
# works, using local podman. needs SELinux label for r/w functionality
$ podman run -it --rm -v "$PWD":/project:Z -w /project ubuntu:20.04 bash
# fails, and is read-only without the SELinux label
$ podman --remote run -it --rm -v "$PWD":/project:Z -w /project ubuntu:20.04 bash
[...]

PR #891, opened 2022-07-01
Add manual overrides for container user namespaces.

Alexhuszagh:
Support manual overrides of the container user namespace via the `CROSS_CONTAINER_USER_NAMESPACE` environment variable. If not set or set to `auto`, it will use the default value for the container engine (`host`). If `none` is provided, no `--userns` flag will be used. If any other value is provided, that will be the value passed to `--userns`.

This is required for using lima/nerdctl, which currently does not support the `--userns` flag.

Related to #888.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -183,14 +183,14 @@ pub(crate) fn register(engine: &Engine, target: &Target, msg_info: MessageInfo) 
             binfmt-support qemu-user-static"
     };
 
-    subcommand(engine, "run")
-        .args(&["--userns", "host"])
-        .arg("--privileged")
-        .arg("--rm")
-        .arg(UBUNTU_BASE)
-        .args(&["sh", "-c", cmd])
-        .run(msg_info, false)
-        .map_err(Into::into)
+    let mut docker = subcommand(engine, "run");
+    docker_userns(&mut docker);
+    docker.arg("--privileged");
+    docker.arg("--rm");
+    docker.arg(UBUNTU_BASE);
+    docker.args(&["sh", "-c", cmd]);
+
+    docker.run(msg_info, false).map_err(Into::into)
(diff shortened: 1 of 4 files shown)


src/docker/shared.rs
```diff
@@ -405,6 +405,20 @@ pub(crate) fn docker_user_id(docker: &mut Command, engine_type: EngineType) {
     }
 }
 
+pub(crate) fn docker_userns(docker: &mut Command, engine_type: EngineType) {
+    // nerdctl does not support user namespace remapping, but
+    // allow users to manually override the provided userns.
+    let is_nerdctl = engine_type == EngineType::Nerdctl;
+    let userns = match env::var("CROSS_CONTAINER_USER_NAMESPACE").ok().as_deref() {
+        None | Some("auto") if !is_nerdctl => Some("host".to_string()),
+        None | Some("auto") | Some("none") => None,
```
Emilgardis, line 414: 1.62 `bool::then_some` would work nicely here

Emilgardis:
I don't know if we should explicitly support this. Maybe instead it should be more that the ContainerEngine is custom (i.e any random binary)

Alexhuszagh:
> I don't know if we should explicitly support this. Maybe instead it should be more that the ContainerEngine is custom (i.e any random binary)

Maybe, we should probably at least support changing the user namespace, which would be necessary for lima support right now. But you're probably right: there's too many niche things here that can be covered in a more general case. Just keep the user namespace manual overrides, get rid of everything else?

Emilgardis:
> Just keep the user namespace manual overrides, get rid of everything else?

yep, sounds good to me

Emilgardis (approved):
lgtm

Alexhuszagh:
bors r=Emilgardis

PR #897, opened 2022-07-01
Fix build/target config file ordering

Alexhuszagh:
Fix the order of values for string (as well as the pre-build vec) values when parsing from config files and environment variables.

The old behavior was:
1. Envvar build.
2. Envvar target
3. Config target
4. Config build

The new behavior is:
1. Envvar target
2. Config target
3. Envvar build.
4. Config build

Some functions were also made more general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Emilgardis (approved):
Lgtm!

We should however probably "structify" it a bit so it's clearly signaled what is target and what is build

Issue #896, opened 2022-07-01
Update `build-std` to support a boolean or array.


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar request
- [ ] This feature could be solved with a [custom docker image](https://github.com/cross-rs/cross#custom-docker-images) (optional)

### Describe your request

When using `-Z build-std`, `cargo` also enables you to pass various different [flags](https://doc.rust-lang.org/cargo/reference/unstable.html#build-std) which will specify the exact components to build, rather than all components. This can be useful if `std` isn't supported, but `core` and `alloc` are. Therefore, `build-std` could be provided as an array or bool (we shouldn't allow a table), and then passed transparent to `-Z build-std`.

```toml
[target.x86_64-unknown-dragonfly]
build-std = true

[target.thumbv6m-none-eabi]
build-std = ["core", "alloc"]
```

### Describe why this would be a good inclusion for `cross`

Allows more comprehensive use of `build-std`, without breaking old behavior.

PR #899, opened 2022-07-02
Better error diagnostics on targets without std.

### LINKED ISSUE #94: Missing rust-std for targets
`x86_64-unknown-dragonfly`, `armv7-unknown-linux-musleabihf`, and `arm-unknown-linux-musleabihf` are affected. I think this is because `rustup` doesn't have support for installation on those platforms (see [here](https://github.com/rust-lang-nursery/rustup.rs/#other-installation-methods)).

### LINKED ISSUE #167: No error on unsupported target
If I choose a supported target (let's say `cross build --target=aarch64-unknown-linux-gnu`), everything works like charm ‒ it automatically downloads the std for that target, downloads the docker image and builds it.

However, if I choose an unsupported target (`aarch64-unknown-linux-musl`), it just tries to call cargo without any complaint. Then of course cargo complains:

```
error[E0463]: can't find crate for `core`
  |
  = note: the `aarch64-unknown-linux-musl` target may not be installed
```

I'd expect to see one of:
* A message that such target is not supported
* A failed attempt at installing the target's std or the docker image

Alexhuszagh:
Provides a warning if a docker command fails and `rust-std` is not available for the target. For example, for `x86_64-unknown-dragonfly`, the added output would be:

```bash
[cross] warning: rust-std is not available for x86_64-unknown-dragonfly
[cross] note: you may need to build components for the target via `-Z build-std=<components>` or in your cross configuration specify `target.x86_64-unknown-dragonfly.build-std`
              the available components are core, std, alloc, and proc_macro
```

This is done solely if the command fails and if the target is a built-in, so we don't get misleading warnings for custom targets, and there is no overhead if the build succeeds.

Closes #94.
Closes #167.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -360,6 +360,24 @@ impl Serialize for Target {
     }
 }
 
+fn warn_on_failure(target: &Target, toolchain: &str, msg_info: MessageInfo) -> Result<()> {
+    let rust_std = format!("rust-std-{target}");
+    if target.is_builtin() {
+        let component = rustup::check_component(&rust_std, toolchain, msg_info)?;
+        if component.is_not_available() {
+            shell::warn(format!("rust-std is not available for {target}"), msg_info)?;
+            shell::note(
+                format_args!(
+                    r#"you may need to build components for the target via `-Z build-std=<components>` or in your cross configuration specify `target.{target}.build-std`
+              the available components are core, std, alloc, and proc_macro"#
+                ),
+                msg_info,
+            )?;
+        }
+    }
+    Ok(())
+}
+
 pub fn run() -> Result<ExitStatus> {
     let target_list = rustc::target_list(Verbosity::Quiet.into())?;
     let args = cli::parse(&target_list)?;
@@ -524,6 +542,9 @@ pub fn run() -> Result<ExitStatus> {
                     .subcommand
                     .map(|sc| sc.needs_host(is_remote))
                     .unwrap_or(false);
+                if !status.success() {
+                    warn_on_failure(&target, &toolchain, args.msg_info)?;
+                }
                 if !(status.success() && needs_host) {
                     return Ok(status);
                 }
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -120,16 +120,52 @@ pub fn install_component(component: &str, toolchain: &str, msg_info: MessageInfo
         .wrap_err_with(|| format!("couldn't install the `{component}` component"))
 }
 
+pub enum Component<'a> {
+    Installed(&'a str),
+    Available(&'a str),
+    NotAvailable(&'a str),
+}
+
+impl<'a> Component<'a> {
+    pub fn is_installed(&'a self) -> bool {
+        matches!(self, Component::Installed(_))
+    }
+
+    pub fn is_not_available(&'a self) -> bool {
+        matches!(self, Component::NotAvailable(_))
+    }
+}
+
+pub fn check_component<'a>(
+    component: &'a str,
+    toolchain: &str,
+    msg_info: MessageInfo,
+) -> Result<Component<'a>> {
+    Ok(Command::new("rustup")
+        .args(&["component", "list", "--toolchain", toolchain])
+        .run_and_get_stdout(msg_info)?
+        .lines()
+        .find_map(|line| {
+            let available = line.starts_with(component);
+            let installed = line.contains("installed");
+            match available {
+                true => Some(installed),
+                false => None,
+            }
+        })
+        .map(|installed| match installed {
+            true => Component::Installed(component),
+            false => Component::Available(component),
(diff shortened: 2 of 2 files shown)


Alexhuszagh:
An example of the new error diagnostics is as follows. We can't remove the `cargo` output without basically recreating it, and trying to tee/capture it would run into issues with terminal coloring, etc. However, the resulting messages are quite nice:

rust_std

If we have an unknown target, we get:

```bash
$ cross build --target x86_64-unknown-dragonfly1
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --target x86_64-unknown-dragonfly1 --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
  --- stderr
  error: Error loading target specification: Could not find specification for target "x86_64-unknown-dragonfly1". Run `rustc --print target-list` for a list of built-in targets
```

And if the component is installed but does not support the standard library, we get:

```bash
$ cross build --target thumbv6m-none-eabi

   Compiling hello v0.1.0 (/project)
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library
  = note: `std` is required by `hello` because it does not declare `#![no_std]`

error: cannot find macro `println` in this scope
 --> src/main.rs:2:5
  |
2 |     println!("Hello, world!");
  |     ^^^^^^^

error: `#[panic_handler]` function required, but not found

For more information about this error, try `rustc --explain E0463`.
error: could not compile `hello` due to 3 previous errors
```

Emilgardis:
I'm not sure this is the correct approach, specifically the wording. If I don't require libstd this message would still always be shown.

Alexhuszagh:
We could specific to use `-Z build-std=<components>` (which we could change once we've implemented #896), so we can make it clear that libstd (it might be core, alloc, etc.) isn't required but may be an issue?

Emilgardis:
What about making component_is_installed returning an enum (or `Option<bool>`), signaling a component is installed, available but not installed, and not available. Then if not available, signal to use build-std

Alexhuszagh:
I've updated the error messages to make it clear `std` itself is not needed, and how `build-std` may be used.
Screenshot_20220702_122407

src/lib.rs
```diff
@@ -360,6 +360,23 @@ impl Serialize for Target {
...
+    if target.is_builtin() {
+        let component = rustup::check_component(&rust_std, toolchain, msg_info)?;
+        if component.is_not_available() {
+            shell::warn(format!("rust-std is not available for {target}"), msg_info)?;
+            shell::note(
+                "you may need to build components for the std library `-Z build-std=<components>`
+              the available components are core, std, alloc, and proc_macro
+              if no components are provided, cargo will build all components",
+                msg_info,
+            )?;
```
Emilgardis, line 374:
```suggestion
            shell::note(
                format_args("you may need to build components for the target via `-Z build-std=<components>` or in your cross configuration specify `target.{target}.build-std`
              the available components are core, std, alloc, and proc_macro"),
                msg_info,
            )?;
```

Emilgardis: don't think the last part is needed, first reading it I was confused as to what it was referring to.

Emilgardis (changes requested):

src/rustup.rs
```diff
@@ -120,16 +120,53 @@ pub fn install_component(component: &str, toolchain: &str, msg_info: MessageInfo
...
+        .map(ToOwned::to_owned)
+        .collect();
+    Ok(
+        if components
+            .iter()
+            .any(|c| c.starts_with(component) && c.contains("installed"))
+        {
+            Component::Installed(component)
+        } else if components.iter().any(|c| c.starts_with(component)) {
+            Component::Available(component)
```
Emilgardis, line 157: what about using find_map instead of looping over twice

Alexhuszagh: Changed to:

```rust
pub fn check_component<'a>(
    component: &'a str,
    toolchain: &str,
    msg_info: MessageInfo,
) -> Result<Component<'a>> {
    Ok(Command::new("rustup")
        .args(&["component", "list", "--toolchain", toolchain])
        .run_and_get_stdout(msg_info)?
        .lines()
        .find_map(|line| {
            let available = line.starts_with(component);
            let installed = line.contains("installed");
            match available {
                true => Some(installed),
                false => None,
            }
        })
        .map(|installed| match installed {
            true => Component::Installed(component),
            false => Component::Available(component),
        })
        .unwrap_or_else(|| Component::NotAvailable(component)))
}
```

Only returning `installed` works since we know available is true.

Emilgardis (approved):
bors r+

PR #900, opened 2022-07-02
Enable skipping of copying build artifacts back to host.

Alexhuszagh:
If using remote docker but primarily wishing to build or test your builds remotely, without caring about the generated artifacts, the time required to copy the generated artifacts over a network connection can be prohibitive.

Setting `CROSS_REMOTE_SKIP_BUILD_ARTIFACTS=1` will disable copying these build artifacts back to the host, for faster build times with remote container engines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


Emilgardis (approved):

Emilgardis (approved):
bors r+

PR #905, opened 2022-07-03
Numerous bug fixes for musl and other images.

Alexhuszagh:
Provide `qemu-runner` that allows native or non-native for any architecture, and install `qemu-user` binaries for x86 musl images. In addition, create symlinks so dynamically-linked binaries can be run, including for C++, with musl images.

Adding `qemu-user` runners is also useful for running dynamically-linked x86 binaries, since otherwise they may be unable to find the linked musl libc, and we cannot install it in `/lib` or `/usr/lib` due to conflicts with the system.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
             - { target: aarch64-unknown-linux-gnu,        os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1, runners: qemu-user qemu-system }
             - { target: arm-unknown-linux-gnueabi,        os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1 }
             - { target: arm-unknown-linux-gnueabihf,      os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1 }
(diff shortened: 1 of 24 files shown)


Alexhuszagh:
This is a very preliminary draft (will need a CHANGELOG entry once it's more established), but it aims to do the following:
- Allow musl images to run dynamically-linked binaries
- Add qemu-user and native runners for all musl images (very handy for x86)
- Provide a single script for symlinking all musl libc locations, to ensure it is found.

Alexhuszagh:
bors try

Alexhuszagh:
I've fixed the bug with CI trying to use `qemu-system` (It had previously hard-coded `qemu-arm`, so `CROSS_RUNNER` was being ignored) for `armv7-unknown-linux-gnueabi`, so this should be ready for review. All targets pass (well, `armv7-unknown-linux-gnueabi` should pass).

Emilgardis (approved):
looks good!

Would it make sense to actually have `MUSL_SYSROOT` as an `ENV` instead of simply a build arg?

Alexhuszagh:
> looks good!
> 
> Would it make sense to actually have `MUSL_SYSROOT` as an `ENV` instead of simply a build arg?

If so we should probably prefix it as `CROSS_MUSL_SYSROOT`, but absolutely.

Emilgardis (approved):
bors r+
🚀

PR #907, opened 2022-07-03
wiki check toml

Emilgardis:


Alexhuszagh:
This TOML check is currently failing on snippets like:

```toml
[target.$TARGET]
image = "$REPO/$TARGET:$TAG"
```

To be fair, this isn't valid TOML, so should handle it like [cross_toml.md](https://github.com/cross-rs/cross/blob/main/docs/cross_toml.md) for how we handle variables? Like `target.TARGET`?

src/tests/toml.rs
```diff
@@ -35,16 +37,40 @@ fn toml_check() -> Result<(), Box<dyn std::error::Error>> {
...
+                } else {
+                    println!("skipping {t}");
+                    continue;
+                }
+            };
+            let fence = matches.get(2).unwrap();
+            let fence_content = fence
+                .as_str()
+                .replace("$TARGET", "x86_64-unknown-linux-gnu")
+                .replace("${target}", "x86_64-unknown-linux-gnu");
```
Emilgardis, line 55: @Alexhuszagh does this not cover that? Runnign this on my machine this passes

Alexhuszagh: Oh right. I forgot to checkout the PR, because I did and forgot and had unstaged changes.

Alexhuszagh (approved):

PR #909, opened 2022-07-04
Allow verbose output in ct-ng images.

Alexhuszagh:
Defaulting to verbose output pollutes the build logs during CI, which can be obscure any errors or status results. However, silencing output on local builds can make it difficult to debug issues. This checks for verbose output and the build script is running on a TTY, so we only get spinner output in those cases (IE, where the spinners actually work and are able to delete the current output line).

### DIFF
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -24,8 +24,8 @@ pub struct BuildDockerImage {
     #[clap(long, env = "LABELS")]
     pub labels: Option<String>,
     /// Provide verbose diagnostic output.
-    #[clap(short, long)]
-    pub verbose: bool,
+    #[clap(short, long, action = clap::ArgAction::Count)]
+    pub verbose: u8,
     /// Do not print cross log messages.
     #[clap(short, long)]
     pub quiet: bool,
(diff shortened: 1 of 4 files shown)


xtask/src/build_docker_image.rs
```diff
@@ -236,6 +236,9 @@ pub fn build_docker_image(
         for arg in &build_arg {
             docker_build.args(&["--build-arg", arg]);
         }
+        if let Ok(env_verbose) = std::env::var("VERBOSE") {
+            docker_build.args(&["--build-arg", &format!("VERBOSE={env_verbose}")]);
+        }
```
Emilgardis, line 241: what about making verbose an `u8`, and if verbosity > 1 then set build-arg `VERBOSE`

```
#[clap(action = clap::ArgAction::Count)]
verbose: u8
```

Emilgardis (commented):
what a

xtask/src/build_docker_image.rs
Alexhuszagh: Sure that works, maybe we should have something like `-vv` or similar then instead?

Emilgardis: isn't that what it enables?

Emilgardis (approved):
lgtm!

bors r+

Issue #861, opened 2022-06-25
Implement pre-build copy


Emilgardis:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar request
- [X] This feature could be solved with a [custom docker image](https://github.com/cross-rs/cross#custom-docker-images) (optional)

### Describe your request

From [reddit](https://www.reddit.com/r/rust/comments/vk2xfc/cross_v022_released/idn2x8e/)

Make it easy to integrate scripts to the pre-build hook.

Alexhuszagh:
I'm not sure this is needed: I remember being initially very favorable of the idea of supporting other commands in the pre-build hooks or customizing the base image, but in retrospect it seems like the solution we implemented was much better.

Currently, we have:

```toml
[target.x86_64-unknown-linux-gnu]
pre-build = [
    "dpkg --add-architecture arm64", 
    "apt-get update && apt-get install --assume-yes libssl-dev:arm64"
]
```

And my initial proposal would have been:

```toml
[target.x86_64-unknown-linux-gnu]
pre-build = [
    "RUN dpkg --add-architecture arm64", 
    "RUN apt-get update && apt-get install --assume-yes libssl-dev:arm64"
]
```

Which honestly in retrospect is not very legible, and a Dockerfile is better in that case. If we use a `copy`, will all copies be done before all runs? How will this affect cached builds if another step is added? What else do we need to implement? `ENV`? `ARG`? I think just using a `Dockerfile` or simple pre-build `RUN` instructions is enough.

Emilgardis:
We only have one `RUN` inserted, each value in `pre-build` is separated by a newline and evaluated with `eval`

I think it'd be fine to include a way to call a script specifically, maybe that usecase is what should be supported and not a general "copy"

Alexhuszagh:
> We only have one `RUN` inserted, each value in `pre-build` is separated by a newline and evaluated with `eval`
> 
> I think it'd be fine to include a way to call a script specifically, maybe that usecase is what should be supported and not a general "copy"

I like that better, and make it clear it would happen after the pre-build run commands, which I believe is best.

Emilgardis:
maybe simply

```toml
[*]
pre-build = "./myscript.sh" # Note: not an array.
```
don't think we need to support pre-build via array and this new alternative

Alexhuszagh:
That makes it super simple, and correct. I think that's the correct approach.

[...]

PR #910, opened 2022-07-04
add way to run script with pre-build

Emilgardis:
Solves #861

### DIFF
--- a/docs/cross_toml.md
+++ b/docs/cross_toml.md
@@ -34,8 +34,26 @@ The `target` key allows you to specify parameters for specific compilation targe
 xargo = false
 build-std = false
 image = "test-image"
-pre-build = ["apt-get update"]
+pre-build = ["apt-get update"] # can also be the path to a file to run
 runner = "custom-runner"
+```
+
+# `target.TARGET.pre-build`
+
+The `pre-build` field can also reference a file to copy and run. This file is relative to the container context, which would be the workspace root, or the current directory if `--manifest-path` is used. For more involved scripts, consider using `target.TARGET.dockerfile` instead to directly control the execution.
+
+This script will be invoked as `RUN ./pre-build-script $CROSS_TARGET` where `$CROSS_TARGET` is the target triple.
+
+```toml
+[target.aarch64-unknown-linux-gnu]
+pre-build = "./scripts/my-script.sh"
+```
+
+```sh
+$ cat ./scripts/my-script.sh
+#!/usr/bin/env bash
+
+apt-get install libssl-dev -y
 ```
 
 # `target.TARGET.env`
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -1,5 +1,6 @@
 use std::io::Write;
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use crate::docker::{DockerOptions, DockerPaths};
 use crate::shell::MessageInfo;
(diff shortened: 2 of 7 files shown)


src/docker/shared.rs
```diff
@@ -498,29 +494,65 @@ pub(crate) fn custom_image_build(
...
+            {
+                let custom = Dockerfile::Custom {
+                    content: format!(
+                        r#"
+            FROM {image}
+            ARG CROSS_DEB_ARCH=
+            ARG CROSS_SCRIPT
+            COPY $CROSS_SCRIPT /script
+            RUN chmod +x /script
+            RUN ./script "#
```
Emilgardis, line 510: I'm not sure this is the correct way to do this, should maybe keep the filename.

I think using chmod is good though, even if it's not strictly needed

src/docker/shared.rs
```diff
@@ -498,29 +494,65 @@ pub(crate) fn custom_image_build(
...
-                    metadata,
-                    engine,
-                    &host_root,
-                    Some(("CROSS_CMD", pre_build.join("\n"))),
-                    target,
-                    msg_info,
-                )
-                .wrap_err("when pre-building")
-                .with_note(|| format!("CROSS_CMD={}", pre_build.join("\n")))?;
-            image = custom.image_name(target, metadata)?;
```
Emilgardis, line 523: Fixed this assumption, which is wrong in a more general case. It works by coincidence

src/docker/shared.rs
Alexhuszagh: It will probably help with me having to avoid pushing commits on Windows or commit, then change permissions on Linux and recommit, so I like the change.

src/docker/custom.rs
```diff
@@ -22,6 +23,40 @@ pub enum Dockerfile<'a> {
     },
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub enum PreBuild {
```
Alexhuszagh, line 27: Should we allow `pre-build = "command" to exist? Or should that only be reserved for scripts? Say someone has a `dpkg` in the Docker context: shouldn't we avoid having conflicts.

docs/cross_toml.md
```diff
@@ -34,10 +34,26 @@ The `target` key allows you to specify parameters for specific compilation targe
 xargo = false
 build-std = false
 image = "test-image"
-pre-build = ["apt-get update"]
+pre-build = ["apt-get update"] # can also be the path to a file to run
 runner = "custom-runner"
 ```
 
+# `target.TARGET.pre-build`
+
```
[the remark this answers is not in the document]
Emilgardis: Right now it's on the host_root, defined as cargo manifest path or cwd if outside workspace. This is rather strange imo, but not sure how to make it as usable as possible.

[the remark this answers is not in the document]
Alexhuszagh: Yeah, I think it's important to document because this is the first situation where people without experience with Dockerfiles might need to know what a context is.

src/docker/custom.rs
Alexhuszagh: I think it's a good idea just to avoid ambiguity. It will make serializing it a bit more difficult, however.

Alexhuszagh:
I'm guessing that how it serializes/deserializes to JSON is irrelevant, since it will never be JSON userfacing, so serializing to/from an object is fine.

Alexhuszagh (approved):

PR #911, opened 2022-07-05
Structify more of the internals to simplify the implementation.

Alexhuszagh:
Reduces the number of arguments to the function signatures, making the code much easier to understand, and relocates some things. For example, `docker_in_docker` is now associated with `Engine`, and now with `cli::Args`. Likewise, unstable doctests is now a part of `Environment` and not `cli::Args`. `MessageInfo` now stores mutable state (needs to clear lines) and is no longer copyable, and a single instance is created in lib and passed around everywhere as a mutable reference.

The changes with `MessageInfo` will allow us to erase lines in the future, both from `stderr` and `stdout`, which we currently don't use but will allow us to close #859.

 This also fixes an issue where `-vv+` was not recognized as a valid verbose flag: `cargo` can handle 1 or more `v` characters. Also allowed parsing of `--color=value` args, when we previously only accepted `--color value`.

### DIFF
--- a/src/bin/commands/clean.rs
+++ b/src/bin/commands/clean.rs
@@ -3,7 +3,7 @@ use std::fs;
 use super::containers::*;
 use super::images::*;
 use clap::Args;
-use cross::shell::{self, MessageInfo};
+use cross::shell::MessageInfo;
 
 #[derive(Args, Debug)]
 pub struct Clean {
@@ -32,7 +32,7 @@ pub struct Clean {
 
 impl Clean {
     pub fn run(self, engine: cross::docker::Engine) -> cross::Result<()> {
-        let msg_info = MessageInfo::create(self.verbose, self.quiet, self.color.as_deref())?;
+        let mut msg_info = MessageInfo::create(self.verbose, self.quiet, self.color.as_deref())?;
         let tempdir = cross::temp::dir()?;
         match self.execute {
             true => {
@@ -40,13 +40,10 @@ impl Clean {
                     fs::remove_dir_all(tempdir)?;
                 }
             }
-            false => shell::print(
-                format!(
-                    "fs::remove_dir_all({})",
-                    cross::pretty_path(&tempdir, |_| false)
-                ),
-                msg_info,
-            )?,
+            false => msg_info.print(format_args!(
+                "fs::remove_dir_all({})",
+                cross::pretty_path(&tempdir, |_| false)
+            ))?,
         }
 
         // containers -> images -> volumes -> prune to ensure no conflicts.
@@ -92,5 +89,21 @@ impl Clean {
         prune_volumes.run(engine)?;
 
         Ok(())
+    }
+
+    pub fn engine(&self) -> Option<&str> {
+        self.engine.as_deref()
+    }
+
+    pub fn verbose(&self) -> bool {
+        self.verbose
+    }
+
+    pub fn quiet(&self) -> bool {
+        self.quiet
+    }
+
+    pub fn color(&self) -> Option<&str> {
+        self.color.as_deref()
     }
 }
--- a/src/bin/commands/clean.rs
+++ b/src/bin/commands/clean.rs
@@ -31,8 +31,11 @@ pub struct Clean {
 }
 
 impl Clean {
-    pub fn run(self, engine: cross::docker::Engine) -> cross::Result<()> {
-        let mut msg_info = MessageInfo::create(self.verbose, self.quiet, self.color.as_deref())?;
+    pub fn run(
+        self,
+        engine: cross::docker::Engine,
+        msg_info: &mut MessageInfo,
+    ) -> cross::Result<()> {
         let tempdir = cross::temp::dir()?;
         match self.execute {
             true => {
@@ -55,7 +58,7 @@ impl Clean {
             execute: self.execute,
             engine: None,
         };
-        remove_containers.run(engine.clone())?;
+        remove_containers.run(engine.clone(), msg_info)?;
 
         let remove_images = RemoveImages {
             targets: vec![],
@@ -67,7 +70,7 @@ impl Clean {
             execute: self.execute,
             engine: None,
         };
-        remove_images.run(engine.clone())?;
+        remove_images.run(engine.clone(), msg_info)?;
 
         let remove_volumes = RemoveAllVolumes {
             verbose: self.verbose,
@@ -77,7 +80,7 @@ impl Clean {
             execute: self.execute,
             engine: None,
         };
-        remove_volumes.run(engine.clone())?;
+        remove_volumes.run(engine.clone(), msg_info)?;
 
         let prune_volumes = PruneVolumes {
             verbose: self.verbose,
@@ -86,7 +89,7 @@ impl Clean {
             execute: self.execute,
             engine: None,
         };
-        prune_volumes.run(engine)?;
+        prune_volumes.run(engine, msg_info)?;
 
         Ok(())
     }
--- a/src/bin/commands/containers.rs
+++ b/src/bin/commands/containers.rs
@@ -1,7 +1,7 @@
 use std::io;
 
 use clap::{Args, Subcommand};
-use cross::shell::{self, MessageInfo, Stream};
+use cross::shell::{MessageInfo, Stream};
 use cross::{docker, CommandExt};
 
 #[derive(Args, Debug)]
@@ -149,6 +149,18 @@ pub enum Volumes {
     Remove(RemoveVolume),
 }
 
+macro_rules! volumes_get_field {
+    ($self:ident, $field:ident $(.$cb:ident)?) => {{
+        match $self {
+            Volumes::List(l) => l.$field$(.$cb())?,
+            Volumes::RemoveAll(l) => l.$field$(.$cb())?,
+            Volumes::Prune(l) => l.$field$(.$cb())?,
+            Volumes::Create(l) => l.$field$(.$cb())?,
+            Volumes::Remove(l) => l.$field$(.$cb())?,
+        }
+    }};
+}
+
 impl Volumes {
     pub fn run(self, engine: docker::Engine, toolchain: Option<&str>) -> cross::Result<()> {
         match self {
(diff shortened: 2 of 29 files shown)


Alexhuszagh:
Still a draft since many things changed and I need to triple check no bugs were introduced. This refactors a lot of things to try to simplify our logic internally. The only actual changes in terms of behavior for the user are bug fixes:

1. `--color=value` is now valid
2. `-vvvvvvvvvvvv` or as many arguments of `v` are now valid

Both are consistent with `cargo`.

src/bin/commands/clean.rs
```diff
@@ -93,4 +90,20 @@ impl Clean {
 
         Ok(())
     }
+
+    pub fn engine(&self) -> Option<&str> {
```
Alexhuszagh, line 94: I can get rid of most of these getters and just replace them with the raw fields now that we use a macro to extract the fields.

src/bin/commands/containers.rs
```diff
@@ -149,6 +149,18 @@ pub enum Volumes {
     Remove(RemoveVolume),
 }
 
+macro_rules! volumes_get_field {
+    ($self:ident, $field:ident $(.$cb:ident)?) => {{
```
Alexhuszagh, line 153: To avoid a lot of redundant code, I added macros so we can extract a single field from an enumeration. The callback allows to add `as_deref` if needed.

src/bin/commands/containers.rs
```diff
@@ -319,9 +318,9 @@ pub fn list_volumes(
     }: ListVolumes,
     engine: &docker::Engine,
 ) -> cross::Result<()> {
-    let msg_info = MessageInfo::create(verbose, quiet, color.as_deref())?;
-    for line in get_cross_volumes(engine, msg_info)?.iter() {
-        shell::print(line, msg_info)?;
+    let mut msg_info = MessageInfo::create(verbose, quiet, color.as_deref())?;
```
Alexhuszagh, line 321: This needs to be moved outwards: `msg_info` should be able to be passed to `get_engine`.

src/shell.rs
```diff
@@ -36,219 +91,205 @@ pub enum ColorChoice {
...
 
-impl Default for MessageInfo {
-    fn default() -> MessageInfo {
-        MessageInfo {
-            color_choice: ColorChoice::Auto,
-            verbosity: Verbosity::Normal,
-        }
-    }
-}
+    fn as_verbosity<T, C: Fn(&mut MessageInfo) -> T>(&mut self, call: C, new: Verbosity) -> T {
```
Alexhuszagh, line 128: I've added these functions so we can temporarily change the verbosity of a function call, either to silence or to be more verbose, etc. It allows us to store the old value, and return a result and ensure the old state is restored.

src/shell.rs
```diff
@@ -36,219 +91,205 @@ pub enum ColorChoice {
...
         Ok(())
-    }};
+    }
 
-    (@status @name $name:ident, $status:expr, $message:expr, $color:ident, $msg_info:expr $(,)?) => {{
-        let mut stream = io::$name();
-        message!(@status stream, $status, $message, $color, $msg_info)
-    }};
-}
+    /// prints a red 'error' message and terminates.
```
Alexhuszagh, line 170: I've also moved away from free functions since we store state now.

src/bin/cross-util.rs
```diff
@@ -49,39 +49,56 @@ fn is_toolchain(toolchain: &str) -> cross::Result<String> {
...
+    let in_docker = match docker_in_docker {
+        true => Some(true),
+        false => None,
+    };
+    docker::Engine::from_path(engine, in_docker, None, msg_info)
+}
+
+macro_rules! get_engine {
+    ($args:ident, $docker_in_docker:expr, $msg_info: ident) => {{
+        get_container_engine($args.engine(), $docker_in_docker, &mut $msg_info)
```
Alexhuszagh, line 69: These macros just simplify a lot of redundant code that existed earlier, while being relatively transparent.

src/cli.rs
```diff
@@ -52,23 +51,86 @@ pub fn group_subcommands(stdout: &str) -> (Vec<&str>, Vec<&str>) {
...
+        msg_info.print("Host Commands:")?;
         for line in cross.iter() {
-            shell::print(line, msg_info)?;
+            msg_info.print(line)?;
         }
     }
     Ok(())
 }
 
+fn is_verbose(arg: &str) -> bool {
```
Alexhuszagh, line 71: This corrects the behavior so `-vvvvv` and similar are accepted as arguments to get the verbosity. Previously this was passed to `cargo`, but not correctly processed for `cross` itself.

src/cli.rs
```diff
@@ -167,37 +223,17 @@ pub fn parse(target_list: &TargetList) -> Result<Args> {
         }
     }
 
-    let msg_info = shell::MessageInfo::create(verbose, quiet, color.as_deref())?;
-    let docker_in_docker = if let Ok(value) = env::var("CROSS_CONTAINER_IN_CONTAINER") {
```
Alexhuszagh, line 171: Not a CLI argument, so it's been moved to `Engine`.

src/cli.rs
```diff
@@ -167,37 +223,17 @@ pub fn parse(target_list: &TargetList) -> Result<Args> {
...
-        }
-        bool_from_envvar(&value)
-    } else if let Ok(value) = env::var("CROSS_DOCKER_IN_DOCKER") {
-        // FIXME: remove this when we deprecate CROSS_DOCKER_IN_DOCKER.
-        bool_from_envvar(&value)
-    } else {
-        false
-    };
-
-    let enable_doctests = env::var("CROSS_UNSTABLE_ENABLE_DOCTESTS")
```
Alexhuszagh, line 186: Not a CLI argument, so this has been moved to `config`.

src/config.rs
```diff
@@ -99,6 +99,16 @@ impl Environment {
...
             .or_else(|| std::env::var("CARGO_BUILD_TARGET").ok())
     }
+
+    fn doctests(&self) -> Option<bool> {
+        env::var("CROSS_UNSTABLE_ENABLE_DOCTESTS")
+            .map(|s| bool_from_envvar(&s))
+            .ok()
+    }
+
+    fn custom_toolchain(&self) -> bool {
```
Alexhuszagh, line 109: I think it's better we don't process all our config options for environment variables in `lib` itself, since variables like this are part of the configuration.

src/cli.rs
Emilgardis: Nice, was meaning to fix this myself

Emilgardis:
Great! looks like a straightforward interface improvement

src/docker/local.rs
```diff
@@ -1,56 +1,46 @@
...
         |_| {},
     )?;
 
     docker.arg("--rm");
 
-    docker_seccomp(&mut docker, engine.kind, target, metadata)
-        .wrap_err("when copying seccomp profile")?;
-    docker_user_id(&mut docker, engine.kind);
+    docker_seccomp(
+        &mut docker,
```
Alexhuszagh, line 37: I've tried not to overly rely on just having the structs passed: if there's any question of if the logic could need to be accessed outside and the logic does not rely on most of the members or it would require a lot of work to create those members, I've avoided passing the struct itself.

src/docker/remote.rs
```diff
@@ -9,35 +9,40 @@ use eyre::Context;
...
 use crate::temp;
 use crate::{Host, Target};
 
 // the mount directory for the data volume.
 pub const MOUNT_PREFIX: &str = "/cross";
 
-struct DeleteVolume<'a>(&'a Engine, &'a VolumeId, MessageInfo);
+// it's unlikely that we ever need to erase a line in the destructors,
+// and it's better than keep global state everywhere, or keeping a ref
+// cell which could have already deleted a line
```
Alexhuszagh, line 27: The deleters are the only location we don't have a single instance of `MessageInfo`, but I think this is fine since it avoids having to pass a `RefCell` everywhere, worrying about avoiding multiple mutable borrows, etc., and the chance we delete a line when invoking the deleters is near 0.

src/docker/shared.rs
```diff
@@ -28,6 +28,217 @@ const DOCKER_IMAGES: &[&str] = &include!(concat!(env!("OUT_DIR"), "/docker-image
...
+    pub fn new(engine: Engine, target: Target, config: Config, uses_xargo: bool) -> DockerOptions {
+        DockerOptions {
+            engine,
+            target,
+            config,
+            uses_xargo,
+        }
+    }
+
+    pub fn in_docker(&self) -> bool {
```
Alexhuszagh, line 49: These getters should probably be removed since we always pass by a non-mutable reference and these structs are only used internally, so it's entirely redundant.

src/extensions.rs
```diff
@@ -3,55 +3,58 @@ use std::fmt;
...
 pub const STRIPPED_BINS: &[&str] = &[crate::docker::DOCKER, crate::docker::PODMAN, "cargo"];
 
 pub trait CommandExt {
-    fn fmt_message(&self, msg_info: MessageInfo) -> String;
+    fn fmt_message(&self, msg_info: &mut MessageInfo) -> String;
 
-    fn print(&self, msg_info: MessageInfo) -> Result<()> {
-        shell::print(&self.fmt_message(msg_info), msg_info)
+    fn print(&self, msg_info: &mut MessageInfo) -> Result<()> {
+        let msg = self.fmt_message(msg_info);
```
Alexhuszagh, line 14: This needs to be in 2 steps to avoid multiple mutable borrows.

src/lib.rs
```diff
@@ -338,7 +338,7 @@ impl From<Host> for Target {
             Host::Aarch64UnknownLinuxMusl => Target::new_built_in("aarch64-unknown-linux-musl"),
             Host::Other(s) => Target::from(
                 s.as_str(),
-                &rustc::target_list(Verbosity::Quiet.into()).unwrap(),
+                &rustc::target_list(&mut Verbosity::Quiet.into()).unwrap(),
```
Alexhuszagh, line 341: Once again, we will never need to delete a line from the output here, so this isn't an issue.

PR #916, opened 2022-07-06
Only check staged content in pre-hooks

Emilgardis:


xtask/src/pre-commit.sh
```diff
@@ -1,2 +1,16 @@
...
+
+cleanup() {
+    git apply --allow-empty --whitespace=nowarn < "$tf" && git stash drop -q
+    rm "$tf"
+}
+
+git diff --full-index --binary > "$tf"
+git stash -q --keep-index
+
+trap cleanup EXIT
```
Alexhuszagh, line 14: This won't work on SIGKILL or SIGSTOP, which would lead to data lost. It also doesn't work for new files, which would mean they would be tests but not staged.

Alexhuszagh: A simple example is:

```bash
$ echo "Sample contents" > file
$ git diff --full-index --binary # no output
$ git stash -q --keep-index # doesn't stash anything
```

PR #921, opened 2022-07-07
Support cargo term variables.

### LINKED ISSUE #919: Support `CARGO_TERM_VERBOSE`, etc.
### Checklist

### Describe your request

Currently, when extracting the verbose mode, quiet mode, or term color, we ignore the environment variables `CARGO_TERM_COLOR`, etc. These should be respected and parsed by `cross` itself.

This is relatively easy to implement and is fully compatible with `cargo`.

Alexhuszagh:
Use the `CARGO_TERM_VERBOSE`, `CARGO_TERM_QUIET`, and `CARGO_TERM_COLOR` environment variables for cross terminal output. This enables us to detect term settings via environment variables, which are overridden by command-line flags, which is our own behavior.

Cargo has CLI flags override environment variables, and verbose and quiet cannot be set for the same level. So if `CARGO_TERM_VERBOSE=true` and `--quiet` are used, then we use the quiet setting, but if `--verbose` and `--quiet` are used, then we print a fatal error message.

Closes #919.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #921 - use `CARGO_TERM_VERBOSE`, `CARGO_TERM_QUIET`, and `CARGO_TERM_COLOR` environment variables for cross terminal output.
 - #913 - added the `x86_64-unknown-illumos` target.
 - #905 - added `qemu-runner` for musl images, allowing use of native or emulated runners.
 - #905 - added qemu emulation to `i586-unknown-linux-gnu`, `i686-unknown-linux-musl`, and `i586-unknown-linux-gnu`, so they can run on an `x86` CPU, rather than an `x86_64` CPU.
--- a/src/bin/commands/clean.rs
+++ b/src/bin/commands/clean.rs
@@ -13,7 +13,7 @@ pub struct Clean {
     /// Do not print cross log messages.
     #[clap(short, long)]
     pub quiet: bool,
-    /// Whether messages should use color output.
+    /// Coloring: auto, always, never
     #[clap(long)]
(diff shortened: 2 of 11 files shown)


Emilgardis (commented):
lgtm!

the clap `Option<String>` should be converted to `ColorChoice` though, with default auto

Alexhuszagh:
Should I change everything to an enumerated values? I had it a list of options with a default earlier, but removed it because of envvars, but I can restore it.

https://github.com/clap-rs/clap/blob/v3.1.18/examples/tutorial_derive/README.md#enumerated-values

Emilgardis:
Was thinking this one: https://docs.rs/clap/3.2.8/clap/builder/struct.EnumValueParser.html

Alexhuszagh:
> Was thinking this one: https://docs.rs/clap/3.2.8/clap/builder/struct.EnumValueParser.html or https://docs.rs/clap/3.2.8/clap/trait.ValueEnum.html

That's the same I believe, just using the derive API.

Alexhuszagh:
It creates really bad help messages when we change those, so I've just made sure the error messages for invalid colors are nicer.

**Using Clap EnumParser**

Screenshot_20220707_195514

**Using Custom Validation**

Screenshot_20220707_195611

I even tried the following, but it doesn't do enough:

```rust
    /// Coloring: auto, always, never
    #[clap(
        long,
        arg_enum,
        //env = "CARGO_TERM_COLOR",
        hide_possible_values = true,
        hide_env_values = true,
        hide_default_value = true,
    )]
    pub color: ColorChoice,
```

I've made sure the output is similar to clap however, identical even on small terms.

**Generic Clap**

Screenshot_20220707_195535

**Cross Error**

Screenshot_20220707_204945

Emilgardis (approved):
bors r+

PR #922, opened 2022-07-08
Adds numerous bug fixes for remote cross.

Alexhuszagh:
Adds simpler container and volume cleanup, and avoiding keeping anonymous volumes for `/cargo/bin` present after the container exited. It also ensures containers are stopped even if a signal is sent to the `cross` command (except `SIGKILL`), and ensures that containers exit quicker if possible. This is done by using the presence of a TTY, and if it is present, running the default entrypoint, which responds to signals. If we do not have a TTY, the image will not respond to signals, so we send the container a `SIGKILL` via `docker stop` using a an instant timeout (0).

This also fixes an issue with stale fingerprint data: if a persistent data volume was deleted and then recreated, previously the project data would not be copied over. This meant that the project could not be found, and it would require users to manually delete the fingerprint data or use single-used volumes.

This also checks if the container has been scheduled for deletion (the drop may not have completed), and errors out early if more calls are made to the container.

This also properly handles `docker rm` calls if the `docker stop` command worked without needing a `SIGKILL`: it silences the output of the error if the container wasn't found.


Alexhuszagh:
There's a lot of comments but ~80% of them are comments, just to document clearly the behavior that's going on, the bugs we aim to prevent, etc.

Emilgardis (approved):
bors r+

nice!

PR #925, opened 2022-07-08
fix dind test

Emilgardis:
same as #924

also fixes no subs on variable passthrough


Alexhuszagh:
This probably also needs to be done in `test-remote` and `test-bisect`.

Emilgardis:
bisect and remote dont need it since it's using a binary with the data needed for cross to determine that it should use `main`

https://github.com/cross-rs/cross/runs/7256289776?check_suite_focus=true

Emilgardis:
bors r=@Alexhuszagh

PR #931, opened 2022-07-10
Add support for cargo aliases

Alexhuszagh:


Alexhuszagh:
This still needs to support the cargo environment variables, which luckily shouldn't be too hard.

Alexhuszagh:
This now supports environment variables, and the design is similar to the cross configuration files (and much more logical), but avoids parsing variables except for those we know and lazily reads environment variables.

Alexhuszagh:
Temporarily marking as a draft since this might have undergone a bad rebase, and I don't have the time right now to verify that it's correct.

Alexhuszagh:
This will need substantial rebases and refactoring, which I'll try to do shortly, so I'll revert this to a draft in the meantime.

PR #817, opened 2022-06-18
Enable usage of other toolchains in images

Emilgardis:
Makes the following work

```toml
# Cross.toml
[build]
default-target = "x86_64-unknown-linux-musl"

[target."x86_64-unknown-linux-musl"]
image.name = "alpine:edge"
image.toolchain = ["x86_64-unknown-linux-musl"]
pre-build = ["apk add --no-cache gcc musl-dev"]
```

```rust
// src/main.rs
fn main() {
    println!("Hello, world!");
}
```

```
> cross run
[..snip]
    Finished dev [unoptimized + debuginfo] target(s) in 0.24s
     Running `/target/x86_64-unknown-linux-musl/debug/cross-minimal`
Hello, world!
```

and also 

```toml
[target.x86_64-unknown-linux-gnu]
pre-build = [
    "apt-get update && apt-get install -y libc6 g++-x86-64-linux-gnu libc6-dev-amd64-cross",
]

[target.x86_64-unknown-linux-gnu.env]
passthrough = [
    "CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=x86_64-linux-gnu-gcc",
    "CC_x86_64_unknown_linux_gnu=x86_64-linux-gnu-gcc",
    "CXX_x86_64_unknown_linux_gnu=x86_64-linux-gnu-g++",
]

[target.x86_64-unknown-linux-gnu.image]
name = "ubuntu:20.04"
toolchain = ["aarch64-unknown-linux-gnu"]
```

```
cross build --target x86_64-unknown-linux-gnu -v
+ cargo metadata --format-version 1 --filter-platform x86_64-unknown-linux-gnu
+ rustc --print sysroot
+ /usr/local/bin/docker
+ /usr/local/bin/docker version '-f "{{ .Server.Os }},{{ .Server.Arch }}"'
+ rustup --verbose toolchain list
+ rustup --verbose target list --toolchain stable-aarch64-unknown-linux-gnu
+ rustup --verbose component list --toolchain stable-aarch64-unknown-linux-gnu
[...]

PR #934, opened 2022-07-11
Add support for pre-build hooks with remote docker.

### LINKED ISSUE #928: Support Pre-Build Hooks Remotely or Document Why Not
Currently, we only support custom docker images/pre-build hooks when `cross` is used locally. If we can use `DOCKER_HOST` (or similar for podman) to allow remote builds (it would need to send the context to docker remotely, which could take a while), then we should support that. If we cannot, then this should be clearly documented.

Alexhuszagh:
Closes #928.

### DIFF
--- a/.changes/934.json
+++ b/.changes/934.json
@@ -1,0 +1,5 @@
+{
+    "type": "added",
+    "description": "add support for pre-build hooks with remote docker",
+    "issues": [928]
+}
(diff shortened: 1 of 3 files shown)


Alexhuszagh:
Tested on the [openssl-certs](https://github.com/cross-rs/wiki_assets/tree/main/Recipes/openssl-certs) sample package.

Emilgardis:
#817 touches on this quite a bit, but for now let's merge this and I can just overwrite these changes there

Emilgardis (approved):
bors r+

Issue #621, opened 2022-01-02
cross picking up wrong target options in cargos config.toml


oliverbestmann:
I've added a `[target.target.x86_64-unknown-linux-gnu]` section to my `$HOME/.cargo/config.toml`. If i am now using cross to compile for a different target, say `cross build --target aarch64-unknown-linux-gnu`, the build picks up the config for x86_64-unknown-linux-gnu.

In my case, i've configured
```
oliver@machine ~> /bin/cat ~/.cargo/config.toml
[target.x86_64-unknown-linux-gnu]
linker = "/usr/bin/clang"
rustflags = [
  "-C", "target-cpu=native",
  "-C", "link-arg=-fuse-ld=/home/oliver/.local/bin/mold"
]
```

And the build fails with
```
oliver@machine ~> cross build --target aarch64-unknown-linux-gnu --release
info: syncing channel updates for '1.56.0-x86_64-unknown-linux-gnu'
  1.56.0-x86_64-unknown-linux-gnu unchanged - rustc 1.56.0 (09c42c458 2021-10-18)
error: linker `/usr/bin/clang` not found
  |
  = note: No such file or directory (os error 2)
```

doesn't work with rust `1.57.0` either.
cross version `0.2.1`

Emilgardis:
We shouldn't currently use that at all, instead you'd have to place the config in `.cargo/config.toml`in the manifest root if you want it to go over to the `cross` builder

perhaps we should support this automatically, but currently to workaround it if you mount the .cargo folder inside the builder with https://github.com/rust-embedded/cross#mounting-volumes-into-the-build-environment to [any of the supported paths ](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure)

Emilgardis:
Note that that feature is currently unreleased, you'd have to install from git

`cargo install cross --git https://github.com/rust-embedded/cross`

oliverbestmann:
Thing is, I don't want cross to pick that file up at all. It currently contains only configuration related to my host machine and target. But when compiling with cross on x86_64 for target aarch64, it picks up the options for the x86_64 target

kdar:
Thank you so much for this issue, as this was driving me up the wall why my cross compiling was failing. I'm a little confused as to why it was happening though. I was compiling a simple program, and it would compile until I added "libc" as a dependency. The failure was:
```
Compiling num-integer v0.1.44
error: linker `/usr/bin/clang` not found
  |
[...]

PR #936, opened 2022-07-11
Allow users to ignore config files in the package.

Alexhuszagh:
Adds the `CROSS_IGNORE_CARGO_CONFIG` environment variable, which if set will mount an anoymous data volume for each `.cargo` subdirectory for the current directory and any parent directories up to the workspace root. If the build is called outside the workspace root or at the workspace root, only mount at the `$PWD/.cargo`.

Partially addresses #621.

### DIFF
--- a/docs/cross_toml.md
+++ b/docs/cross_toml.md
@@ -21,6 +21,7 @@ For example:
 
 ```toml
 [build.env]
+ignore-cargo-config = false
 volumes = ["VOL1_ARG", "VOL2_ARG"]
 passthrough = ["IMPORTANT_ENV_VARIABLES"]
 ```
--- a/src/config.rs
+++ b/src/config.rs
@@ -96,6 +96,10 @@ impl Environment {
         self.get_target_var(target, "RUNNER")
     }
 
+    fn ignore_cargo_config(&self, target: &Target) -> (Option<bool>, Option<bool>) {
+        self.get_values_for("ENV_IGNORE_CARGO_CONFIG", target, bool_from_envvar)
+    }
+
     fn passthrough(&self, target: &Target) -> (Option<Vec<String>>, Option<Vec<String>>) {
         self.get_values_for("ENV_PASSTHROUGH", target, split_to_cloned_by_ws)
     }
@@ -273,6 +277,14 @@ impl Config {
 
     pub fn build_std(&self, target: &Target) -> Option<bool> {
         self.bool_from_config(target, Environment::build_std, CrossToml::build_std)
+    }
+
+    pub fn ignore_cargo_config(&self, target: &Target) -> Option<bool> {
+        self.bool_from_config(
+            target,
+            Environment::ignore_cargo_config,
+            CrossToml::ignore_cargo_config,
+        )
     }
 
     pub fn image(&self, target: &Target) -> Result<Option<String>> {
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -9,14 +9,11 @@ use crate::cargo::{cargo_metadata_with_args, CargoMetadata};
 use crate::config::{bool_from_envvar, Config};
 use crate::errors::*;
 use crate::extensions::{CommandExt, SafeCommand};
-use crate::file::{self, write_file, ToUtf8};
+use crate::file::{self, write_file, PathExt, ToUtf8};
 use crate::id;
 use crate::rustc::{self, VersionMetaExt};
 use crate::shell::{MessageInfo, Verbosity};
 use crate::Target;
-
-#[cfg(target_os = "windows")]
-use crate::file::PathExt;
 
 pub use super::custom::CROSS_CUSTOM_DOCKERFILE_IMAGE_PREFIX;
 
@@ -37,15 +34,23 @@ pub struct DockerOptions {
     pub target: Target,
     pub config: Config,
     pub uses_xargo: bool,
+    pub ignore_cargo_config: bool,
 }
 
 impl DockerOptions {
-    pub fn new(engine: Engine, target: Target, config: Config, uses_xargo: bool) -> DockerOptions {
+    pub fn new(
+        engine: Engine,
+        target: Target,
+        config: Config,
+        uses_xargo: bool,
+        ignore_cargo_config: bool,
+    ) -> DockerOptions {
         DockerOptions {
             engine,
             target,
             config,
             uses_xargo,
+            ignore_cargo_config,
         }
     }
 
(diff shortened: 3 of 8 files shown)


src/config.rs
```diff
@@ -115,6 +115,12 @@ impl Environment {
             .ok()
     }
 
+    fn ignore_cargo_config(&self) -> Option<bool> {
+        env::var("CROSS_IGNORE_CARGO_CONFIG")
```
Emilgardis, line 119:
```suggestion
        self.get_var("CROSS_IGNORE_CARGO_CONFIG")
```

src/docker/shared.rs
```diff
@@ -496,19 +504,60 @@ pub(crate) fn docker_envvars(
...
+        let mut relpath = Path::new(cwd).strip_prefix(root_path).wrap_err_with(|| {
+            eyre::eyre!("cwd \"{cwd}\" must be a subdirectory of root \"{root}\"")
+        })?;
+
+        while let Some(parent) = relpath.parent() {
+            let path = root_path.join(parent);
+            docker.args(&["-v", &path.join(".cargo").as_posix()?]);
+            relpath = parent;
+        }
+    }
```
Emilgardis, line 532: doesnt this mount more than it needs?

Alexhuszagh: There's still some permission issues here I need to fix, however.

src/config.rs
Alexhuszagh: Should we support this in `Cross.toml` actually?

src/docker/shared.rs
Alexhuszagh: But we don't override $HOME/.cargo. We only go up to when `relpath == "/"`, which is the current project directory or CWD. `root` is the project root or CWD. You can test this in a workspace:

```bash
$ cd test-workspace\workspace\binary
$ cross run --target arm-unknown-linux-gnueabihf --verbose
+ "C:\Program Files\Docker\Docker\resources\bin\docker.EXE" run --userns host -e 'PKG_CONFIG_ALLOW_CROSS=1' -e 'XARGO_HOME=/xargo' -e 'CARGO_HOME=/cargo' -e 'CARGO_TARGET_DIR=/target' -e 'CROSS_RUNNER=' -e TERM -e 'USER=ahusz' -v 'C:\Users\ahusz\git\test-workspace\external\external_lib:/mnt/c/Users/ahusz/git/test-workspace/external/external_lib' --rm --user 1000:1000 -v 'C:\Users\ahusz\.xargo:/xargo:Z' -v 'C:\Users\ahusz\.cargo:/cargo:Z' -v /cargo/bin -v 'C:\Users\ahusz\git\test-workspace\workspace:/mnt/c/Users/ahusz/git/test-workspace/workspace:Z' -v 'C:\Users\ahusz\.rustup\toolchains\stable-x86_64-unknown-linux-gnu:/rust:Z,ro' -v 'C:\Users\ahusz\git\test-workspace\workspace\target:/target:Z' -w /mnt/c/Users/ahusz/git/test-workspace/workspace/binary -v /mnt/c/Users/ahusz/git/test-workspace/workspace/binary/.cargo -v /mnt/c/Users/ahusz/git/test-workspace/workspace/.cargo -i -t ghcr.io/cross-rs/arm-unknown-linux-gnueabihf:main sh -c 'PATH=$PATH:/rust/bin cargo run --target arm-unknown-linux-gnueabihf --verbose'
```

Or, abridged for clarity: we only prevent `/mnt/c/Users/ahusz/git/test-workspace/workspace/binary/.cargo` and `/mnt/c/Users/ahusz/git/test-workspace/workspace/.cargo` from being mounted.

Alexhuszagh:
I've moved the configuration to `Cross.toml`, but this still needs to create the `.cargo` directories (or better yet, check if they exist and only mount them if they do).

docs/cross_toml.md
```diff
@@ -21,6 +21,7 @@ For example:
 
 ```toml
 [build.env]
+ignore-cargo-config = false
```
Alexhuszagh, line 24: We now use the `Cross.toml` config settings to allow people to skip the cargo config either via envvars or the config file.

src/docker/shared.rs
```diff
@@ -499,8 +512,49 @@ pub(crate) fn docker_envvars(
     Ok(())
 }
 
-pub(crate) fn docker_cwd(docker: &mut Command, paths: &DockerPaths) -> Result<()> {
+fn mount_to_ignore_cargo_config(
+    docker: &mut Command,
+    paths: &DockerPaths,
+    ignore_cargo_config: bool,
+) -> Result<()> {
+    let check_mount =
```
Alexhuszagh, line 520: This checks if the path exists on the host (canonicalized), and only if it does, mounts it. This avoids any permission issues or creating undesired directories.

src/docker/shared.rs
```diff
@@ -499,8 +512,49 @@ pub(crate) fn docker_envvars(
...
+                cmd.args(&["-v", &mount.join(&cargo_dir).as_posix()?]);
+            }
+
+            Ok(())
+        };
+    let is_cargo_home_parent = paths
+        .cargo_home()
+        .parent()
+        .map(|x| x == paths.host_root())
+        .unwrap_or_default();
```
Alexhuszagh, line 534: This check is probably unnecessary. This check should probably be removed. Thoughts? This won't effect mounting the cargo home directory to `/cargo`.

src/docker/shared.rs
```diff
@@ -499,8 +512,49 @@ pub(crate) fn docker_envvars(
...
+    if ignore_cargo_config && !is_cargo_home_parent {
+        let mount_root = Path::new(paths.mount_root());
+        let mount_cwd = Path::new(paths.mount_cwd());
+        check_mount(docker, &paths.cwd, mount_cwd, Path::new(""))?;
+        // CWD isn't guaranteed to be a subdirectory of the mount root.
```
Alexhuszagh, line 539: Since we've removed `/project`, this changes the logic here, so we don't necessarily have a relpath to the workspace root.

docs/cross_toml.md
Emilgardis: this is good, makes it more discoverable

src/docker/shared.rs
Alexhuszagh: I've removed this check, since I've tested to see if invoking a build from the parent of `CARGO_HOME` works, and it does, and doesn't touch `CARGO_HOME` at all. It just prevents that directory from being mounted at the same path as on the host, but still mounts it with all the data at `/cargo`.

Alexhuszagh:
I'm likely going to merge this into a PR with #931, since the two are complimentary and I also need that to support allowing full passing of Cargo configuration files to cross. So, we can therefore have the default behavior (`CARGO_HOME` and the project config files are passed), the behavior where everything except `CARGO_HOME` is skipped, and allowing everything to be included.

Alexhuszagh:
Superseded by #931, which now includes this functionality.

PR #939, opened 2022-07-11
Fix cross when using absolute paths to the current project.

### LINKED ISSUE #938: Remove `/project` Mount
If we provide an absolute path to a manifest path in a non-workspace directory, or use other code that expects to find data using an absolute path to the current project (a terrible decision, but a possible one), then `cross` current breaks because it will not be able to find the manifest path.

For example, if we use:

```bash
$ cd "$HOME"
$ git clone https://github.com/cross-rs/rust-cpp-hello-word
$ cd rust-cpp-hello-word
$ cross run --manifest-path "$HOME"\rust-cpp-hello-word\Cargo.toml --target arm-unknown-linux-gnueabihf
error: manifest path `/mnt/c/Users/ahusz/git/rust-cpp-hello-word/Cargo.toml` does not exist
```

This and #920 should add full support for user expectations on how paths will work in cross.

Alexhuszagh:
Closes #938.

### DIFF
--- a/.changes/939.json
+++ b/.changes/939.json
@@ -1,0 +1,5 @@
+{
+    "type": "changed",
+    "description": "Remove `/project` mounting, instead always mount in the same path as on the host",
+    "issues": [938]
+}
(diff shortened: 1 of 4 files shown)


Emilgardis (approved):
if path needs to be anonymized, prefer using `--remap-path-prefix`

I'd say this is a breaking change also, but maybe not

otherwise, lgtm, a bit sad but has to be done

Emilgardis (approved):
bors r+

PR #940, opened 2022-07-11
check for newlines in test

Emilgardis:


Alexhuszagh (approved):
bors r+

Alexhuszagh:
We're going to need to fix this: it's too active in the files it chooses. It checks files ignored by `.gitignore`, including binary files, and files locally ignored in `.git/info/exclude`.

PR #941, opened 2022-07-11
replace walkdir with ripgreps ignore crate

Emilgardis:


src/tests.rs
```diff
@@ -27,23 +27,23 @@ pub fn get_cargo_workspace() -> &'static Path {
...
     root: &'_ Path,
-    skip: &'a [impl AsRef<OsStr>],
-    ext: impl for<'s> Fn(Option<&'s std::ffi::OsStr>) -> bool + 'static,
-) -> impl Iterator<Item = Result<walkdir::DirEntry, walkdir::Error>> + 'a {
-    walkdir::WalkDir::new(root)
-        .into_iter()
+    skip: &'static [impl AsRef<OsStr> + Send + Sync + 'a],
+    ext: impl for<'s> Fn(Option<&'s std::ffi::OsStr>) -> bool + Sync + Send + 'static,
+) -> impl Iterator<Item = Result<ignore::DirEntry, ignore::Error>> {
+    ignore::WalkBuilder::new(root)
```
Alexhuszagh, line 33: I believe this should use the [gitignore](https://docs.rs/ignore/0.4.18/ignore/gitignore/index.html) builder.

Alexhuszagh: We should probably use the global builder and also the `add` for `.git/info/exclude` if present, since we do probably want to check dotfiles as well (like the ones in `.changes`).

Emilgardis: Not needed, the builders defaults already does all that

https://docs.rs/ignore/0.4.18/ignore/struct.WalkBuilder.html#ignore-rules

Alexhuszagh (approved):
bors r+

Emilgardis:
oops, we might want to pin a start and end on the bisect test. I've been toying with doing an actual useful bisect so that rustc forge artifacts are used as well but for now --start and --end will work fine.

Emilgardis:
0dd1f90556be6c7eccb843caaf64d92cab7f13b3 doesn't work, need to find another solution (like actually doing a bisect)

Alexhuszagh:
It might be worth actually keeping the range though: we don't need to go to nightly builds in 2017.

Emilgardis:
bors retry

Emilgardis:
Error is rustup, as reported on [reddit](https://old.reddit.com/r/rust/comments/vwqhs7/rustup_1250_released/iftpmh3/)

will probably file an issue

Emilgardis:
unable to reproduce in a new repo: https://github.com/Emilgardis/rustup-windows-issue/runs/7297777595?check_suite_focus=true

lets try windows specifically

bors try --target x86_64-pc-windows-msvc

Emilgardis:
it failed again, might be a concurrency issue, unsure how the windows virtual environment works for this. Will bypass with `rustup set auto-self-update disable`

bors cancel

Alexhuszagh (approved):

PR #904, opened 2022-07-03
Various bug fixes for cross.

Alexhuszagh:
Fixed `cargo metadata` failing with unstable arguments in rustflags if the default toolchain was not nightly. We also log any warnings that occurred if we cannot get metadata. Also fixed a lack of formatting in messages with remote docker support. The mount paths are also correctly processed if using docker-in-docker for mounted volumes.

Closes #901.
Closes #903.
Required for #873.

PR #942, opened 2022-07-12
Use non-canonical paths for mount paths.

### LINKED ISSUE #920: Don't Canonicalize Paths
Path canonicalization is generally a good idea for security reasons, and we should do it for the uniqueness of our hashes, but it can affects crates that wish to mount volumes that are symlinked and get data specifically using relative paths or absolute paths to the volume, which can differ from the canonical path.

Say my crate does something like:

**main.rs**

```rust
pub fn main() {
    println!("{}", std::fs::read_to_string(std::path::Path::new("/tmp/config.toml")).unwrap());
}
```

Now, `/tmp/config.toml` is on macOS, which is actually mounted at `/private/tmp/config.toml`, so canonicalizing the path will mean `/tmp` is not mounted, but `/private/tmp` is. So in this case, we have a path where the parent directory is a symlink and this code will now fail.

Alexhuszagh:
Symlinks and canonical paths can destroy assumptions about paths: say I have a file at `/tmp/path/to/file` I reference in my crate, and ask to mount `/tmp/path/to` on the container. Currently, on macOS, since `/tmp` is a symlink to `/private/tmp`, the code that expects `/tmp/path/to/file` will fail because in the container it will be mounted at `/private/tmp/path/to/file`.

In addition, this fixes `DeviceNS` parsing with drive letters (this previously didn't matter since it never occurred, due to `dunce` canonicalization). This PR also fixes a minor bug where the host root instead of the var names was passed as the environment variable (introduced in #904).

Closes #920.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -12,14 +12,11 @@ use crate::cargo::{cargo_metadata_with_args, CargoMetadata};
 use crate::config::{bool_from_envvar, Config};
 use crate::errors::*;
 use crate::extensions::{CommandExt, SafeCommand};
-use crate::file::{self, write_file, ToUtf8};
+use crate::file::{self, write_file, PathExt, ToUtf8};
 use crate::id;
 use crate::rustc::QualifiedToolchain;
 use crate::shell::{MessageInfo, Verbosity};
 use crate::Target;
-
-#[cfg(target_os = "windows")]
-use crate::file::PathExt;
 
 pub use super::custom::CROSS_CUSTOM_DOCKERFILE_IMAGE_PREFIX;
 
@@ -287,6 +284,10 @@ impl Directories {
         }
         create_target_dir(target)?;
 
+        // get our mount paths prior to canonicalizing them
+        let cargo_mount_path = cargo.as_posix_absolute()?;
+        let xargo_mount_path = xargo.as_posix_absolute()?;
+
         // now that we know the paths exist, canonicalize them. this avoids creating
         // directories after failed canonicalization into a shared directory.
         let cargo = file::canonicalize(&cargo)?;
@@ -307,17 +308,15 @@ impl Directories {
             &metadata.workspace_root
         });
 
-        // root is either workspace_root, or, if we're outside the workspace root, the current directory
-        // NOTE: host root has already found the mount path
-        let mount_root = canonicalize_mount_path(&host_root)?;
+        // on Windows, we can not mount the directory name directly. Instead, we use wslpath to convert the path to a linux compatible path.
+        // NOTE: on unix, host root has already found the mount path
+        let mount_root = host_root.as_posix_absolute()?;
         let mount_cwd = mount_finder.find_path(cwd, false)?;
 
         toolchain.set_sysroot(|p| mount_finder.find_mount_path(p));
 
         // canonicalize these once to avoid syscalls
-        let cargo_mount_path = canonicalize_mount_path(&cargo)?;
-        let xargo_mount_path = canonicalize_mount_path(&xargo)?;
-        let sysroot_mount_path = canonicalize_mount_path(toolchain.get_sysroot())?;
+        let sysroot_mount_path = toolchain.get_sysroot().as_posix_absolute()?;
 
         Ok(Directories {
             cargo,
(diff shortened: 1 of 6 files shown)


Alexhuszagh:
I still have to test this extensively on macOS and with symlinks, just to ensure this works well (and also for Docker-in-Docker).

Alexhuszagh:
This mostly works: there's some issues with Docker-in-Docker it seems, but the core logic is sound enough. There's no way to use non-canonical paths for the current project directory (easily), since `cargo metadata` resolves any symlinks. That being said, everything else seems to work.

Specifically, this fails in workspaces for docker-in-docker: this is the mount point:
```bash
-v $PREFIX/merged/tmp/tmp.BHCKdn/external/external_lib:/tmp/tmp.BHCKdn/external/external_lib
```

And this is where it was supposed to be read:

```bash
failed to read `/var/lib/docker/overlay2/c7c0edabf2c42ce958ffef1d9eb8b59b903c0bdf2605682a8ca2cb54a2662ef0/merged/tmp/tmp.BHCKdn/external/external_lib/Cargo.toml`
```

Which means that the lack of the canonicalization seems to have undone this. We may have to canonicalize paths for docker-in-docker.

src/file.rs
```diff
@@ -261,32 +292,44 @@ mod tests {
     }
 
     #[test]
-    fn as_posix() {
-        result_eq(p!(".").join("..").as_posix(), Ok("./..".to_owned()));
-        result_eq(p!(".").join("/").as_posix(), Ok("/".to_owned()));
-        result_eq(p!("foo").join("bar").as_posix(), Ok("foo/bar".to_owned()));
-        result_eq(p!("/foo").join("bar").as_posix(), Ok("/foo/bar".to_owned()));
+    fn as_posix_relative() {
+        result_eq(
```
Emilgardis, line 296: nit not related to these changes

`result_eq` should have a `#[track_caller]` attr

src/docker/shared.rs
```diff
@@ -537,37 +516,39 @@ pub(crate) fn docker_mount(
             None => env::var(var),
         };
 
+        // NOTE: we use canonical paths on the host, since it's unambiguous.
+        // to avoid changing paths on the mount, we use absolute paths.
```
Emilgardis, line 520: this note does not make much snse to me, specifically `to avoid changing paths`, what does "changing paths" mean?

src/docker/shared.rs
```diff
@@ -537,37 +516,39 @@ pub(crate) fn docker_mount(
...
-            let mount_path = mount_cb(docker, host_path.as_ref())?;
-            docker.args(&["-e", &format!("{}={}", host_path, mount_path)]);
+            let canonical_path = file::canonicalize(&val)?;
+            let host_path = paths.mount_finder.find_path(&canonical_path, true)?;
+            let absolute_path = Path::new(&val).as_posix_absolute()?;
+            let mount_path = paths
+                .mount_finder
+                .find_path(Path::new(&absolute_path), true)?;
+            mount_cb(docker, host_path.as_ref(), mount_path.as_ref())?;
+            docker.args(&["-e", &format!("{}={}", var, mount_path)]);
```
Emilgardis, line 530: oops

Emilgardis (changes requested):

src/docker/shared.rs
Alexhuszagh: It means the path canonicalization without the symlinks expected for them on the host can change the mounted path and cause code to break. It's meant to convey the same thing as in https://github.com/cross-rs/cross/issues/920#issue-1298118330. I'll make it a bit clearer.

Alexhuszagh: I've updated the comment with an example and clearer wording:

```rust
// NOTE: we use canonical paths on the host, since it's unambiguous.
// however, for the mounted paths, we use the same path as was
// provided. this avoids canonicalizing symlinks which then causes
// the mounted path to differ from the path expected on the host.
// for example, if `/tmp` is a symlink to `/private/tmp`, canonicalizing
// it would lead to us mounting `/tmp/process` to `/private/tmp/process`,
// which would cause any code relying on `/tmp/process` to break.
```

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #945, opened 2022-07-14
Fix pre-build Docker image names with Podman.

### LINKED ISSUE #944: Podman Adds Registry/Repository to Images
When building an image, if creating a custom image with the following logic:

```bash
docker build --tag cross-custom-hellotest:x86_64-unknown-linux-gnu-e4cf7-pre-build -f Dockerfile .
```

With a Dockerfile as follows:

```Dockerfile
FROM ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main
```

Docker names it to `docker.io/library/cross-custom-hellotest:x86_64-unknown-linux-gnu-e4cf7-pre-build`: IE, it uses the default Dockerhub registry.

Running the same command with Podman fails:

```bash
podman build --tag cross-custom-hellotest:x86_64-unknown-linux-gnu-e4cf7-pre-build -f Dockerfile .
```

Since it names it to `localhost/cross-custom-hellotest:x86_64-unknown-linux-gnu-e4cf7-pre-build`.

### Solution

The solution seems to provide it a repository with a registry built-in, likely `localhost` or something similar. For example, if we change both to `localhost/cross-rs/cross-custom-hellotest:x86_64-unknown-linux-gnu-e4cf7-pre-build`, both build to the same tag. This likely would also affect any non-Docker container engines.

Alexhuszagh:
Fixes #944.

### DIFF
--- a/.changes/945.json
+++ b/.changes/945.json
@@ -1,0 +1,5 @@
+{
+    "type": "fixed",
+    "description": "fix pre-build hook image naming with podman",
+    "issues": [944]
+}
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -9,7 +9,7 @@ use crate::{errors::*, file, CommandExt, ToUtf8};
 
 use super::{image_name, parse_docker_opts, path_hash};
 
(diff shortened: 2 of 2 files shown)


Emilgardis (approved):
bors r+

PR #946, opened 2022-07-14
 Ignore tests that touch the filesystem if sandboxed. 

### LINKED ISSUE #943: Unit tests failing during NixOS build
### Checklist

### Describe your issue

When updating the NixOS package from 0.2.1 to 0.2.4, one of tests fail during execution. Log is below:

```
test docker::shared::tests::directories::test_docker_in_docker ... ok
test docker::shared::tests::directories::test_host ... FAILED
test tests::toml::toml_check ... ok
test cross_toml::tests::merge ... ok

failures:

---- docker::shared::tests::directories::test_host stdout ----
Error: Permission denied (os error 13)

Location:
    src/docker/shared.rs:965:13
thread 'docker::shared::tests::directories::test_host' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /build/rustc-1.61.0-src/library/test/src/lib.rs:186:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    docker::shared::tests::directories::test_host
```

The affected code is:
https://github.com/cross-rs/cross/blob/6cd09b3eb28da7f4e21ee9e3ecee89a3d309c1dc/src/docker/shared.rs#L966-L983

_No response_

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [x] docker
- [x] podman

### cross version

cross 0.2.4

### Example

_No response_

_No response_

Alexhuszagh:
Adds an environment variable `CROSS_SANDBOXED` that skips tests that touch the filesystem if set.

Closes #943.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -23,6 +23,11 @@ fn main() {
         .unwrap()
         .write_all(commit_info().as_bytes())
         .unwrap();
+
+    if env::var("CROSS_SANDBOXED").is_ok() {
+        println!("cargo:rustc-cfg=cross_sandboxed");
+    }
+    println!("cargo:rerun-if-env-changed=CROSS_SANDBOXED");
 }
 
 fn commit_info() -> String {
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -1059,6 +1059,7 @@ mod tests {
         }
 
         #[test]
+        #[cfg_attr(cross_sandboxed, ignore)]
         fn test_host() -> Result<()> {
             let vars = unset_env();
             let mount_finder = MountFinder::new(vec![]);


Alexhuszagh:
I think using an `#[ignore]` attribute is the best solution, since we still want to be able to run the tests if desired on NixOS, and there might be other scenarios where access to the filesystem is denied other than just NixOS. Adding it to the build script feels a bit overkill.

Alexhuszagh:
> I wish there was a way to make ignore configurable with envvars or something. I'm fine with this, however I'd prefer it to be not ignored unconditionally. A runtime skip maybe?

Sure, we could look for `NIX_PATH`, which is always set. And we could check for another envvar to not skip it if present. This way, it can still be tested. The other way is to add it based on the envvar to the build script, and do `#[cfg_attr(nixos, ignore)]`.

otavio (approved):

Emilgardis:
worst case for that, just pass `--ignored`

src/build.rs
```diff
@@ -28,6 +28,10 @@ fn main() {
         .unwrap()
         .write_all(docker_images().as_bytes())
         .unwrap();
+
+    if env::var("NIX_PATH").is_ok() {
+        println!("cargo:rustc-cfg=cross_nixos");
+    }
```
otavio, line 34: In Nix when using Flake, this is not set.

Alexhuszagh: Any other envvar that will be set? I tried using the base system for the Docker image, which is missing a lot of things. I can try checking if one of a subset of envvars are set, short-circuiting once the first is.

Alexhuszagh: In the base Docker image for NixOS, the only two NIX-specific envvars set are:
- `NIX_PATH`
- `NIX_SSL_CERT_FILE`

I don't the latter is reliable. `uname` gets the host data for the container, so it's also not reliable. There is no `/etc/os-release` either here. Is checking for `/etc/nix/nix.conf` reliable? If not I'll just change it back to `#[ignored]`.

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --locked --all-targets --workspace --all-features
+          args: --locked --all-targets --workspace --all-features -- --include-ignored
```
Emilgardis, line 86: this is unfortunate, right now there's no effect but may cause issues in the future if more ignored tests are added

Alexhuszagh: Not sure the best way to solve this then.

Emilgardis: what about just requiring an env var to be set to disable, the nix builder can set the env var.

Emilgardis (approved):

Alexhuszagh:
@otavio I've added the `CROSS_SANDBOXED` environment variable, which if set skips any tests that touch the filesystem. Seem good?

Alexhuszagh:
@otavio Does this look good?

otavio (approved):
We can merge this and I handle this in NixOS side.

PR #947, opened 2022-07-14
Resolve symlinks for cargo and xargo home.

### LINKED ISSUE #280: Cross compiling forces a recompile for other architectures
Running the following: 
```
cargo build --release
cross build --release --target armv7-unknown-linux-musleabihf
cargo build --release
cross build --release --target armv7-unknown-linux-musleabihf
```

Results in 4 full recompiles. I'm building for two architectures simultaneously and this means my compile and run cycle take ~ 7 minutes rather than ~30s. Is there any way to cache the build for each architecture so that I'm only compiling my changes each time?

### LINKED ISSUE #373: Podman Mounts Symlinks as Root
**Maintainers Note:** The originally title and post contents are below. This has been modified since the current issue is related, but different, as described [here](https://github.com/cross-rs/cross/issues/373#issuecomment-1173190087).

When using Podman, if we have a symlink to our bind mount, we a permissions error, saying it cannot access the data within the volume. This only happens with paths we do not canonicalize. Here, we create an example with symlinks for our `CARGO_HOME` and `XARGO_HOME`. This probably also occurs for other bind mounts.

```bash
$ mkdir registry
$ export CARGO_HOME=$(realpath registry)/.cargo
$ export XARGO_HOME=$(realpath registry)/.xargo
$ ln -s ~/.cargo "${CARGO_HOME}"
$ ln -s ~/.xargo "${XARGO_HOME}"
$ git clone https://github.com/cross-rs/rust-cpp-hello-word
$ cd rust-cpp-hello-word
$ cargo build                                           # works
$  cross build --target aarch64-unknown-linux-gnu       # also just works
```

If we run in Podman, these directories are owned by `root`, not by the current user.

```bash
$ ls -la / | grep argo
drwxrwxr-x.   1 root   root     164 Jul  3 22:57 cargo
drwxrwxr-x.   1 root   root       6 Jun  2 03:36 xargo
```

Meanwhile, in Docker, they're mounted as the current user, and `cross` works fine:

```bash
ls -la / | grep argo
drwxrwxr-x.   1 1000 1000  164 Jul  3 22:57 cargo
drwxrwxr-x.   1 1000 1000    6 Jun  2 03:36 xargo
```

---
# cross fails with "failed to create directory - file exists"

Steps to reproduce:
```sh
mkdir -p /tmp/.cargo
rm -rf ~/.cargo
ln -sT /tmp/.cargo ~/.cargo
git clone https://github.com/zserik/crulz-rs.git
cd crulz-rs
git checkout 754196bbbc4d99ed6634e83b8e3d2b7571ce7209
cross check --target armv7-unknown-linux-musleabihf
```
(working directory (should be rather irrelevant) is equivalent to: https://github.com/zserik/crulz-rs/tree/754196bbbc4d99ed6634e83b8e3d2b7571ce7209)

It happens when either `$CARGO_HOME`, `$CARGO_HOME/registry`, `$CARGO_HOME/registry/cache` or `$CARGO_HOME/registry/index` (or multiple of them) are symlinks.

I tried to run (with the same target):
* `cross check`
* `cross clippy`
* `cross build`
* `cross build --release`

I got the following error:
```
error: failed to create directory `/cargo/registry/cache/github.com-1ecc6299db9ec823`
error: failed to create directory `/cargo/registry/index/github.com-1ecc6299db9ec823`

Caused by:
  File exists (os error 17)
```

### LINKED ISSUE #551: Running cross and cargo builds for different targets breaks incremental compilation
When building a project for a couple of different targets, one within cross and one outside, swapping targets seems to cause the whole project to rebuild. I'm not _entirely_ sure of the mechanism, but, suspect it is either related to changes in the profile (per https://github.com/rust-lang/cargo/issues/4961) or different compiler versions. This would seem to imply the profile is project-global rather than per-target and I haven't dug into this at all.

In case anyone else runs into this, if you only have one target under `cross` you can work around the issue by placing the project in a workspace then building natively (via `cargo build`) from the workspace folder and via cross in the project folder (as the docker container cannot see up to the workspace directory) so that the target directories are split, in which case incremental builds work properly. If you need more than one `cross` target, I'm not sure there's a great option, maybe propagating `sccache` config / files into the `cross` container :-/

Alexhuszagh:
Resolve symlinks for the xargo and cargo home (as well as the Nix store) prior to mounting, since they are mounted at a fixed location anyway. This is because podman mounts symlinks as root by default. We always canonicalize the paths on the host, after creating them, to ensure that any symlinks are resolved.

Say we have `/path/to/home`, and `/path/to` is a symlink but `/path/to/home` doesn't exist yet. Trying to canonicalize `/path/to/home/.xargo` will fail, and will still be a symlink if we maybe canonicalize before. First creating the directories and canonicalizing them on the host should always work.

Change the mount points from `/cargo`, `/xargo`, and `/rust` to the same paths as on the host, which avoids unnecessary recompilation when `cargo` and `cross` are intermittently used.

Closes #280.
Closes #373.
Closes #551.

### DIFF
--- a/.changes/947.json
+++ b/.changes/947.json
@@ -1,0 +1,12 @@
+[
+    {
+        "type": "internal",
+        "description": "resolve symlinks to cargo and xargo home before mounting",
+        "issues": [373]
+    },
+    {
+        "type": "fixed",
+        "description": "mount cargo, xargo, and the sysroot at the same path as on the host to avoid unnecessary recompilation.",
+        "issues": [551]
+    }
+]
--- a/src/docker/remote.rs
+++ b/src/docker/remote.rs
@@ -241,24 +241,19 @@ fn copy_volume_files_nocache(
 pub fn copy_volume_container_xargo(
     engine: &Engine,
     container: &str,
-    xargo_dir: &Path,
-    target: &Target,
+    dirs: &Directories,
     mount_prefix: &Path,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
-    // only need to copy the rustlib files for our current target.
-    let triple = target.triple();
-    let relpath = Path::new("lib").join("rustlib").join(&triple);
-    let src = xargo_dir.join(&relpath);
-    let dst = mount_prefix.join("xargo").join(&relpath);
-    if Path::new(&src).exists() {
+    let dst = mount_prefix.join(&dirs.xargo_mount_path_relative()?);
+    if dirs.xargo.exists() {
         create_volume_dir(
             engine,
             container,
             dst.parent().expect("destination should have a parent"),
             msg_info,
         )?;
-        copy_volume_files(engine, container, &src, &dst, msg_info)?;
+        copy_volume_files(engine, container, &dirs.xargo, &dst, msg_info)?;
     }
 
     Ok(())
@@ -267,23 +262,23 @@ pub fn copy_volume_container_xargo(
 pub fn copy_volume_container_cargo(
     engine: &Engine,
     container: &str,
-    cargo_dir: &Path,
+    dirs: &Directories,
     mount_prefix: &Path,
     copy_registry: bool,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
-    let dst = mount_prefix.join("cargo");
+    let dst = mount_prefix.join(&dirs.cargo_mount_path_relative()?);
     let copy_registry = env::var("CROSS_REMOTE_COPY_REGISTRY")
         .map(|s| bool_from_envvar(&s))
         .unwrap_or(copy_registry);
 
     if copy_registry {
-        copy_volume_files(engine, container, cargo_dir, &dst, msg_info)?;
+        copy_volume_files(engine, container, &dirs.cargo, &dst, msg_info)?;
     } else {
         // can copy a limit subset of files: the rest is present.
         create_volume_dir(engine, container, &dst, msg_info)?;
-        for entry in fs::read_dir(cargo_dir)
-            .wrap_err_with(|| format!("when reading directory {cargo_dir:?}"))?
+        for entry in fs::read_dir(&dirs.cargo)
+            .wrap_err_with(|| format!("when reading directory {:?}", dirs.cargo))?
         {
             let file = entry?;
             let basename = file
@@ -371,17 +366,17 @@ fn warn_symlinks(had_symlinks: bool, msg_info: &mut MessageInfo) -> Result<()> {
 fn copy_volume_container_rust_base(
     engine: &Engine,
     container: &str,
-    sysroot: &Path,
+    dirs: &Directories,
     mount_prefix: &Path,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
     // the rust toolchain is quite large, but most of it isn't needed
     // we need the bin, libexec, and etc directories, and part of the lib directory.
-    let dst = mount_prefix.join("rust");
+    let dst = mount_prefix.join(&dirs.sysroot_mount_path_relative()?);
     let rustlib = Path::new("lib").join("rustlib");
     create_volume_dir(engine, container, &dst.join(&rustlib), msg_info)?;
     for basename in ["bin", "libexec", "etc"] {
-        let file = sysroot.join(basename);
+        let file = dirs.get_sysroot().join(basename);
         copy_volume_files(engine, container, &file, &dst, msg_info)?;
     }
 
@@ -393,9 +388,9 @@ fn copy_volume_container_rust_base(
     // SAFETY: safe, single-threaded execution.
     let tempdir = unsafe { temp::TempDir::new()? };
     let temppath = tempdir.path();
-    fs::create_dir_all(&temppath.join(&rustlib))?;
+    file::create_dir_all(&temppath.join(&rustlib))?;
     let mut had_symlinks = copy_dir(
-        &sysroot.join("lib"),
+        &dirs.get_sysroot().join("lib"),
         &temppath.join("lib"),
         true,
         0,
@@ -404,7 +399,7 @@ fn copy_volume_container_rust_base(
 
     // next, copy the src/etc directories inside rustlib
     had_symlinks |= copy_dir(
-        &sysroot.join(&rustlib),
+        &dirs.get_sysroot().join(&rustlib),
         &temppath.join(&rustlib),
         true,
         0,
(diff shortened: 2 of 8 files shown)


Emilgardis (approved):
bors r+

Alexhuszagh:
Yeah, I planned on converting this to a draft before my internet died.

Alexhuszagh:
I think this is ready: it fails if the symlink exists but is broken, but that's a feature I believe, and will error with the following warning:

```bash
rror:                                                                                                                                                                                              
   0: File exists (os error 17)
```

We probably want to wrap that error to explain what file exists, however.

.changes/947.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "type": "internal",
+        "description": "resolve symlinks for cargo and xargo home",
+        "issues": [373]
+    },
+    {
+        "type": "fixed",
+        "description": "mount cargo, xargo, and the sysroot at the same path as on the host to avoid unnecessary recompilation.",
```
Emilgardis, line 9: this is only true for linux, no?

.changes/947.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "type": "internal",
+        "description": "resolve symlinks for cargo and xargo home",
```
Emilgardis, line 4:
```suggestion
        "description": "resolve symlinks to cargo and xargo home before mounting",
```

src/docker/remote.rs
```diff
@@ -482,41 +479,29 @@ pub fn copy_volume_container_rust_triple(
 pub fn copy_volume_container_rust(
     engine: &Engine,
     container: &str,
+    dirs: &Directories,
     toolchain: &QualifiedToolchain,
     target_triple: Option<&TargetTriple>,
     mount_prefix: &Path,
     msg_info: &mut MessageInfo,
```
Emilgardis, line 486: dirs contains toolchain

src/docker/remote.rs
```diff
@@ -1066,9 +1038,18 @@ pub(crate) fn run(
...
+            &dirs.xargo,
+            mount_prefix_path.join(Path::new(&dirs.xargo_mount_path_relative()?)),
+        ),
+        (
+            &dirs.cargo,
+            mount_prefix_path.join(Path::new(&dirs.cargo_mount_path_relative()?)),
+        ),
+        (
+            &sysroot,
+            mount_prefix_path.join(Path::new(&dirs.sysroot_mount_path_relative()?)),
```
Emilgardis, line 1051: `MountFinder::join` takes an `AsRef<Path>`, the manual `Path::new` shouldn't be needed.

Emilgardis (changes requested):

.changes/947.json
Alexhuszagh: Nope. Also true for Windows (tested) and macOS. It turns out, it doesn't need a recompile if the Windows-style path points to the same location as the WSL path.

Emilgardis:
This is a cool change, it also makes it more clear in verbose build that you're actually running a binary mounted from your host system, since `/my/home/.rustup/toolchain/toolchain` is more prevalent now

src/docker/remote.rs
```diff
@@ -371,17 +366,17 @@ fn warn_symlinks(had_symlinks: bool, msg_info: &mut MessageInfo) -> Result<()> {
...
     container: &str,
-    sysroot: &Path,
+    dirs: &Directories,
     mount_prefix: &Path,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
     // the rust toolchain is quite large, but most of it isn't needed
     // we need the bin, libexec, and etc directories, and part of the lib directory.
-    let dst = mount_prefix.join("rust");
+    let dst = mount_prefix.join(Path::new(&dirs.sysroot_mount_path_relative()?));
```
Emilgardis, line 375: same thing here, `join` takes a `AsRef<Path>`

src/docker/remote.rs
```diff
@@ -418,21 +413,21 @@ fn copy_volume_container_rust_base(
...
     container: &str,
-    sysroot: &Path,
+    dirs: &Directories,
     mount_prefix: &Path,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
     // copy over all the manifest files in rustlib
     // these are small text files containing names/paths to toolchains
-    let dst = mount_prefix.join("rust");
+    let dst = mount_prefix.join(Path::new(&dirs.sysroot_mount_path_relative()?));
```
Emilgardis, line 422: ditto

src/docker/remote.rs
```diff
@@ -447,18 +442,20 @@ fn copy_volume_container_rust_manifest(
...
+    dirs: &Directories,
     target_triple: &TargetTriple,
     mount_prefix: &Path,
     skip_exists: bool,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
-    let sysroot = toolchain.get_sysroot();
     // copy over the files for a specific triple
-    let dst = mount_prefix.join("rust");
+    let dst = mount_prefix.join(Path::new(&dirs.sysroot_mount_path_relative()?));
```
Emilgardis, line 452: d3

src/docker/remote.rs
```diff
@@ -267,23 +262,23 @@ pub fn copy_volume_container_xargo(
...
     engine: &Engine,
     container: &str,
-    cargo_dir: &Path,
+    dirs: &Directories,
     mount_prefix: &Path,
     copy_registry: bool,
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
-    let dst = mount_prefix.join("cargo");
+    let dst = mount_prefix.join(Path::new(&dirs.cargo_mount_path_relative()?));
```
Emilgardis, line 270: d4

src/docker/shared.rs
```diff
@@ -323,6 +326,39 @@ impl Directories {
...
+        canonicalize_mount_path(&self.cargo)
+    }
+
+    pub fn xargo_mount_path(&self) -> Result<String> {
+        canonicalize_mount_path(&self.xargo)
+    }
+
+    pub fn sysroot_mount_path(&self) -> Result<String> {
+        canonicalize_mount_path(self.get_sysroot())
+    }
```
Emilgardis, line 340: I feel like these should be canonicalized in `create`, to avoid syscalls

Emilgardis (changes requested):
lgtm! overall a huge improvement in ux

I feel like we should make some path shims though, to easily descern between what's on the host and what is in the image, that can be done separately though.

Some small things left

Emilgardis:
test_host is failing due to `stable-x86_64-unknown-linux-gnu` not existing, I think it's fine to pre-install this toolchain in the test job (only in the test job though)

Alexhuszagh:
> test_host is failing due to `stable-x86_64-unknown-linux-gnu` not existing, I think it's fine to pre-install this toolchain in the test job (only in the test job though)

We might need a better strategy. Let me think...

Alexhuszagh:
Ok I've changed this to use the host target triple (only for the unittest), so the toolchain will always be installed.

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #949, opened 2022-07-15
Remove OpenSSL and SQLite3 from FreeBSD images.

Alexhuszagh:


Alexhuszagh:
Still a draft: I need to add a `CHANGELOG` entry to this, but the rationale is as follows: FreeBSD's packagelist doesn't provide downloads for the correct versions of OpenSSL and SQLite3. For example, using the following settings:

```bash
$ pkg_source="https://pkg.freebsd.org/FreeBSD:12:amd64/quarterly"
$ curl --retry 3 -sSfL "$pkg_source/packagesite.txz" -O
$ tar xvf packagesite.txz
$ openssl_ver=$(jq -c '. | select ( .name == "openssl" ) | .version' "packagesite.yaml")
$ openssl_ver=${openssl_ver//'"'/}
$ echo "${openssl_ver}"
1.1.1q,1
$ wget "${pkg_source}/All/openssl-${openssl_ver}.txz" 
Will not apply HSTS. The HSTS database must be a regular and non-world-writable file.
ERROR: could not open HSTS store at '/home/ahuszagh/.wget-hsts'. HSTS will be disabled.
--2022-07-15 00:59:00--  https://pkg.freebsd.org/FreeBSD:12:i386/quarterly/All/openssl-1.1.1q,1.txz
Resolving pkg.freebsd.org (pkg.freebsd.org)... 2610:1c1:1:606c::50:1, 96.47.72.71
Connecting to pkg.freebsd.org (pkg.freebsd.org)|2610:1c1:1:606c::50:1|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2022-07-15 00:59:01 ERROR 404: Not Found.
```

Note that this doesn't reproduce with other packages, such as `zstd`. In short, FreeBSD builds are fragile, especially with their packages, and we can't trust the package lists to actually contain the package versions they describe. We've removed `openssl` for every other image: it's natural to remove it here as well.

PR #950, opened 2022-07-15
Fix FreeBSD migration to `.pkg` extensions.

Alexhuszagh:
Uses the `path` key rather than the `version` extension to get the proper location for the package URL, which is still compressed as a tar XZ archive.

### DIFF
--- a/docker/Dockerfile.i686-unknown-freebsd
+++ b/docker/Dockerfile.i686-unknown-freebsd
@@ -10,12 +10,14 @@ RUN /cmake.sh
 COPY xargo.sh /
 RUN /xargo.sh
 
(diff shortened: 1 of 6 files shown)


Alexhuszagh:
bors try --target x86_64-unknown-freebsd

Alexhuszagh:
bors try-
bors try --target x86_64-unknown-freebsd

Emilgardis (approved):

Alexhuszagh:
It might be worth adding a very rudimentary package install script here, since we effectively do this anyway, which would make it much simpler to install FreeBSD packages.

Alexhuszagh:
bors try --target x86_64-unknown-freebsd

Emilgardis (approved):
bors r+

PR #952, opened 2022-07-16
Run non-x86 binaries natively if on a compatible host.

Alexhuszagh:


Alexhuszagh:
This needs some changes, specifically if we use non-Debian-based hosts. IE, we need a way to get the host architecture if we're using Fedora/CentOS/RHEL, Alpine, Pacman, etc.

Alexhuszagh:
This uses `dpkg` as a reasonable default, and then falls back to `uname -m` if present, which is then normalized to the Debian value. This attempts to determine if the host can natively run the binary if `CROSS_RUNNER` is not provided.

docker/base-runner.sh
```diff
@@ -0,0 +1,142 @@
...
+    # there's numerous compatibility modes, so we want
+    # to ensure that these are valid. we also want to
+    # use dpkg if it's available since it gives hard-float
+    # information on compatible architectures
+    local host
+    local arch
+    if dpkg >/dev/null 2>&1; then
+        host=$(dpkg --print-architecture)
+        arch="${host}"
+    else
```
Alexhuszagh, line 13: Normalize to Debian, because it holds slightly more information than `uname -m`, and all the supported architectures that Docker can run images on natively are Debian-supported targets.

docker/base-runner.sh
```diff
@@ -0,0 +1,142 @@
...
+
+        case "${arch}" in
+            aarch64|armv8b|armv8l)
+                arch=arm64
+                ;;
+            aarch64_be)
+                arch=arm64be
+                ;;
+            arm)
+                arch=unknown
```
Alexhuszagh, line 25: We don't know if the target is ARMv5TE, ARMv6, or ARMv7, or if it's a soft-float or hard-float target. Always default to Qemu if it's a non-Debian-based 32-bit ARM host.

docker/base-runner.sh
```diff
@@ -0,0 +1,142 @@
...
+                ;;
+            *)
+                ;;
+        esac
+    fi
+
+    echo "${arch}"
+}
+
+normalize_arch() {
```
Alexhuszagh, line 50: Normalize to Debian, same as above.

docker/base-runner.sh
```diff
@@ -0,0 +1,142 @@
...
+
+is_native_binary() {
+    # determines if the binary can run natively on the host
+    local arch="${1}"
+    local host
+    local target
+    host=$(host_architecture)
+    target=$(normalize_arch "${arch}")
+
+    # FIXME: this is not comprehensive. add more compatible architectures.
```
Alexhuszagh, line 98: Probably a lot more architectures (especially ARMv8) with ARMv7.

docker/Dockerfile.thumbv7neon-unknown-linux-gnueabihf
```diff
@@ -23,7 +23,7 @@ RUN /dropbear.sh
 COPY linux-image.sh /
 RUN /linux-image.sh armv7
 
-COPY linux-runner /
+COPY linux-runner base-runner.sh /
 
 # Export all target binutils just in case required.
 ENV CARGO_TARGET_THUMBV7NEON_UNKNOWN_LINUX_GNUEABIHF_LINKER=arm-linux-gnueabihf-gcc \
```
Emilgardis, line 29: should specify `runner = hf` I think

Emilgardis (changes requested):

docker/base-runner.sh
Alexhuszagh: This actually needs to be `arm*`, since it can outp[ut `armv7l`, even on ARMv6 (at least according to Alpine).

Emilgardis (approved):
bors r+

docker/linux-runner
```diff
@@ -2,6 +2,9 @@
 
 set -e
 
+# shellcheck disable=SC1091
+. base-runner.sh
```
Alexhuszagh, line 6: Wait, this needs to be `. /base-runner.sh`.

Alexhuszagh:

Alexhuszagh:
bors r=Emilgardis

PR #953, opened 2022-07-16
Refactor into a base image.

Alexhuszagh:


Alexhuszagh:
This still needs to fix deployment and CI: we need to make sure the base image is always built first, and that it is staged for our other image build. It also needs to then be deployed as `ghcr.io/cross-rs/base:$tag`. The other solution would be to use templates: we run a script which would configure every image, which would avoid deploying another image. But I'm not sure I like that solution.

xtask/src/util.rs
```diff
@@ -257,53 +257,6 @@ pub fn read_dockerfiles(msg_info: &mut MessageInfo) -> cross::Result<Vec<(PathBu
...
-            .map(|(k, _)| k)
-            .ok_or_else(|| eyre::eyre!("have no dockerfiles"))?;
-
-        if actual_base != max_base {
-            eyre::bail!("most common base image is {max_base} but source code has {actual_base}")
-        } else {
-            Ok(())
-        }
-    }
-}
```
otavio, line 305: Shouldn't this check the base image instead?

Alexhuszagh: Fixed as of the latest builds, which checks the base in `Dockerfile.ubuntu` is the same as the `CROSS_IMAGE` version.

Alexhuszagh:
As expected, this is failing because `ghcr.io/cross-rs/ubuntu:main` doesn't exist.
https://github.com/cross-rs/cross/actions/runs/3159167477/jobs/5142204364#step:10:95

```
#3 [internal] load metadata for ghcr.io/cross-rs/ubuntu:main
#3 ERROR: failed to authorize: failed to fetch anonymous token: unexpected status: 403 Forbidden
------
 > [internal] load metadata for ghcr.io/cross-rs/ubuntu:main:
------
Dockerfile.x86_64-unknown-linux-gnu:1
--------------------
   1 | >>> FROM ghcr.io/cross-rs/ubuntu:main
   2 |     ARG DEBIAN_FRONTEND=noninteractive
   3 |     
--------------------
ERROR: failed to solve: failed to fetch anonymous token: unexpected status: 403 Forbidden
```

We either need to do this incrementally (IE, deploy once then have our builds rely on that, in which case our builds would be out-of-sync with the base images, which would be bad), or need to build and stage the image from the `build-base` (which I'd need to figure out how to do).

Emilgardis:
I feel like there should be a GH action already made for saving an image to artifacts and the loading it

Alexhuszagh:
Will still fail, just testing saving to tarball.

Alexhuszagh:

Alexhuszagh:

.github/workflows/ci.yml
```diff
@@ -109,10 +109,109 @@ jobs:
           COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
           COMMIT_AUTHOR: ${{ github.event.head_commit.author.username }}
 
+  build-base:
```
Alexhuszagh, line 112: This also needs to have our conclusion job depend on the build-base job.

.github/workflows/ci.yml
```diff
@@ -109,10 +109,109 @@ jobs:
...
+      - uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.image }}-image-tarball
+          path: ${{ matrix.image }}-${{ env.GITHUB_REF_NAME }}.tar
+
+      - uses: ./.github/actions/cargo-install-upload-artifacts
+        if: matrix.deploy
+        with:
+          target: ${{ matrix.image }}
+          image: ${{ steps.build-docker-image.outputs.image }}
```
Emilgardis, line 190: this is not needed, always skipped here and is only used to deploy binary to release

Emilgardis (commented):
I think I personally prefer having these as

```
ghcr.io/cross-rs/base:ubuntu-vxxx
```

less "waste" so to say

.github/workflows/ci.yml
```diff
@@ -109,10 +109,108 @@ jobs:
...
+        env:
+          IMAGE: ${{ matrix.image }}
+          SUB: ${{ matrix.sub }}
+          LABELS: ${{ steps.docker-meta.outputs.labels }}
+          LATEST: ${{ needs.check.outputs.is-latest || 'false' }}
+        shell: bash
+
+      - name: Save Docker Image
+        id: save-docker-image
+        run: docker save "${IMAGE}:${TAG}" -o "${IMAGE}-${TAG}-${SUB}.tar"
```
Alexhuszagh, line 179: Our `docker load` step always needs to load these as `main`, since that's what we use for our base images:

```Dockerfile
FROM ghcr.io/cross-rs/base:emscripten-main
```

We technically don't have to deploy these images, but it might be nice for people trying to build new images based off this logic locally, especially cross maintainers.

Alexhuszagh:
bors try --target x86_64-unknown-linux-gnu --target x86_64-unknown-linux-gnu.centos --target wasm32-unknown-emscripten

Alexhuszagh:
bors try --target x86_64-unknown-linux-gnu --target x86_64-unknown-linux-gnu.centos --target wasm32-unknown-emscripten

Alexhuszagh:

Alexhuszagh:

Alexhuszagh:
pain

Alexhuszagh:
why is download-artifact producing a directory? what?

Alexhuszagh:
pain

Alexhuszagh:
It seems that the path is the directory for downloading artifacts, not the file name... This should be fixable.

```bash
$ ls -la base.tar
drwxr-xr-x  2 runner docker       4096 Oct  4 19:30 .
drwxr-xr-x 15 runner docker       4096 Oct  4 19:30 ..
-rw-r--r--  1 runner docker 1001731072 Oct  4 19:30 base-main-ubuntu.tar
```

Alexhuszagh:

Alexhuszagh:
bors try-

Alexhuszagh:

Alexhuszagh:
bors try-

Alexhuszagh:
Is.... the docker images within the same job not preserved between steps?

https://github.com/cross-rs/cross/actions/runs/3185987473/jobs/5196287429#step:12:14

**Load Base Image**

```
Loaded image: ghcr.io/cross-rs/base:main-ubuntu
```

**Build Docker Image**

```
#3 [internal] load metadata for ghcr.io/cross-rs/base:main-ubuntu
#3 ERROR: failed to authorize: failed to fetch anonymous token: unexpected status: 403 Forbidden
------
 > [internal] load metadata for ghcr.io/cross-rs/base:main-ubuntu:
------
Dockerfile.x86_64-unknown-linux-gnu:1
--------------------
   1 | >>> FROM ghcr.io/cross-rs/base:main-ubuntu
   2 |     ARG DEBIAN_FRONTEND=noninteractive
   3 |     
--------------------
ERROR: failed to solve: failed to fetch anonymous token: unexpected status: 403 Forbidden
```

I guess I can convert load and build to a single step then.

Emilgardis:
No it 100% should work, otherwise we wouldnt be able to do our tests

This is the issue I had with buildx, you need to specify the output 

https://github.com/cross-rs/cross/blob/4facf8f14b5c46869b1548ad43b81e8b50642ec1/src/docker/custom.rs#L157

Alexhuszagh:
> No it 100% should work, otherwise we wouldnt be able to do our tests
I actually found it, it's the `--pull` flag where there's no registry entry. Since we never need this in CI, I've added an environment variable to disable it and set it to true in our CI.

Alexhuszagh:

Emilgardis:
thought it could be this https://github.com/cross-rs/cross/pull/817#issuecomment-1184920948

but maybe not.

Emilgardis:
We might want to use https://github.com/containers/skopeo#copying-images

PR #959, opened 2022-07-18
Support `x86_64-unknown-linux-gnu` with a runner using Intel SDE emulation.

### LINKED ISSUE #159: Support x86_64-unknown-linux-gnu using Intel sde emulation
See https://github.com/gnzlbg/stdsimd/commit/93cc2500b0ca8d55e3ad08b0c25225f1f50ba496

Alexhuszagh:
Closes #159.

### DIFF
--- a/.changes/959.json
+++ b/.changes/959.json
@@ -1,0 +1,5 @@
+{
+    "description": "support `x86_64-unknown-linux-gnu` with a runner using Intel SDE emulation.",
+    "type": "added",
+    "issues": [159]
+}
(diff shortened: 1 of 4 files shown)


Alexhuszagh:
I'm not sure if this should be done in cross or in `cross-toolchains`, since this technically isn't a target and I don't believe Intel's SDE emulator works on anything besides x86_64, which would interfere with our cross-platform logic. I'm happy to move this over there and close this if desired.

Emilgardis:
I think this fits better in `cross-toolchains`, however we could build the image for consumption instead of having the user do it.

I'm fine with either alternative though

Alexhuszagh:
The layers are almost entirely shared with the default `x86_64-unknown-linux-gnu`, just without the Linux image and Qemu build, and the last step is just downloading and unpacking a single file, so I don't think it would be that prohibitive to let the user just make it. Even if the default `x86_64-unknown-linux-gnu` isn't downloaded, the longest build step is `xargo`, which really isn't that big of a deal.

I'll move it over there.

PR #251, opened 2019-01-19
Support mounting on SELinux

cyplo:
This should allow `cross` to work on both non-SELinux enabled systems
and SELinux ones.

I've tested on Fedora 29 for SELinux system.

Fixes #112 that was originally reported from Fedora as well I believe.

PR #962, opened 2022-07-18
Fix SELinux labels to allow shared use.

### LINKED ISSUE #961: Fix SELinux Labels
According to the documentation for Docker, SELinux labels should be used carefully, and since we mount much more than just the project, we're using the wrong SELinux labels:

> If you use selinux you can add the z or Z options to modify the selinux label of the host file or directory being mounted into the container. This affects the file or directory on the host machine itself and can have consequences outside of the scope of Docker.
> 
> - The z option indicates that the bind mount content is shared among multiple containers.
> - The Z option indicates that the bind mount content is private and unshared.
> 
> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

We use the `Z` labels by default, which seems like it could deadlock the system or cause other issues. This seems to relate to #251, which means we might need to use the `z` option or some other option. This was discovered while debugging #496. This doesn't affect Windows due to file locks, but seems to deadlock on Linux. I use Fedora with SELinux by default on the enforcing setting, so I should be able to test these changes.

Alexhuszagh:
Ensure that the volumes are not mounted as private, unshared volumes since we might mount with the host filesystem. This also fixes permissions issues with reading data from a mounted volume using a rootless container engine.

Fixes a bug introduced in #251.
Closes #961.

This is because the `Z` SELinux label assumes the data is not shared between containers and not being used by the host, as documented below:

> If you use selinux you can add the z or Z options to modify the selinux label of the host file or directory being mounted into the container. This affects the file or directory on the host machine itself and can have consequences outside of the scope of Docker.
> 
> - The z option indicates that the bind mount content is shared among multiple containers.
> - The Z option indicates that the bind mount content is private and unshared.
> 
> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

Prior to this, we used the `Z` label, when we should have been using the `z` label.

### DIFF
--- a/.changes/962.json
+++ b/.changes/962.json
@@ -1,0 +1,5 @@
+{
+    "description": "fix SELinux labels to allow use in multiple containers and/or the host filesystem.",
+    "type": "fixed",
+    "issues": [961]
+}
--- a/src/docker/local.rs
+++ b/src/docker/local.rs
@@ -14,7 +14,7 @@ fn mount(docker: &mut Command, host_path: &Path, absolute_path: &Path, prefix: &
     let mount_path = absolute_path.as_posix_absolute()?;
     docker.args(&[
         "-v",
-        &format!("{}:{prefix}{}", host_path.to_utf8()?, mount_path),
+        &format!("{}:{prefix}{}:z", host_path.to_utf8()?, mount_path),
     ]);
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


Alexhuszagh:
You can test the logic behind this with the following (use Fedora since it comes with SELinux enforced by default):

```bash
$ getenforce
Enforcing
# no label
$ podman run -it --rm -v "$PWD":"$PWD" -w "$PWD" ubuntu:20.04 bash -c "ls"
ls: cannot open directory '.': Permission denied
$ podman run -it --rm -v "$PWD":"$PWD":z -w "$PWD" ubuntu:20.04 bash -c "ls"
Cargo.lock  Cargo.toml  src  target
$ podman run -it --rm -v "$PWD":"$PWD":Z -w "$PWD" ubuntu:20.04 bash -c "ls"
Cargo.lock  Cargo.toml  src  target
```

Note that once the `z` or `Z` SELinux label is used on a directory, Podman or Docker will allow it to be read even if a label is not provided later.

Emilgardis (approved):
bors r+

Emilgardis:
Sorry for the lack of response, I want to do this but haven't taken the time to do so. I'm not sure what else we should include in v0.2.5 but just this. I'll make the release tomorrow

PR #963, opened 2022-07-18
Add process lock to avoid parallel cross issues.

### LINKED ISSUE #496: Error when building in parallel: `sh: 1: cargo: not found`
When running `cross` in multiple instances at once, the following error shows up:
```
sh: 1: cargo: not found
```

If this is too difficult to fix, there could be an improved error, or a warning.

Alexhuszagh:
Docker can have deadlocks or similar issues when invoked multiple times
in rapid succession, so this implements a small and correct process lock
with an identical API to `named_lock`. This allows us to prevent
invoking Docker multiple times.

The implementation uses `flock` on Linux/UNIX-like systems, which on
Linux is guaranteed to be cleaned up when the process exits. On Windows, it uses `CreateMutexW`, which also guarantees to be cleaned up when the process exits.

We use this rather than `fd-lock` because `fd-lock` has more dependencies, and it also uses `LockFile` on Windows which is not guaranteed to be cleaned up in a timely manner. The API is also designed for file locks, not process locks, which is only a good abstraction on UNIX-like systems.

We use this rather than `named_lock` since it has numerous, unnecessary dependencies. This allows us to have no external dependencies besides libc and the Windows API, for an extremely lightweight locking library.

Fixes #496.

### DIFF
--- a/.changes/963.json
+++ b/.changes/963.json
@@ -1,0 +1,5 @@
+{
+    "description": "add lock to prevent deadlock with docker when invoked multiple times.",
+    "type": "fixed",
+    "issues": [496]
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,6 +196,7 @@ dependencies = [
  "serde_json",
  "shell-escape",
  "shell-words",
+ "small-lock",
  "tempfile",
  "thiserror",
  "toml",
@@ -469,6 +470,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "num_cpus"
+version = "1.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "19e64526ebdee182341572e50e9ad03965aa510cd94427a4549448f285e957a1"
+dependencies = [
+ "hermit-abi",
+ "libc",
+]
+
+[[package]]
 name = "object"
 version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


Alexhuszagh:
This is still a draft since it doesn't handle SIGKILL at all, which would lead to indefinite blocking, with a file that isn't transparent at all. This likely needs to be improved or refactored dramatically to ensure it still works even after SIGKILL.

Emilgardis:
this adds a fair bit of dependencies, is there another (simpler) crate for this?

Alexhuszagh:
> this adds a fair bit of dependencies, is there another (simpler) crate for this?

I'll see, we probably could use parking_lot directly. Or we could just document this relatively prominently. We could also just use this on Unix-only, where there's significantly less dependencies, since I believe due to how Windows works, this is unlikely to be an issue.

Or, we could probably just use `libc` directly and spin this out into a simple crate (since it has general purpose): using `flock` and `CreateMutexW` alone should be fairly doable, if we don't need to worry about any other dependencies.

Alexhuszagh:
I've changed it to use `CreateMutexW` and `flock` directly. However, it is worth looking over this code carefully (I'll be running a few Miri checks on it, and likely have a data directory with UTF-8 encoded data to test my conversion function) since it requires unsafe code due to using the Windows API. Specifically, this function is the dangerous one:

```rust
/// convert a string to an owned wide string
fn to_wide_string(s: &str) -> Result<Vec<WCHAR>> {
    // SAFETY: safe since we determine the length, then get the extracted string
    unsafe {
        // on Windows, you convert to a wide string from UTF-8 in 2 steps
        // first, you calculate the length using a 0-length size for the
        // out string. then, using the length, you create a buffer of sufficient
        // length and convert the string. the string must be nul-terminated.

        // we first need to get a null-terminated string, since our wstr must be.
        let mut bytes = s.as_bytes().to_vec();
        bytes.push(b'\0');
        let ptr = bytes.as_ptr() as *const i8;
        let len = match MultiByteToWideChar(CP_UTF8, 0, ptr, -1, ptr::null_mut(), 0) {
            i if i <= 0 => return Err(get_last_error()),
            i => i as usize,
        };
        let mut vec = vec![0; len];
        let converted = MultiByteToWideChar(CP_UTF8, 0, ptr, -1, vec.as_mut_ptr(), len as _);
        if converted != len as i32 {
            return Err(get_last_error());
        }
        Ok(vec)
    }
}
```

PR #967, opened 2022-07-23
use new color-eyre function to optionally display location section

Emilgardis:


Emilgardis:
why is changelog check not working?

Alexhuszagh:
> why is changelog check not working? @Alexhuszagh

This is definitely a bug, let me check.

Alexhuszagh:
Seems like the `readarray` steps are failing and reading 1 array element when are none are present, so it never notices missing changes. It seems like `jq` always has an input item, which seems like an issue.

```bash
# no single quotes
$ readarray -t removed <<<"$(jq -r '.[]' <<< [])"
$ removed_count=${#removed[@]}
$ echo $removed_count
1

# with single quotes
$ readarray -t removed <<<"$(jq -r '.[]' <<< '[]')"
$ removed_count=${#removed[@]}
$ echo $removed_count
1
```

Alexhuszagh (approved):

PR #968, opened 2022-07-23
introduce verbosity levels

Emilgardis:
this will make it so rustup doesn't show verbose unless `-vv` or `-v` and `CROSS_DEBUG` is used.

### DIFF
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -39,7 +39,7 @@ fn rustup_command(msg_info: &mut MessageInfo, no_flags: bool) -> Command {
         Verbosity::Quiet => {
             cmd.arg("--quiet");
         }
-        Verbosity::Verbose => {
+        Verbosity::Verbose(2..) => {
             cmd.arg("--verbose");
         }
         _ => (),
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -15,7 +15,7 @@ static WORKSPACE: OnceCell<PathBuf> = OnceCell::new();
 /// Returns the cargo workspace for the manifest
 pub fn get_cargo_workspace() -> &'static Path {
     let manifest_dir = env!("CARGO_MANIFEST_DIR");
-    let mut msg_info = crate::shell::Verbosity::Verbose.into();
+    let mut msg_info = crate::shell::Verbosity::Verbose(2).into();
     #[allow(clippy::unwrap_used)]
     WORKSPACE.get_or_init(|| {
         crate::cargo_metadata_with_args(Some(manifest_dir.as_ref()), None, &mut msg_info)
--- a/xtask/src/util.rs
+++ b/xtask/src/util.rs
@@ -20,7 +20,7 @@ pub fn get_cargo_workspace() -> &'static Path {
         cross::cargo_metadata_with_args(
             Some(manifest_dir.as_ref()),
             None,
-            &mut MessageInfo::create(true, false, None).expect("should not fail"),
+            &mut MessageInfo::create(2, false, None).expect("should not fail"),
         )
         .unwrap()
         .unwrap()
@@ -274,7 +274,7 @@ mod tests {
     fn check_ubuntu_base() -> cross::Result<()> {
         // count all the entries of FROM for our images
         let mut counts = BTreeMap::new();
-        let mut msg_info = Verbosity::Verbose.into();
+        let mut msg_info = Verbosity::Verbose(2).into();
         let dockerfiles = read_dockerfiles(&mut msg_info)?;
         for (path, dockerfile) in dockerfiles {
             let lines: Vec<&str> = dockerfile.lines().collect();
--- a/.changes/968.json
+++ b/.changes/968.json
@@ -1,0 +1,4 @@
+{
+    "description": "only print rustup --verbose if `-vv` or `CROSS_VERBOSE=1` is used",
+    "type": "fixed"
+}
--- a/src/bin/cross.rs
+++ b/src/bin/cross.rs
@@ -18,7 +18,11 @@ pub fn main() -> cross::Result<()> {
     let target_list = rustc::target_list(&mut Verbosity::Quiet.into())?;
     let args = cli::parse(&target_list)?;
     let subcommand = args.subcommand;
-    let mut msg_info = shell::MessageInfo::create(args.verbose, args.quiet, args.color.as_deref())?;
+    let mut msg_info = shell::MessageInfo::create(
+        args.verbose + cross::config::bool_from_envvar("CROSS_DEBUG") as u8,
+        args.quiet,
+        args.color.as_deref(),
+    )?;
     let status = match cross::run(args, target_list, &mut msg_info)? {
         Some(status) => status,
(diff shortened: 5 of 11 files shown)


src/tests.rs
```diff
@@ -15,7 +15,7 @@ static WORKSPACE: OnceCell<PathBuf> = OnceCell::new();
 /// Returns the cargo workspace for the manifest
 pub fn get_cargo_workspace() -> &'static Path {
     let manifest_dir = env!("CARGO_MANIFEST_DIR");
-    let mut msg_info = crate::shell::Verbosity::Verbose.into();
+    let mut msg_info = crate::shell::Verbosity::Verbose(2).into();
```
Alexhuszagh, line 18: We actually probably can just use the verbosity level 1 for this. Cargo just ignores the verbosity except for a warning about the metadata version (if quiet).

xtask/src/target_info.rs
```diff
@@ -63,7 +63,7 @@ fn image_info(
     command.arg(image);
     command.args(&["bash", "-c", TARGET_INFO_SCRIPT]);
     command
-        .run(msg_info, msg_info.is_verbose())
+        .run(msg_info, !msg_info.is_verbose())
```
Alexhuszagh, line 66: Why was this changed? If you want it only to use the verbosity at 2+, then that should be done, but I think it should be used if any verbosity level is set.

xtask/src/util.rs
```diff
@@ -20,7 +20,7 @@ pub fn get_cargo_workspace() -> &'static Path {
         cross::cargo_metadata_with_args(
             Some(manifest_dir.as_ref()),
             None,
-            &mut MessageInfo::create(true, false, None).expect("should not fail"),
+            &mut MessageInfo::create(2, false, None).expect("should not fail"),
```
Alexhuszagh, line 23: Once again, this can just use the normal `Verbose(1)` level, since the only difference will be if we print the command. We also ignore the level in `run_and_get_output`, etc.

src/rustup.rs
```diff
@@ -39,7 +39,7 @@ fn rustup_command(msg_info: &mut MessageInfo, no_flags: bool) -> Command {
         Verbosity::Quiet => {
             cmd.arg("--quiet");
         }
-        Verbosity::Verbose => {
+        Verbosity::Verbose(2..) => {
```
Alexhuszagh, line 42: Is this the desired behavior? I'd expect all subcommands would print verbose output if I ran `cross --verbose` or something similar.

Alexhuszagh:
It might also be worth linking this with #797, and adding the diagnostics using `msg_info.note` or something similar. Something like:

```diff
diff --git a/src/docker/mod.rs b/src/docker/mod.rs
index cd80b20..237fc80 100644
--- a/src/docker/mod.rs
+++ b/src/docker/mod.rs
@@ -51,6 +51,7 @@ pub fn run(
             1,
         );
     }
+    msg_info.note(format_args!("running {} subcommand in container.", options.cargo_variant.to_str()))?;
     if options.is_remote() {
         remote::run(options, paths, args, msg_info).wrap_err("could not complete remote run")
     } else {
diff --git a/src/docker/remote.rs b/src/docker/remote.rs
index 23215f6..2349309 100644
--- a/src/docker/remote.rs
+++ b/src/docker/remote.rs
@@ -984,6 +984,7 @@ pub(crate) fn run(
     };
     let mount_prefix_path = mount_prefix.as_ref();
     if let VolumeId::Discard = volume {
+        msg_info.note("copying rust toolchain to temporary data volume.")?;
         copy_volume_container_xargo(engine, &container, dirs, mount_prefix_path, msg_info)
             .wrap_err("when copying xargo")?;
         copy_volume_container_cargo(engine, &container, dirs, mount_prefix_path, false, msg_info)
@@ -999,6 +1000,7 @@ pub(crate) fn run(
         .wrap_err("when copying rust")?;
     } else {
         // need to copy over the target triple if it hasn't been previously copied
+        msg_info.note(format_args!("copying rust target toolchain \"{}\" to data volume if it was not previously copied.", target.target()))?;
         copy_volume_container_rust_triple(
             engine,
             &container,
@@ -1027,6 +1029,7 @@ pub(crate) fn run(
         )
         .wrap_err("when creating mount root")?;
     }
+    msg_info.note("copying project and mounted volume files to data volume.")?;
     copy_volume_container_project(
         engine,
         &container,
diff --git a/src/rustup.rs b/src/rustup.rs
index 2b47a4d..27fdd62 100644
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -148,8 +148,9 @@ fn version(msg_info: &mut MessageInfo) -> Result<Version> {
 }
 
 pub fn install_toolchain(toolchain: &QualifiedToolchain, msg_info: &mut MessageInfo) -> Result<()> {
-    let mut command = rustup_command(msg_info, false);
     let toolchain = toolchain.to_string();
+    msg_info.note(format_args!("installing \"{toolchain}\" toolchain via rustup."))?;
+    let mut command = rustup_command(msg_info, false);
     command.args(&["toolchain", "add", &toolchain, "--profile", "minimal"]);
     if version(msg_info)? >= semver::Version::new(1, 25, 0) {
         command.arg("--force-non-host");
@@ -166,6 +167,7 @@ pub fn install(
 ) -> Result<()> {
     let target = target.triple();
     let toolchain = toolchain.to_string();
+    msg_info.note(format_args!("adding \"{target}\" target for toolchain \"{toolchain}\" via rustup."))?;
     rustup_command(msg_info, false)
         .args(&["target", "add", target, "--toolchain", &toolchain])
         .run(msg_info, false)
@@ -178,6 +180,7 @@ pub fn install_component(
     msg_info: &mut MessageInfo,
 ) -> Result<()> {
     let toolchain = toolchain.to_string();
+    msg_info.note(format_args!("installing \"{component}\" component for toolchain \"{toolchain}\" via rustup."))?;
     rustup_command(msg_info, false)
         .args(&["component", "add", component, "--toolchain", &toolchain])
         .run(msg_info, false)

```

xtask/src/target_info.rs
Emilgardis: I agree, not sure why I changed this

src/rustup.rs
Emilgardis: this was basically the whole point of the pr, my expectation for `cross --verbose` is to only print the command invocations and show their outputs. Like `cargo -v`, who wont print all `rustc` invocations, but will with `-vv`

Alexhuszagh:
Removed some of my comments since the rationale for the PR was made clearer elsewhere: the goal is so every external call to a `rustup`, `cargo`, etc. command is only printed if the verbosity level is >= 2 (similar to cargo, which only displays `rustc` commands only if `-vv` or higher is provided).

Alexhuszagh (changes requested):
Just fix the one review comment above and lgtm, the other suggestions I posted can be moved to another PR.

Alexhuszagh (approved):

Emilgardis:
bors r=@Alexhuszagh

PR #971, opened 2022-07-25
Fix changelog check for no changes.

Alexhuszagh:


Alexhuszagh:
This initial changelog check [fails](https://github.com/cross-rs/cross/runs/7508276853?check_suite_focus=true#step:5:37). Then, when modified to have the no-changelog label, it [passes](https://github.com/cross-rs/cross/runs/7508292341?check_suite_focus=true).

Emilgardis (approved):
bors r+

Issue #22, opened 2016-12-30
Newb Help


drusellers:
Will this help me compile a rust binary on my Mac that I want to distribute and run on an Ubuntu server?

japaric:
Nope. `cross` only works on x86_64 Linux. You could, however, use Travis CI to build and upload binaries to GitHub releases, which then you can download and use on an Ubuntu server.

drusellers:
Ah, ok. thank you.

PR #973, opened 2022-07-27
Update images to Ubuntu 22.04 LTS

Jules-Bertholet:


Emilgardis (changes requested):
For changelog, please edit https://github.com/cross-rs/cross/blob/main/.changes/591.json, rename to `591-973.json` and change 20.04 to 22.04

Emilgardis:
readme table needs to be updated.

You can generate the table with 

`cargo xtask target-info --tag local` (all the images need to be built though beforehand)

Emilgardis:
bors try --target linux

Alexhuszagh:
Marked this for the 0.4.0 milestone since it updates glibc from 2.21 to 2.35, which isn't supported on any Debian distribution as of yes (even on the unstable release, sid).

Alexhuszagh:
Looks like the error in the images containing a Linux image for full system emulation have an unrelated error: [dropbear-2022.82](https://matt.ucc.asn.au/dropbear/releases/dropbear-2022.82.tar.bz2) has a 520 error right now. We might need to switch to another mirror if this stays for a while.

Alexhuszagh:
Now that the mirror issues have been fixed...
bors try

Emilgardis:
seems `bzip2` is no longer in default installation

Alexhuszagh:
Looks like `ncurses-base` is no longer a valid package or has no installation candidate: all the `linux-image.sh` scripts are failing with:

```text
apt-get -d --no-install-recommends download libcrypt1:mips64el busybox:mips64el dropbear-bin:mips64el libtommath1:mips64el libtomcrypt1:mips64el libgmp10:mips64el libc6:mips64el linux-image-5.10.0-16-5kc-malta:mips64el ncurses-base zlib1g:mips64el
#22 20.92 E: Can't find a source to download version '6.3-2' of 'ncurses-base:amd64'
```

Alexhuszagh:
> Looks like `ncurses-base` is no longer a valid package or has no installation candidate: all the `linux-image.sh` scripts are failing with:
> 
> ```
> apt-get -d --no-install-recommends download libcrypt1:mips64el busybox:mips64el dropbear-bin:mips64el libtommath1:mips64el libtomcrypt1:mips64el libgmp10:mips64el libc6:mips64el linux-image-5.10.0-16-5kc-malta:mips64el ncurses-base zlib1g:mips64el
> #22 20.92 E: Can't find a source to download version '6.3-2' of 'ncurses-base:amd64'
> ```

Oh, the issue is Ubuntu 22.04 added the package `ncurses-base`, which has version `6.3-2`, which is higher than the `ncurses-base` in bullseye (`6.2+20201114-2`). We should also be using bookworm or sid for this upgrade as well, since Ubuntu 22.04 is based off of those versions. We also might need to choose an exact installation candidate, similar to how we do so for conflicting packages on amd64 in `linux-image.sh`.

flxo:
Is there anything I can help here?

In the meantime `bindgen` removed [support for ancient clang versions](https://github.com/rust-lang/rust-bindgen/blob/master/CHANGELOG.md). The clang version in 18.04 is considered ancient. 
As a result `bindgen` (>=`0.60.0`) doesn't work in any of those images out of the box. Probably a solution would be to upgrade `clang` in a custom `Dockerfile` which is required anyways to install the bindgen requirements.

Alexhuszagh:
> Is there anything I can help here?
> 
> In the meantime `bindgen` removed [support for ancient clang versions](https://github.com/rust-lang/rust-bindgen/blob/master/CHANGELOG.md). The clang version in 18.04 is considered ancient. As a result `bindgen` (>=`0.60.0`) doesn't work in any of those images out of the box. Probably a solution would be to upgrade `clang` in a custom `Dockerfile` which is required anyways to install the bindgen requirements.

We use 20.04 on the latest mains and are working on a 0.3.0 release currently. I believe those should still work with bindgen.

flxo:
> > Is there anything I can help here?
> > In the meantime `bindgen` removed [support for ancient clang versions](https://github.com/rust-lang/rust-bindgen/blob/master/CHANGELOG.md). The clang version in 18.04 is considered ancient. As a result `bindgen` (>=`0.60.0`) doesn't work in any of those images out of the box. Probably a solution would be to upgrade `clang` in a custom `Dockerfile` which is required anyways to install the bindgen requirements.
> 
> We use 20.04 on the latest mains and are working on a 0.3.0 release currently. I believe those should still work with bindgen. We'd like ideally to have a release on 20.04 before migrating to 22.04, personally, and I'd like a few bug fixes in the meantime. I'll test this later today with a more recent build.

Thanks. I didn't know that there's 20.04 in between. Haven't tried but it's very likely this works.

Any plans to include the `bindgen` dependencies by default? I have multiple projects with a `Cross.toml` and Dockerfiles just for adding the `biindgen` dependencies.

Emilgardis:
> Any plans to include the `bindgen` dependencies by default? I have multiple projects with a `Cross.toml` and Dockerfiles just for adding the `biindgen` dependencies.

Not currently, the clang deps take up a lot of space.

You can skip the additional `Dockerfile`s and only use a `Cross.toml` (or even just `Cargo.toml` like [this](https://github.com/cross-rs/wiki_assets/blob/main/Configuration/Cargo.toml))

```toml
# Cross.toml

[target.TARGET]
pre-build = ["apt-get update && apt-get install --assume-yes --no-install-recommends libclang-3.9-dev clang-3.9"]
```

ShayBox:
What's the ETA on this change? currently unable to compile anything using libssl as Ubuntu 20.04 only provides libssl1 while every distro now uses libssl3, and Ubuntu 22.04 doesn't have an official package for libssl1.

Emilgardis:
> What's the ETA on this change?

there is no ETA on 0.4.0, we still have to do 0.3.0 which needs on other changes. development has slowed down a bit due to availability, but any help is greatly appreciated

pvshvp-oss:
Is there a TODO list on this?

Dylan-DPC:
I wonder if it's worth going straight to Ubuntu 24.04 LTS now instead of 22.04

pvshvp-oss:
> I wonder if it's worth going straight to Ubuntu 24.04 LTS now instead of 22.04

On a related note, I think it is totally worth having an Arch linux image that is automatically periodically updated.

polarathene:
I've not been keeping up with this project, but didn't it use old versions of Ubuntu for minimizing the glibc version to link against for broader portability?

These days you can use `cargo zigbuild` to cross compile to targets. It handles the glibc minimal version for you so that you can run on a modern distro release, and you don't have to worry about toolchains for different archs, just add the targets via rustup IIRC.

13r0ck:
I assume that by now this should be closed in favor of cross using Ubuntu 24.04?

Emilgardis:
I'll keep it open, but yes, it would probably make sense to jump directly to 24. I've. previously said though that we should release 22.04 first but I think that would not be wise anymore.

JanDiederich:
Look at https://github.com/cross-rs/cross/pull/1580, if you're still interested in this PR. It uses Ubuntu 24.04 LTS and it's working, at least for ARM64.

PR #974, opened 2022-07-27
Change `mips64-unknown-linux-gnuabi64` to hard-float.

### LINKED ISSUE #906: MIPS64 Musl Targets Assume Hard-Float but Fail with Hard-Float Toolchain
### Checklist

### Describe your issue

When compiling with external dependencies for MIPS64 (MIPS64 or MIPS64EL), the musl targets assume hard-floats, but the actual targets should used soft-floats.

The target spec shows:

```text
$ rustc +nightly --print target-features -Z unstable-options --target mips64-unknown-linux-muslabi64
Features supported by rustc for this target:
    fp64                     - Support 64-bit FP registers.
    msa                      - Mips MSA ASE.
    crt-static               - Enables C Run-time Libraries to be statically linked.

Code-generation features supported by LLVM for this target:
...
    nomadd4                  - Disable 4-operand madd.fmt and related instructions.
    nooddspreg               - Disable odd numbered single-precision registers.
    p5600                    - The P5600 Processor.
    ptr64                    - Pointers are 64-bit wide.
    single-float             - Only supports single precision float.
    soft-float               - Does not support floating point instructions.
    sym32                    - Symbols are 32 bit on Mips64.
...
```

mips64el-unknown-linux-muslabi64, mips64-unknown-linux-muslabi64

- [ ] macOS
- [ ] Windows
- [ ] Linux / BSD

- [ ] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [ ] podman

### cross version

cross 0.2.2

### Example

A sample project with the following C-dependency and Rust binary will fail to compile:

**sf.c**

```c
float truncate(float start)
{
    float first = __trunctfsf2((long double) start);
    return (float)first;
}
```

**main.rs**

```rust
fn main() {
    let truncated = unsafe {
        truncate(50.0)
    };
    println!("truncated={:?}", truncated);
}

extern "C" {
    fn truncate(_: f32) -> f32;
}
```

However, attempting to build the project causes the following error message:

```text
  = note: /usr/local/bin/../lib/gcc/mips64-linux-muslsf/9.2.0/../../../../mips64-linux-muslsf/bin/ld: warning: /target/mips64-unknown-linux-muslabi64/debug/deps/hellopp-c76d728dba2f1a1d uses -mhard-float (set by /rust/lib/rustlib/mips64-unknown-linux-muslabi64/lib/self-contained/crt1.o), /target/mips64-unknown-linux-muslabi64/debug/build/hellopp-749888b7d3fb49a8/out/libhello.a(hello.o) uses -msoft-float
```

This is related to https://github.com/rust-lang/rust/issues/50890.

Alexhuszagh:
Changes `mips64-unknown-linux-gnuabi64` to use a hard-float musl toolchain, rather than a soft-float one.

Closes #906.

### DIFF
--- a/docker/mips64-linux-muslabi64-gcc.sh
+++ b/docker/mips64-linux-muslabi64-gcc.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# this fixes an issue of missing symbols from the command lines
+# these soft-float routines are required even for hard-float targets.
+#   (strtod.lo): undefined reference to symbol '__trunctfsf2@@GCC_3.0'
+
+set -x
+set -euo pipefail
+
+main() {
+    if [[ $# -gt 0 ]]; then
+        exec mips64-linux-musl-gcc "${@}" -lgcc -static-libgcc
+    else
+        exec mips64-linux-musl-gcc "${@}"
+    fi
+}
+
+main "${@}"
--- a/.changes/974.json
+++ b/.changes/974.json
@@ -1,0 +1,11 @@
+[
+    {
+        "description": "change `mips64-unknown-linux-muslabi64` target to hard-float target.",
+        "type": "fixed",
+        "issues": [906]
+    },
+    {
+        "description": "build static libgcc and link to static libgcc for `mips64-unknown-linux-muslabi64` target.",
+        "type": "fixed"
+    }
+]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,7 +167,7 @@ jobs:
             - { target: mipsel-unknown-linux-gnu,         os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1, runners: qemu-user qemu-system }
             - { target: mips64-unknown-linux-gnuabi64,    os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1 }
             - { target: mips64el-unknown-linux-gnuabi64,  os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1, runners: qemu-user qemu-system }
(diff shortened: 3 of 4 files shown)


Alexhuszagh:
Still a draft because I'm attempting to build a fully static musl toolchain, since there's linker errors if the `-lgcc` is not provided, but this causes a symbol dependency on `/usr/lib64/libc.so.1`. This is based off of the comment [here](https://github.com/richfelker/musl-cross-make/issues/64#issuecomment-497881830). This is due to an undefined reference to `__trunctfsf2@@GCC_3.0` in strtod, needed for `iostream`, which is unusual since this is a soft-float routine. Providing it however allows the binaries to run.

docker/mips64-linux-muslabi64-gcc.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# this fixes an issue of missing symbols from the command lines
+# these soft-float routines are required even for hard-float targets.
+#   (strtod.lo): undefined reference to symbol '__trunctfsf2@@GCC_3.0'
```
Emilgardis, line 5: Should we ln this exec to make sure it's executed? Is that maybe not really needed (since cwd _shouldn't_ change)

Alexhuszagh: The `cwd` won't change from where the script is invoked, so I don't think this is an issue.

Alexhuszagh:
bors try --target mips64-unknown-linux-gnuabi64

Alexhuszagh:
Windows issue, for some reason the permissions got reset even on Git bash.

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
Currently waiting on #978.
bors r-

Alexhuszagh:
bors retry

Alexhuszagh:
bors r=Emilgardis

PR #976, opened 2022-07-27
supplement `TARGETPLATFORM` with `RUST_TOOLCHAIN`

Emilgardis:
this is not strictly needed, but could be nice

### DIFF
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -189,6 +189,11 @@ pub fn build_docker_image(
             docker_build.arg("--load");
         }
 
+        docker_build.args(&[
+            "--build-arg",
+            &format!("RUST_TOOLCHAIN={}", platform.target),
+        ]);
+
         let mut tags = vec![];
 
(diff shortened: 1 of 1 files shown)


xtask/src/build_docker_image.rs
```diff
@@ -189,6 +189,11 @@ pub fn build_docker_image(
             docker_build.arg("--load");
         }
 
+        docker_build.args(&[
+            "--build-arg",
+            &format!("RUST_TOOLCHAIN={}", platform.target),
```
Alexhuszagh, line 194: Maybe this should be `CROSS_RUST_TOOLCHAIN`?

Alexhuszagh (approved):

Alexhuszagh:
As long as the build-arg stays a build-arg and is never exported as an environment variables, this should be good.

bors r+

Alexhuszagh:
bors retry

Emilgardis:
bors cancel

after further investigation, this is unhelpful

if it's truly needed, it can be added later

Issue #902, opened 2022-07-02
`x86_64-unknown-linux-musl` links to Glibc With C++ Dependencies


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

**UPDATE**: All targets except `x86_64-unknown-linux-musl` have been patched as of #905, and the MIPS64 targets in #906.

When compiling for some musl targets, they link to dynamically to glibc rather than musl libc statically. This only occurs with C++ dependencies.  A simple test is as follows:

```bash
$ git clone https://github.com/cross-rs/rust-cpp-hello-word
$ cd rust-cpp-hello-word
$ cross build --target aarch64-unknown-linux-musl
   Compiling cc v1.0.73
   Compiling hellopp v0.1.0 (/home/ahuszagh/Desktop/cross/rust-cpp-hello-word)
    Finished dev [unoptimized + debuginfo] target(s) in 2.58s

$ file target/aarch64-unknown-linux-musl/debug/hellopp
target/aarch64-unknown-linux-musl/debug/hellopp: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, with debug_info, not stripped
```

Some targets, such as `armv7-unknown-linux-musleabihf` link properly, however, while running, they cannot find the correct dynamic library loader.

This is related to #101, although not exactly.

### What target(s) are you cross-compiling for?

aarch64-unknown-linux-musl

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [x] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [x] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [ ] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.2 (latest main)

Alexhuszagh:
So the two targets that fail are:
- `aarch64-unknown-linux-musl`
- `x86_64-unknown-linux-musl`

ARM64 musl is actually fine: it just wants to name what it links against as a glibc loader, but it only works with a musl loader. x86_64 just doesn't work, nor will it.

Emilgardis:
should we include this in 0.2.3?

Alexhuszagh:
[...]

PR #982, opened 2022-07-28
Add concept of a native Dockerfile.

Alexhuszagh:


Alexhuszagh:
The idea behind this is so we can provide native toolchains if the image platform and the cross-compilation target are the same. This simplifies our logic for say, `linux/arm64` images for an `aarch64-unknown-linux-gnu` toolchain. There's still some work to do, however:
1. Need to change `x86_64-unknown-linux-gnu` to actually work when cross-compiled.
2. Change `i686-unknown-linux-gnu` and `i586-unknown-linux-gnu` to allow use from a non-x86_64 image.
3. Test the Native toolchain on a non-x86_64 image.

Emilgardis:
I'm not sure I understand what this does and how it differs. How is it used?

Alexhuszagh:
> I'm not sure I understand what this does and how it differs. How is it used?

Say if I want to build `aarch64-unknown-linux-gnu` for an image running Ubuntu:20.04 on `linux/arm64`: I just want to install GCC, G++, glibc, and don't want a cross-compiler: I just want the same logic as the current `Dockerfile.x86_64-unknown-linux-gnu`.

docker/Dockerfile.native
```diff
@@ -0,0 +1,27 @@
...
+
+COPY dropbear.sh /
+RUN /dropbear.sh
+
+COPY linux-image.sh /
+RUN /linux-image.sh $TARGET_ARCH
+
+COPY linux-runner /
+
+ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="/linux-runner $TARGET_ARCH"
```
Emilgardis, line 27: this should use `TARGETPLATFORM`, `TARGETOS`, and/or `TARGETARCH`

Alexhuszagh: I was planning to, but then it needs to normalize those values, which is something similar to what we do in #952, since the values normalize differently than what we currently accept as values. But yes, it should be done and would simplify things a lot.

Emilgardis:
with buildkit it's possible to do conditional stages, I think that could help to not use a script file

```
# arm64-specific stage
FROM base AS build-arm64
RUN echo "This stage is used on arm64"

# amd64-specific stage
FROM base AS build-amd64
RUN echo "This stage is used on amd64 (x86)"

# common steps
FROM build-${TARGETARCH} AS build
RUN echo "This stage is used on all architectures"
```

for example

Alexhuszagh:
> with buildkit it's possible to do conditional stages, I think that could help to not use a script file
> 
> ```
> # arm64-specific stage
> FROM base AS build-arm64
> RUN echo "This stage is used on arm64"
> 
> # amd64-specific stage
> FROM base AS build-amd64
> RUN echo "This stage is used on amd64 (x86)"
> 
> # common steps
> FROM build-${TARGETARCH} AS build
> RUN echo "This stage is used on all architectures"
> ```
> 
> for example

In retrospect, I don't think we need a script (just native/non-native) unless we want to support targets where we have to build from source (which I think we can omit) or use Debian repositories only for a different architecture (specifically, MIPS).

docker/Dockerfile.x86_64-unknown-linux-gnu
```diff
@@ -10,6 +10,15 @@ RUN /cmake.sh
 COPY xargo.sh /
 RUN /xargo.sh
 
+RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
+    g++-x86-64-linux-gnu \
```
Alexhuszagh, line 14: This still hasn't been tested: I'm testing it now.

Alexhuszagh:

docker/native-linux-image.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
Alexhuszagh, line 1: The major obstacle to not having a wrapper for this is `linux/arm`, which is `armv7`, but `linux/arm/v6` is `arm`, so we can't just use `${TARGETARCH}${TARGETVARIANT}`.

docker/native-qemu.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
Alexhuszagh, line 1: This could be merged with some of the logic in #952in `base-runner.sh`.

docker/linux-image.sh
```diff
@@ -171,12 +171,14 @@ main() {
         sharutils \
         gnupg
 
-    # amd64 has conflicting versions of the packages installed, so
+    # image arch has conflicting versions of the packages installed, so
     # we need to remove the system installs later. since apt relies
     # on these packages, we need to download them and reinstall
     # using dpkg later, since we cannot redownload via apt.
+    local image_arch
+    image_arch=$(dpkg --print-architecture)
```
Emilgardis, line 179: this is unfortunate, would be nice if it could be grabbed in "build" time for the dockerfile

docker/linux-image.sh
```diff
@@ -171,12 +171,14 @@ main() {
         sharutils \
         gnupg
 
-    # amd64 has conflicting versions of the packages installed, so
+    # image arch has conflicting versions of the packages installed, so
```
Emilgardis, line 174:
```suggestion
    # conflicting versions of some packages will be installed already for the host platform, so
```

docker/linux-image.sh
```diff
@@ -171,12 +171,14 @@ main() {
...
     # we need to remove the system installs later. since apt relies
     # on these packages, we need to download them and reinstall
     # using dpkg later, since we cannot redownload via apt.
+    local image_arch
```
Emilgardis, line 178:
```suggestion
    local host_arch
```

.changes/982.json
```diff
@@ -0,0 +1,4 @@
+{
+    "type": "internal",
+    "description": "use a generic dockerfiles for when the toolchain and image platfom match."
```
Emilgardis, line 3: if we switch this to be actually used, this should be a change

Emilgardis (changes requested):
looks good, I'm just not sure I like the naming.

maybe `same-platform`?

I'd also like it if with this pr that the image is made available and used when possible.

docker/linux-image.sh
Alexhuszagh: We could run a script and export it to a file during the build.

Alexhuszagh: This could also affect #952, since we also use `dpkg --print architecture` there. It would be likely a good idea then to store that information in a file if we wish to deduplicate that effort. Maybe store the information in `/opt/image-arch`?

docker/linux-image.sh
Alexhuszagh: Should this really be `host_arch` and not `image_arch`? We could also just use `TARGETARCH` here likely, but I feel `image_arch` is slightly more specific since it clarifies this is the architecture of the generated Docker image, and not the container engine or anything else.

Emilgardis: sounds good, if it's kept though I think `dpkg_arch` makes sense, dunno why I thought `host_arch` was a good alternative

.changes/982.json
Alexhuszagh: I guess I left this as internal since it doesn't change any existing images at all, since before `aarch64-unknown-linux-gnu` and `x86_64-unknown-linux-gnu` would fail for a `linux/arm64` platform. So, I wasn't sure if this should be changed since it doesn't change anything exposed to a user (although it certainly will add new images), and everything is mostly an internal implementation detail.

docker/linux-image.sh
Alexhuszagh: I think this is actually fine, since `dpkg --print-architecture` just avoids us having to use any case statements, and since this is only done once, it's quite simple and explicit. I've changed it to `dpkg_arch` as suggested below.

Alexhuszagh: > looks good, I'm just not sure I like the naming.
> maybe same-platform?

So `Dockerfile.same-platform`? Sounds good, I was just keeping it consistent with our runner naming.

> I'd also like it if with this pr that the image is made available and used when possible.

Not sure what you mean by this?

Emilgardis: > > I'd also like it if with this pr that the image is made available and used when possible.
> 
> Not sure what you mean by this?

This pr is only a Dockerfile, it's not built and published, and not used by cross

Alexhuszagh: It is though:


```rust

pub fn build_docker_image(
    BuildDockerImage {
        ..
    }: BuildDockerImage,
    engine: &docker::Engine,
    msg_info: &mut MessageInfo,
) -> cross::Result<()> {
    let native_dockerfile = docker_root.join("Dockerfile.native").to_utf8()?.to_string();
    let mut results = vec![];
    for (platform, (target, dockerfile)) in targets
        .iter()
        .flat_map(|t| platforms.iter().map(move |p| (p, t)))
    {
        ... 

        if target.sub.is_none() && is_native(docker_platform.as_str(), target, msg_info)? {
            dockerfile = native_dockerfile.clone();
        }

        ...
    }

    ...
}

fn is_armv6() -> bool {
    cfg!(any(
        cross_triple = "arm-unknown-linux-gnueabi",
        cross_triple = "arm-unknown-linux-musleabi"
    ))
}

fn is_armv7() -> bool {
    cfg!(any(
        cross_triple = "armv7-unknown-linux-gnueabihf",
        cross_triple = "armv7-unknown-linux-musleabihf"
    ))
}

fn is_native(
    platform: &str,
    target: &crate::ImageTarget,
    msg_info: &mut MessageInfo,
) -> cross::Result<bool> {
    Ok(match target.sub {
        Some(_) => false,
        None => match (platform, target.name.as_str()) {
            ("linux/386", "i686-unknown-linux-gnu")
            | ("linux/amd64", "x86_64-unknown-linux-gnu")
            | ("linux/arm64" | "linux/arm64/v8", "aarch64-unknown-linux-gnu")
            | ("linux/ppc64le", "powerpc64le-unknown-linux-gnu")
            | ("linux/riscv64", "riscv64gc-unknown-linux-gnu")
            | ("linux/s390x", "s390x-unknown-linux-gnu") => true,
            ("linux/arm/v6", "arm-unknown-linux-gnueabi") if is_armv6() => {
                note_host_target_detection(msg_info)?;
                true
            }
            ("linux/arm" | "linux/arm/v7", "armv7-unknown-linux-gnueabihf") if is_armv7() => {
                note_host_target_detection(msg_info)?;
                true
            }
            _ => false,
        },
    })
}
```

https://github.com/Alexhuszagh/cross/blob/3920f34c0c2b2b1a3e70ad75aeb194b4433c61a2/xtask/src/build_docker_image.rs#L191-L193

https://github.com/Alexhuszagh/cross/blob/3920f34c0c2b2b1a3e70ad75aeb194b4433c61a2/xtask/src/build_docker_image.rs#L349-L374

We detect if the image architecture is the same as the architecture the container engine is running on (IE, it is not being emulated), and use `Dockerfile.native` in that case, otherwise we use the non-native image.

docker/Dockerfile.native
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:20.04
```
Emilgardis, line 1:
```suggestion
# This dockerfile is used when the target matches the images platform in `build-docker-image`
FROM ubuntu:20.04
```

.changes/982.json
Emilgardis: I had the wrong idea about how this actually works, dismiss this

xtask/src/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let target = std::env::var("TARGET").expect("rustc must set the TARGET environment variable");
+    println!("cargo:rustc-cfg=cross_triple=\"{target}\"",);
+    println!("cargo:rustc-env=CROSS_TRIPLE=\"{target}\"",);
```
Emilgardis, line 4:
```suggestion
    println!("cargo:rustc-cfg=host_triple=\"{target}\"",);
    println!("cargo:rustc-env=HOST_TRIPLE=\"{target}\"",);
```

xtask/src/build_docker_image.rs
```diff
@@ -179,7 +180,17 @@ pub fn build_docker_image(
...
         docker_build.current_dir(&docker_root);
 
-        docker_build.args(&["--platform", &platform.docker_platform()]);
+        // add our platform, and determine if we need to use a native docker image
+        let docker_platform = platform.docker_platform();
+        let mut dockerfile = dockerfile.clone();
+        docker_build.args(&["--platform", &docker_platform]);
+        docker_build.args(&[
+            "--build-arg",
+            &format!("CROSS_TRIPLE={}", env!("CROSS_TRIPLE")),
```
Emilgardis, line 189: Why is this needed?

Alexhuszagh:
Based on comments elsewhere,  I should clarify that this is not used to publish a `ghcr.io/cross-rs/native:$tag` image: it is meant so `Dockerfile.native` is used for `ghcr.io/cross-rs/x86_64-unknown-linux-gnu:$tag` on `linux/amd64` and `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:$tag` on `linux/arm64`, and that `Dockerfile.x86_64-unknown-linux-gnu` is only used if the platform is not `linux/amd64` (same with `Dockerfile.aarch64-unknown-linux-gnu` and `linux/arm64`).

When building the image, we change the Dockerfile used for the build. If the image architecture is the same as the target architecture, and the target triple is `*-linux-gnu`, then use `Dockerfile.native`. Otherwise, use the normal Dockerfile. This is required since packages like `g++` are not for `x86_64` on a `linux/arm64` platform, and packages like [g++-x86-64-linux-gnu](https://packages.ubuntu.com/jammy/g++-x86-64-linux-gnu) are not available on `linux/amd64`.

xtask/src/build_docker_image.rs
```diff
@@ -321,6 +332,51 @@ pub fn build_docker_image(
...
+            | ("linux/amd64", "x86_64-unknown-linux-gnu")
+            | ("linux/arm64" | "linux/arm64/v8", "aarch64-unknown-linux-gnu")
+            | ("linux/ppc64le", "powerpc64le-unknown-linux-gnu")
+            | ("linux/riscv64", "riscv64gc-unknown-linux-gnu")
+            | ("linux/s390x", "s390x-unknown-linux-gnu") => true,
+            ("linux/arm/v6", "arm-unknown-linux-gnueabi") if is_armv6() => {
+                note_host_target_detection(msg_info)?;
+                true
+            }
+            ("linux/arm" | "linux/arm/v7", "armv7-unknown-linux-gnueabihf") if is_armv7() => {
```
Emilgardis, line 367: this is already partly implemented with `ImagePlatform.target`

Emilgardis (changes requested):

xtask/src/build_docker_image.rs
Emilgardis: my concern is that `CROSS_TRIPLE` is the builders platform

Alexhuszagh: @Emilgardis It's used to set the environment variable for the runner, since we might want to allow running the binaries via Qemu:

```dockerfile
ENV CROSS_TARGETARCH=$TARGETARCH
ENV CROSS_TARGETVARIANT=$TARGETVARIANT
ENV CARGO_TARGET_${CROSS_TRIPLE}_RUNNER="/native-linux-runner"
```
and

```bash
#!/usr/bin/env bash

set -eo pipefail

# shellcheck disable=SC1091
. lib.sh

main() {
    local arch
    arch=$(docker_to_linux_arch "${CROSS_TARGETARCH}" "${CROSS_TARGETVARIANT}")
    exec /linux-runner "${arch}" "${@}"
}

main "${@}"

```

docker/native-linux-runner
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# shellcheck disable=SC1091
+. lib.sh
+
+main() {
```
Alexhuszagh, line 8: We might actually be able to default to the native cross runner here, since this will always be a native runner:

```bash
if [ -Z "${CROSS_RUNNER}" = "" ]; then
   export CROSS_RUNNER=native
fi
```

xtask/src/build_docker_image.rs
Emilgardis: if I run this from a aarch64 system, the `CROSS_TRIPLE` will be different, is that the intended effect?

Alexhuszagh: > my concern is that `CROSS_TRIPLE` is the builders platform

I'll try to think of a better way: this is also partially an issue with the ARM detection as well.

Alexhuszagh: > if I run this from a aarch64 system, the `CROSS_TRIPLE` will be different, is that the intended effect?

Yes but also no. The `CROSS_TRIPLE` should change based on the host architecture, but this will fail if cross-compiling `cross` or if the cross triple isn't `*-linux-gnu`. So yes, but this logic doesn't work.

Emilgardis: > I guess we could always use `&target.nam`

I think that or `platform.target` makes sense, they should both be the same.

Alexhuszagh:
Waiting on #1006 to merge, since this will be handy for native/non-native Dockerfiles for CentOS.

Alexhuszagh:
Currently adding the concept of a native CentOS image and then have a specialized x86_64 one since we now support an aarch64 CentOS image. I'll be factoring in the requested changes shortly.

docker/native-qemu.sh
Alexhuszagh: Actually, nevermind, they're pretty substantially different and this is the simplest logic available.

xtask/src/build_docker_image.rs
Alexhuszagh: I think `platform.target` is the right solution, otherwise we duplicate the work done in detecting the target platform, getting the default target, etc.

Alexhuszagh: Actually, `platform.target` doesn't work, because we actually want the target provided, so it should actually just be `target.name`, as simple as that is.

Alexhuszagh: Ok this has been solved: we should have been using `target.triple()` anyway, since we want to check if the target platform matches the Docker image platform.

xtask/src/build_docker_image.rs
Alexhuszagh: We might have an issue in `Architecture::from_target` since it doesn't differentiate ARMv6 and ARMv7 like this does, and Docker platforms support both ARMv6 and ARMv7.

Emilgardis:
How does one build the native image for other platforms?

Alexhuszagh:
> How does one build the native image for other platforms?

Just using `cargo build-docker-image aarch64-unknown-linux-gnu` if using an ARMv8 Linux or mac M1 Docker host, or by doing `cargo build-docker-image --platform=arch64-unknown-linux-gnu aarch64-unknown-linux-gnu` (on any host).

docker/linux-image.sh
Emilgardis: I misunderstood what this pr did :)

Alexhuszagh:
A re-review of this would be wonderful, since then it can form the basis of adding Alpine images as well easily which should solve #902 and #1013.

Emilgardis:
I'm not sure I agree with adding alpine as a sub target, but I'm excited to see what you have in mind :D

xtask/src/build_docker_image.rs
```diff
@@ -338,6 +355,54 @@ pub fn build_docker_image(
...
+        "arm-unknown-linux-gnueabi" | "arm-unknown-linux-musleabi"
+    )
+}
+
+fn is_armv7(target: &str) -> bool {
+    matches!(
+        target,
+        "armv7-unknown-linux-gnueabihf" | "armv7-unknown-linux-musleabihf"
+    )
+}
```
Emilgardis, line 370: could these be methods on `ImageTarget` instead?

xtask/src/build_docker_image.rs
```diff
@@ -338,6 +355,54 @@ pub fn build_docker_image(
...
+            }
+            _ => false,
+        },
+        Some(_) => false,
+    })
+}
+
+fn note_host_target_detection(msg_info: &mut MessageInfo) -> cross::Result<()> {
+    msg_info.note("using the rust target triple to determine the host triple to determine if the docker platform is native. this may fail if cross-compiling xtask.")
+}
```
Emilgardis, line 404: this function should probably be declared in the function that uses it

xtask/src/build_docker_image.rs
```diff
@@ -183,7 +183,24 @@ pub fn build_docker_image(
...
-        docker_build.args(["--platform", &platform.docker_platform()]);
+        // add our platform, and determine if we need to use a native docker image
+        let docker_platform = platform.docker_platform();
+        let mut dockerfile = dockerfile.clone();
+        docker_build.args(["--platform", &docker_platform]);
+        docker_build.args([
+            "--build-arg",
+            &format!("CROSS_TARGET_TRIPLE={}", target.name),
+        ]);
+        if is_native(docker_platform.as_str(), target, msg_info)? {
```
Emilgardis, line 194: this read weird imo.

maybe: if `has_native_image()` is better

Alexhuszagh:
> I'm not sure I agree with adding alpine as a sub target, but I'm excited to see what you have in mind :D

It would only be for the native image, so `Dockerfile.alpine.native`, and not any other targets, since we have broken C++ builds for these. We can also do the Linux image similar to our CentOS images, so we'd only have to worry about `qemu.sh` and `dropbear.sh` for package installations. But we'll see, I'll propose something soon.

xtask/src/build_docker_image.rs
```diff
@@ -183,7 +183,24 @@ pub fn build_docker_image(
...
+            let dockerfile_name = match target.sub.as_deref() {
+                Some(sub) => format!("Dockerfile.native.{sub}"),
+                None => "Dockerfile.native".to_owned(),
+            };
+            let dockerfile_path = docker_root.join(dockerfile_name);
+            if dockerfile_path.exists() {
+                dockerfile = dockerfile_path.to_utf8()?.to_string();
+            }
```
Emilgardis, line 202: shouldn't this error?

xtask/src/build_docker_image.rs
```diff
@@ -183,7 +183,24 @@ pub fn build_docker_image(
         };
         docker_build.current_dir(&docker_root);
 
-        docker_build.args(["--platform", &platform.docker_platform()]);
+        // add our platform, and determine if we need to use a native docker image
```
Emilgardis, line 186: this comment should probably be moved down to the `if`

Emilgardis (changes requested):

xtask/src/build_docker_image.rs
```diff
@@ -355,25 +358,29 @@ pub fn build_docker_image(
...
 }
 
-fn is_armv7(target: &str) -> bool {
+fn is_armv7(target: &ImageTarget) -> bool {
     matches!(
-        target,
+        target.name.as_str(),
         "armv7-unknown-linux-gnueabihf" | "armv7-unknown-linux-musleabihf"
     )
 }
```
Emilgardis, line 374: My bad! I meant more that these could live in a `impl ImageTarget`

Emilgardis (approved):
Awesome! This is a great step towards more and easier platform support 🥳 

Looks good, I'm fine with the `is_*` being kept as they are right now.

xtask/src/build_docker_image.rs
Alexhuszagh: Ah that makes sense. That's a good idea, I'll move them in then.

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
bors try --target x86_64-unknown-linux-gnu.centos

Alexhuszagh:
bors try --target x86_64-unknown-linux-gnu.centos

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
bors r-
We've got a minor issue with `CROSS_TARGET_TRIPLE` being the target triple without the substitutions and uppercase. Since the runners are native, this won't show up in tests.

Alexhuszagh:
bors r=Emilgardis

docker/native-linux-runner
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# shellcheck disable=SC1091
+. lib.sh
```
Emilgardis, line 6: this is the cause for the error, this should be

```suggestion
. /lib.sh
```

docker/native-linux-image.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -x
+set -eo pipefail
+
+# shellcheck disable=SC1091
+. lib.sh
```
Emilgardis, line 7: same here

```suggestion
. /lib.sh
```

Emilgardis (changes requested):

docker/native-linux-image.sh
Alexhuszagh: This happens during the build script, so this example this is fine (since it's always run from root, like `linux-image.sh`, etc). But since we're explicit with `linux-image.sh`, no reason to not be explicit here.

Alexhuszagh: I've left this consistent with our other build scripts, like `linux-image.sh` and `common.sh`, so I've kept it as-is.

Emilgardis:
bors try --target x86_64-unknown-linux-gnu*

docker/Dockerfile.native
```diff
@@ -0,0 +1,31 @@
...
+COPY common.sh lib.sh /
+RUN /common.sh
+
+COPY cmake.sh /
+RUN /cmake.sh
+
+COPY xargo.sh /
+RUN /xargo.sh
+
+ARG TARGETARCH
```
Alexhuszagh, line 14: This is done after the shared code, which will be nice when creating a base image but also for cached layers even before then.

Emilgardis (approved):
forgot about this

looks good

bors r+

docker/lib.sh
```diff
@@ -86,3 +86,72 @@ download_gcc() {
...
+        arm)
+            case "${variant}" in
+                v6)
+                    echo "arm"
+                    ;;
+                ""|v7)
+                    echo "armv7"
+                    ;;
+                *)
+                echo "Unknown Docker image variant, got \"${variant}\"." >&2
```
Alexhuszagh, line 142: This should be indented.

Emilgardis:
bors cancel

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
This target did not change and passed the last time...
https://github.com/cross-rs/cross/actions/runs/3207655589/jobs/5242799975#step:12:146

Emilgardis:
I see no reason for why it would not work now except for some filesystem error

lets try again

bors r+

Emilgardis:
bors cancel

bors retry

Emilgardis:
bors r+

doesn't seem like it got queued https://app.bors.tech/repositories/45789

PR #983, opened 2022-07-28
Change gcc-multilib to use i686 gcc compiler.

Alexhuszagh:


Alexhuszagh:
This is necessary to allow these toolchains to be built for a non `linux/amd64` Docker image, and the specifics internally are the same, as we can see from the [package details](https://packages.ubuntu.com/focal/gcc-9-multilib):

`g++-multilib` needs `gcc-multilib` (both not supported on aarch64), which uses [libc6-dev-i386](https://packages.ubuntu.com/focal/libc6-dev-i386) for the i686 glibc. In short, this should produce identical code.

Alexhuszagh:
bors try --target i686-unknown-linux-gnu

docker/Dockerfile.i586-unknown-linux-gnu
```diff
@@ -11,12 +11,24 @@ COPY xargo.sh /
 RUN /xargo.sh
 
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-multilib
+    g++-i686-linux-gnu \
```
Alexhuszagh, line 14: i586 should actually continue to use multilib, since there is no native Linux image for it, nor support for this architecture anymore in Debian stable (it isn't even supported in oldoldstable, since it was no longer supported in stretch).

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #984, opened 2022-07-28
Add Linux images for `riscv64gc-unknown-linux-gnu`.

Alexhuszagh:


Alexhuszagh:
As of sid, Debian provides a Linux image for RISCV64. As of bullseye, Debian provides a Linux image for ARMv6 Raspberry PIs (little-endian, Raspberry Pi and Pi Zero). Since this doesn't affect anything besides full system emulation, the latter won't break any existing code, and the primary demand for ARMv6 targets is for RPi0.

Alexhuszagh:
borstry --target riscv64gc-unknown-linux-gnu arm-unknown-linux-gnueabi

Alexhuszagh:
bors try --target riscv64gc-unknown-linux-gnu arm-unknown-linux-gnueabi

Alexhuszagh:
Looks like we need to change the devices, likely something to:

```bash
    armv7)
        opt="-machine virt"
        driver9p="virtio-9p-device"
        drivernet="virtio-net-device"
        ;;
```

Alexhuszagh:
Removed the Linux image for `arm-unknown-linux-gnueabi` since we could only do it for a very specific chip, otherwise Qemu would never boot.
bors try --target riscv64gc-unknown-linux-gnu

docker/linux-runner
```diff
@@ -140,12 +143,16 @@ esac
...
         ${QEMU_CMD} >> "${LOG}" 2>&1 &
     fi
+    qemu_pid=$!
 
     # wait for dropbear
     for _ in $(seq 240); do
         if grep -q "Not backgrounding" "${LOG}"; then
             READY=1
             break
+        elif ! (ps -p "${qemu_pid}" >/dev/null 2>&1); then
```
Alexhuszagh, line 153: This fixes an issue where if `QEMU_CMD` fails early for any reason, then we will attempt to boot via dropbear for at least 120s. With checking if the process ID still exists for the Qemu command, we can exit early if the command fails.

Emilgardis (approved):
bors r+

PR #989, opened 2022-08-01
added default nix_store value

IllustratedMan-code:
This pull request fixes the nix related issues described in #260. Its super simple, I just made the nix_store match statement default to `/nix/store` if the NIX_STORE variable is not set. 

There might be some edge cases where this won't fix the problem, but should be fine for any normal NixOS installation.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -305,7 +305,9 @@ impl Directories {
             Some(store) => {
                 eyre::bail!("unable to find provided nix-store directory {store:?}");
             }
-            None if cfg!(target_os = "linux") && default_nix_store.exists() => Some(default_nix_store),
+            None if cfg!(target_os = "linux") && default_nix_store.exists() => {
+                Some(default_nix_store)
+            }
             None => None,
         };
 
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -294,7 +294,7 @@ impl Directories {
         let xargo = file::canonicalize(&xargo)?;
         let nix_store = match nix_store {
             Some(store) => Some(file::canonicalize(&store)?),
-            None => None,
+            None => Some(PathBuf::from(r"/nix/store")),
         };
 
         let cargo = mount_finder.find_mount_path(cargo);
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -270,7 +270,10 @@ impl Directories {
         let cargo = home::cargo_home()?;
         let xargo =
             env::var_os("XARGO_HOME").map_or_else(|| home_dir.join(".xargo"), PathBuf::from);
-        let nix_store = env::var_os("NIX_STORE").map(PathBuf::from);
+        // NIX_STORE_DIR is an override of NIX_STORE, which is the path in derivations.
+        let nix_store = env::var_os("NIX_STORE_DIR")
+            .or_else(|| env::var_os("NIX_STORE"))
+            .map(PathBuf::from);
         let target = &metadata.target_directory;
 
         // create the directories we are going to mount before we mount them,
@@ -292,9 +295,18 @@ impl Directories {
         // directories after failed canonicalization into a shared directory.
         let cargo = file::canonicalize(&cargo)?;
         let xargo = file::canonicalize(&xargo)?;
+
+        let default_nix_store = PathBuf::from("/nix/store");
         let nix_store = match nix_store {
-            Some(store) => Some(file::canonicalize(&store)?),
-            None => Some(PathBuf::from(r"/nix/store")),
+            Some(store) if store.exists() => {
+                let path = file::canonicalize(&store)?;
+                Some(path)
+            }
+            Some(store) => {
+                eyre::bail!("unable to find provided nix-store directory {store:?}");
+            }
+            None if cfg!(target_os = "linux") && default_nix_store.exists() => Some(default_nix_store),
+            None => None,
         };
 
         let cargo = mount_finder.find_mount_path(cargo);
--- a/.changes/989.json
+++ b/.changes/989.json
@@ -1,0 +1,5 @@
+{
+    "type": "changed",
+    "description": "add default nix_store value to solve nix-related issues",
+    "issues": [260]
+}


otavio (approved):

IllustratedMan-code:
Thanks @otavio for the approval! Let me know if there is anything else I need to do to get this merged.

src/docker/shared.rs
```diff
@@ -294,7 +294,7 @@ impl Directories {
         let xargo = file::canonicalize(&xargo)?;
         let nix_store = match nix_store {
             Some(store) => Some(file::canonicalize(&store)?),
-            None => None,
+            None => Some(PathBuf::from(r"/nix/store")),
         };
```
Alexhuszagh, line 298: First, a minor nit: this doesn't need to be a raw string. Also, I think we should only provide a default if the host OS is Linux. Something like:

```rust
let nix_store = match nix_store {
    Some(store) => Some(file::canonicalize(&store)?),
    None if cfg!(target_os = "linux") => Some(PathBuf::from("/nix/store")),
    None => None,
};
```

Alexhuszagh (changes requested):
This also needs a changelog entry so we can auto-generate the changelog, see [here](https://github.com/cross-rs/cross/blob/main/.changes/554.json) for an example. The filename should be PR number (`989.json`), the type should be `changed`.

The description is also required, and the `issues` key is optional, but should likely reference #260.

Alexhuszagh:
@otavio You'd know better than me, but `NIX_STORE` isn't the normal environment variable to locate the `nix-store` location, right? Isn't it [NIX_STORE_DIR](https://nixos.org/manual/nix/stable/command-ref/env-common.html#env-NIX_STORE_DIR)? Should we soft-deprecate `NIX_STORE` (IE, provide a warning if it's used, and suggests `NIX_STORE_DIR`, replace it with `NIX_STORE_DIR`, and only provide a default for the latter?

src/docker/shared.rs
Alexhuszagh: Also, the following needs to occur after providing the default if we do modify this:

```
        if let Some(ref nix_store) = nix_store {
            file::create_dir_all(nix_store)?;
        }
```

This might be an issue, since if it's not as root it will fail, and then we can't mount the nix store. IE, we can only bind-mount it if we have permissions to write there as the current user.

otavio:
@Alexhuszagh also, in Nix Flake, the environment isn't exported as it is pure.

src/docker/shared.rs
Emilgardis: I think it makes sense to check for aliveness and only specify a -v mount if the folder exists

Alexhuszagh:
> @Alexhuszagh also, in Nix Flake, the environment isn't exported as it is pure.

Any good way to detect the nix store there? From reading on flakes, I see that:

> All files tracked by the version control system (e.g. git or mercurial) will be copied to the nix store when the flake is evaluated.

> Nix flakes run in pure evaluation mode, which is underdocumented.

It seems it uses the default nix store, but doesn't provide the environment variable? I might be wildly off-base however (as you've seen from prior contents, my knowledge is NixOS is relatively rudimentary). I think if that's the case, it's reasonable for users to provide the actual nix store directory within the flake.
I think it may be reasonable to expect users to provide

src/docker/shared.rs
Emilgardis: e.g, I don't want a `/nix/store` folder to be created on every system that uses cross.

otavio:
Checking if it exists might be safe enough.

src/docker/shared.rs
Alexhuszagh: Ok then probably something like this?

```rust
let nix_store = env::var_os("NIX_STORE_DIR")
    .or_else(|_| env::var_os("NIX_STORE"))
    .map(PathBuf::from);

let nix_store = match nix_store {
    Some(store) => {
        let path = file::canonicalize(&store)?;
        fs::create_dir_all(&path)?;
        Some(path)
    }
    None if cfg!(target_os = "linux") && fs::exists("/nix/store") => Some(PathBuf::from("/nix/store")),
    None => None,
};
```

This way we now support the `NIX_STORE_DIR` override, fallback to `NIX_STORE`, and only provide the default if we're on Linux and the directory exists (so we don't make a filesystem call unless required).

Alexhuszagh:
> Checking if it exists might be safe enough.

Does this look good then?

https://github.com/cross-rs/cross/pull/989#discussion_r934766483

src/docker/shared.rs
otavio: I imagine it'd work.  I wonder why we are calling the `fs::create_dir_all` and if it'd be needed in both cases to allow mounting the volume.

Alexhuszagh: > I imagine it'd work. I wonder why we are calling the `fs::create_dir_all` and if it'd be needed in both cases to allow mounting the volume.

Because we want it to fail if it doesn't exist and we don't have write permissions, since Docker will create the directory as root if it doesn't exist (or any container engine running as root). We could also just error if `NIX_STORE` or `NIX_STORE_DIR` is provided but the directory doesn't exist (and not error if an implicit `/nix/store` doesn't exist).

IllustratedMan-code: I tried this, but I am new to rust so I might be missing something. Here is the error I got:
```
error[E0425]: cannot find function `exists` in module `fs`
   --> src/docker/shared.rs:304:54
    |
304 |             None if cfg!(target_os = "linux") && fs::exists("/nix/store") => Some(PathBuf::from("/nix/store")),
    |                                                      ^^^^^^ not found in `fs`

error[E0593]: closure is expected to take 0 arguments, but it takes 1 argument
   --> src/docker/shared.rs:275:14
    |
275 |             .or_else(|_| env::var_os("NIX_STORE"))
    |              ^^^^^^^ --- takes 1 argument
    |              |
    |              expected closure that takes 0 arguments
```

Alexhuszagh: You can try:

```rust
let default_nix_store = PathBuf::from("/nix/store");

let nix_store = env::var_os("NIX_STORE_DIR")
    .or_else(|| env::var_os("NIX_STORE"))
    .map(PathBuf::from);

let nix_store = match nix_store {
    Some(store) => {
        let path = file::canonicalize(&store)?;
        fs::create_dir_all(&path)?;
        Some(path)
    }
    None if cfg!(target_os = "linux") && default_nix_store.exists() => Some(default_nix_store),
    None => None,
};
```

IllustratedMan-code: That one works!

Alexhuszagh: @IllustratedMan-code This is what I was thinking:

```diff
diff --git a/src/docker/shared.rs b/src/docker/shared.rs
index 9fb7675..09632ce 100644
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -270,7 +270,10 @@ impl Directories {
         let cargo = home::cargo_home()?;
         let xargo =
             env::var_os("XARGO_HOME").map_or_else(|| home_dir.join(".xargo"), PathBuf::from);
-        let nix_store = env::var_os("NIX_STORE").map(PathBuf::from);
+        // NIX_STORE_DIR is an override of NIX_STORE, which is the path in derivations.
+        let nix_store = env::var_os("NIX_STORE_DIR")
+            .or_else(|| env::var_os("NIX_STORE"))
+            .map(PathBuf::from);
         let target = &metadata.target_directory;
 
         // create the directories we are going to mount before we mount them,
@@ -279,9 +282,6 @@ impl Directories {
         // if it has other issues (such as permission errors).
         file::create_dir_all(&cargo)?;
         file::create_dir_all(&xargo)?;
-        if let Some(ref nix_store) = nix_store {
-            file::create_dir_all(nix_store)?;
-        }
         create_target_dir(target)?;
 
         // get our mount paths prior to canonicalizing them
@@ -294,6 +294,20 @@ impl Directories {
         let xargo = file::canonicalize(&xargo)?;
         let nix_store = match nix_store {
             Some(store) => Some(file::canonicalize(&store)?),
+            None => Some(PathBuf::from("/nix/store")),
+        };
+
+        let default_nix_store = PathBuf::from("/nix/store");
+
+        let nix_store = match nix_store {
+            Some(store) if store.exists() => {
+                let path = file::canonicalize(&store)?;
+                Some(path)
+            }
+            Some(store) => {
+                eyre::bail!("unable to find provided nix-store directory {store:?}");
+            }
+            None if cfg!(target_os = "linux") && default_nix_store.exists() => Some(default_nix_store),
             None => None,
         };
 
```

IllustratedMan-code: Looks good to me! I'll go ahead and commit a changelog file as well as the changes you suggested.

IllustratedMan-code:
Added the changelog and suggested changes

Alexhuszagh (approved):

Alexhuszagh:
Can you squash the number of commits down to 1 (see below if you haven't done this before) and this is ready for merging:

```bash
# soft-reset the repository to before the first commit was made
git reset HEAD~3
# add all files, like before, commit, and force push
git add .changes src/
git commit
git push --force
```

otavio (approved):

Emilgardis:
> Can you squash the number of commits down to 1 (see below if you haven't done this before) and this is ready for merging:

easier way is `git rebase HEAD~3 -i` and change every instruction to squash or fixup

Alexhuszagh:
bors r=Alexhuszagh,otavio

PR #991, opened 2022-08-08
modify README.md

honganan:
Do not copy the '$' symbol when copy the command


README.md
```diff
@@ -54,9 +54,7 @@ default to `docker`.
 
 ## Installation
 
-```
-$ cargo install cross --git https://github.com/cross-rs/cross
-```
+$ ```cargo install cross --git https://github.com/cross-rs/cross```
```
Emilgardis, line 57: this is technically not a valid code fence.

should be 
```
$ `cargo install cross --git https://github.com/cross-rs/cross`
```

however, I don't like that, instead I think


    ```sh
    cargo install cross --git https://github.com/cross-rs/cross`
    ```

would be good, not the added "`sh`"

Emilgardis (changes requested):
Im not sure the dollar sign is even needed, but I've never seen it included like you have here.

in my opinion, we can keep it as is (since it's a common way to do it), or remove it fully

README.md
Alexhuszagh: I like the latter: the `$` symbols are only useful when you need to show both input and output, and therefore the block isn't meant to be copied and pasted directly to a terminal.

honganan: The dollar symbol is not needed, if click the copy button will get `$ cargo install ...`, Paste it to terminal directly and execute users will caught:
```
-bash: $: command not found
```
I think the `$` symbol is just a visual guidance says to user: Yes, this is a shell command.

So, let's make a quick decision, I prefer the first one you give, However, I can accommodate, Haha

Emilgardis: Lets go with 

	  ```sh
	  cargo install cross --git https://github.com/cross-rs/cross`
	  ```

Emilgardis (changes requested):
Thanks!

No changelog is needed for this, please remove it and squash the commits into one and we can merge this!

Emilgardis:
confusing, since your main branch technically was in sync with our main branch this pr got marked as merged. A suggestion for you @honganan is to not do pr development on main/master, but instead creating new branches.


this pr is now implemented in and a duplicate of #995

honganan:
> confusing, since your main branch technically was in sync with our main branch this pr got marked as merged. A suggestion for you @honganan is to not do pr development on main/master, but instead creating new branches.
> 
> this pr is now implemented in and a duplicate of #995

Sorry I forget to create a new branch

PR #992, opened 2022-08-08
Fix cargo-bisect-rustc issues during CI.

Alexhuszagh:
Works around https://github.com/rust-lang/cargo-bisect-rustc/issues/186.

### DIFF
--- a/ci/test-bisect.sh
+++ b/ci/test-bisect.sh
@@ -37,7 +37,7 @@ cargo -V | grep 2022-06
 ' > bisect.sh
     chmod +x bisect.sh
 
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
This change will always be valid, since we can always invoke a `cargo` custom subcommand via `cargo-*`. Since our CI is currently broken until this is fixed, we can revert it if the upstream issue is fixed.

Emilgardis (approved):

Alexhuszagh:
Bisect now [passes](https://github.com/cross-rs/cross/runs/7728593220?check_suite_focus=true).

bors r=Emilgardis

PR #1003, opened 2022-09-01
Add CentOS7-compatible target for aarch64

oooohhoo:
Add Dockerfile.aarch64-unknown-linux-gnu.centos for centos7, using glibc-2.17

### DIFF
--- a/docker/aarch64-linux-gnu-glibc.sh
+++ b/docker/aarch64-linux-gnu-glibc.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+set -x
+set -euo pipefail
+
+
+main() {
+  yum -y install epel-release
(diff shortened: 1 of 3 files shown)


docker/aarch64-linux-gnu-glibc.sh
```diff
@@ -0,0 +1,93 @@
...
+  ln -s libgcc_s.so.1 $prefix/lib/libgcc_s.so
+  cd ..
+
+  popd
+
+  rm -rf "${td}"
+
+}
+
+main "${@}"
```
Emilgardis, line 93: please add a trailing newline here

Emilgardis (changes requested):
please add the target to the CI and then run `cargo xtask codegen`

if you need help I can guide how to add it!

PR #1005, opened 2022-09-01
Add CentOS7-compatible target for aarch64

oooohhoo:
Add Dockerfile.aarch64-unknown-linux-gnu.centos for centos7, using glibc-2.17

### DIFF
--- a/.changes/1005.json
+++ b/.changes/1005.json
@@ -1,0 +1,4 @@
+{
+    "type": "added",
+    "description": "add CentOS7-compatible target for aarch64"
+}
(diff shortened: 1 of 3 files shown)


Emilgardis:
it's the force-pushes that make github not happy, since the branch is the default branch. On a non-default branch it'd be fine.

```
git checkout -b aarch64-centos
```

Emilgardis:
To add the image to ci, edit

https://github.com/cross-rs/cross/blob/47df5c76e7cba682823a0b6aa6d95c17b31ba63a/.github/workflows/ci.yml#L151-L212


and add 

```yml
- { target: aarch64-unknown-linux-gnu, sub: centos, os: ubuntu-latest,  cpp: 1, dylib: 1, std: 1, run: 1, runners: native qemu-user qemu-system }
```

then run `cargo xtask codegen`

oooohhoo:
> it's the force-pushes that make github not happy, since the branch is the default branch. On a non-default branch it'd be fine.
> 
> ```
> git checkout -b aarch64-centos
> ```

Thanks for advice. I'll checkout a  new branch and add the image to ci in that branch

PR #1007, opened 2022-09-02
volume improvements

Emilgardis:
image

### DIFF
--- a/src/bin/commands/containers.rs
+++ b/src/bin/commands/containers.rs
@@ -1,12 +1,10 @@
 use std::io;
 
 use clap::{Args, Subcommand};
+use cross::docker::ImagePlatform;
+use cross::rustc::{QualifiedToolchain, Toolchain};
 use cross::shell::{MessageInfo, Stream};
 use cross::{docker, CommandExt, TargetTriple};
-use cross::{
-    docker::ImagePlatform,
-    rustc::{QualifiedToolchain, Toolchain},
-};
 
 #[derive(Args, Debug)]
 pub struct ListVolumes {
@@ -105,7 +103,7 @@ pub struct CreateVolume {
     pub engine: Option<String>,
     /// Toolchain to create a volume for
     #[clap(long, default_value = TargetTriple::DEFAULT.triple(), )]
-    pub toolchain: TargetTriple,
+    pub toolchain: String,
 }
 
 impl CreateVolume {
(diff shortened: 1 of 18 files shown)


Emilgardis:
clippy now complains about out IntoIterator crimes, fixed that too

src/bin/commands/containers.rs
```diff
@@ -576,3 +568,20 @@ pub fn remove_all_containers(
...
+
+fn toolchain_or_target(
+    s: &str,
+    msg_info: &mut MessageInfo,
+) -> Result<QualifiedToolchain, color_eyre::Report> {
+    let config = cross::config::Config::new(None);
+    let mut toolchain = QualifiedToolchain::default(&config, msg_info)?;
+    let target_list = cross::rustc::target_list(msg_info)?;
+    if target_list.contains(s) {
+        toolchain.replace_host(&s.parse()?);
```
Alexhuszagh, line 580: Couldn't this just use `ImagePlatform::from_target(s.into())` or something similar? This will always be a target triple, so we don't need to attempt to parse the platform.

Emilgardis: I'm not sure I follow, we need the channel, date and host, with image platform we only get the host.

Alexhuszagh: I don't believe so, because:
- ` target_list.contains(s)` will only return true of `s` matches any triple in `target_list`.
- `rustc --print target-list` only gives us the triples.
- `ImagePlatform::parse` splits on the first `'='`, which then is split into `(platform, toolchain)`. The toolchain is converted into `TargetTriple` using `toolchain.into()`, which is always just the target triple (either as a known, pre-defined enum or an other fallthrough with a string).

In short, I don't believe in any case this can have a codepath other than `from_target`: this just adds unnecessary layers. Or, we have a logic error somewhere.

Emilgardis: `ImagePlatform::parse` splits on = with `(platform, target_triple)`, not a toolchain.

https://github.com/cross-rs/cross/blob/ef332d92e256c4ffe73343a84db27fd37762fde2/src/docker/image.rs#L179

I see that the naming is here is bad, it should really be `target`.



> In short, I don't believe in any case this can have a codepath other than `from_target`: this just adds unnecessary layers. Or, we have a logic error somewhere.

Say for example I supply `--toolchain stable-x86_64-unknown-linux-gnu`, where is `stable` stored and parsed with `ImagePlatform::from_target`?

Emilgardis: Oh, I see now what you mean, there wasn't enough context in the comment, but seeing it fully I understand now, this is in the "target" branch.

Yes I think we could use what you suggest

Emilgardis: implemented!

Alexhuszagh (approved):
bors r+
Just ran some tests to confirm the logic is sound, works as expected. Looks good.

PR #1008, opened 2022-09-03
convert and improve matrix generation with a xtask script

Emilgardis:
Allows for more fine grained try control

```
> bors try --target x86_64*linux* --cpp false
>> Building x86_64-unknown-linux-musl
```
```
OPTIONS:
        --cpp <CPP>               [possible values: true, false]
        --dylib <DYLIB>           [possible values: true, false]
    -h, --help                    Print help information
    -r, --runners <RUNNERS>...    
        --run <RUN>               [possible values: true, false]
        --std <STD>               [possible values: true, false]
    -t, --target <TARGET>...      
```

### DIFF
--- a/targets.toml
+++ b/targets.toml
@@ -1,0 +1,465 @@
+# This file contains all the "targets" cross can be used with by default and is used for generating `src/docker/provided_images.rs` and our images
+# the only required value for a `target` entry is its name in `target.target` and the `os` to use in CI
+#
+# spec is available in `xtask/src/util.rs` on `CiTarget`
+
+[[target]]
+target = "x86_64-apple-darwin"
+os = "macos-12"
+special = true
+deploy = true
+
+[[target]]
+target = "x86_64-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "native qemu-user qemu-system"
+deploy = true
+
+[[target]]
+target = "x86_64-unknown-linux-musl"
+os = "ubuntu-latest"
+std = true
+run = true
+deploy = true
+
+[[target]]
+target = "x86_64-pc-windows-msvc"
+os = "windows-2019"
+special = true
+deploy = true
+
+[[target]]
+target = "x86_64-unknown-linux-gnu"
+sub = "centos"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "native qemu-user qemu-system"
+
+[[target]]
+target = "aarch64-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "arm-unknown-linux-gnueabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "arm-unknown-linux-gnueabihf"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "armv7-unknown-linux-gnueabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user"
+
+[[target]]
+target = "armv7-unknown-linux-gnueabihf"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "thumbv7neon-unknown-linux-gnueabihf"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "i586-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "i686-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "native qemu-user qemu-system"
+
+[[target]]
+target = "mips-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user"
+
+[[target]]
+target = "mipsel-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "mips64-unknown-linux-gnuabi64"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "mips64el-unknown-linux-gnuabi64"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "mips64-unknown-linux-muslabi64"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "mips64el-unknown-linux-muslabi64"
+os = "ubuntu-latest"
+std = true
+run = true
+
+[[target]]
+target = "powerpc-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "powerpc64-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "powerpc64le-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "riscv64gc-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "s390x-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "sparc64-unknown-linux-gnu"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "aarch64-unknown-linux-musl"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "arm-unknown-linux-musleabihf"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "arm-unknown-linux-musleabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "armv5te-unknown-linux-gnueabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "armv5te-unknown-linux-musleabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "armv7-unknown-linux-musleabi"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "armv7-unknown-linux-musleabihf"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+
+[[target]]
+target = "i586-unknown-linux-musl"
+os = "ubuntu-latest"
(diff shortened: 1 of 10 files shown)


Emilgardis:
bors try --target centos --aaaa

Emilgardis:
bors try --target centos

Emilgardis:
bors try --target aaaaaaaa

Emilgardis:
bors try --target centos

targets.json
```diff
@@ -0,0 +1,446 @@
+[
```
Emilgardis, line 1: should this maybe be toml?

Emilgardis: something like that yeah, I'm not sure if I have a preference

Alexhuszagh: I think TOML's much more compact and legible in this case, so I quite like it. Since we're not doing nested keys, this is a perfect use-case for TOML.

Emilgardis: switched to toml, while documenting the fields I noticed some improvements we can do also, so that's nice

targets.toml
```diff
@@ -0,0 +1,473 @@
+# This file contains all the "targets" cross can be used with by default and is used for generating `src/docker/provided_images.rs` and our images
+# the only required value for a `target` entry is its name in `target.target` and the `os` to use in CI
+#
+# ## spec
+#
+# ### Required
+# * `target`: The name of the target. This can either be a target triple, or if the
```
Emilgardis, line 7: md wants a newline here

Emilgardis (commented):
this is nice, we can with this also be more explicit about what platforms to build for more easily

Emilgardis:
I believe this is ready

xtask/src/ci/target_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::util::{get_matrix, gha_output, gha_print, CiTarget, ImageTarget};
+
+pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Report> {
+    let mut matrix: Vec<CiTarget> = get_matrix().clone();
+    if author == "bors[bot]" && message.starts_with("Try #") {
+        if let Some((_, args)) = message.split_once(": ") {
+            let app = TargetMatrixArgs::parse_from(args.split(' '));
+            if app.target.is_empty() {
+            } else {
+                matrix.retain(|m| {
```
Alexhuszagh, line 14: Does this support wildcards? It would be nice to expand globs, if possible.

xtask/src/ci/target_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+                        .iter()
+                        .any(|t| t.parse::<ImageTarget>().unwrap() == matrix_target)
+                        || app
+                            .target
+                            .iter()
+                            .any(|t| matrix_string.contains(t.as_str()))
+                })
+            };
+            if let Some(std) = app.std {
+                matrix.retain(|m| m.std.unwrap_or_default() == std)
```
Alexhuszagh, line 28: Is this just so we can provide `bors try --std`, to try all targets supporting `std`? Same with everything else, obviously? Same with everything else including `runners`?

Emilgardis: yes, I don't see the need for it immediately but you never know.

xtask/src/ci/target_matrix.rs
Emilgardis: ah no, I forgot about that. I'm not super enthustiastic about adding regex to the deps here, but with caching I guess it's fine, our xtask is already pretty heavy

Emilgardis: and for globs specifically, it could be possible to implement with some algorithm I'm not aware of easily I suppose

xtask/src/ci/target_matrix.rs
Alexhuszagh: I think it's good then: future-proofing or unexpected needs is quite good. There might be a situation we only want targets that are `no-std`, and this would dramatically simplify the CI steps.

xtask/src/ci/target_matrix.rs
Alexhuszagh: I don't want regex itself, just glob support which I think we can do. I wonder if we can have a lightweight matcher for just the expansion of `*` characters.

Emilgardis: supporting trailing `*` would be easy

Alexhuszagh: As mentioned in Matrix, it looks like [wildmatch](https://crates.io/crates/wildmatch/2.1.1/dependencies) is a lightweight and good fit for this. We might also want some very simple tests for what the matrix matcher produces (check if some targets we'll always have are retained).

xtask/src/ci/target_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+use serde::Serialize;
+
+use crate::util::{get_matrix, gha_output, gha_print, CiTarget, ImageTarget};
+
+pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Report> {
+    let mut matrix: Vec<CiTarget> = get_matrix().clone();
+    if author == "bors[bot]" && message.starts_with("Try #") {
+        if let Some((_, args)) = message.split_once(": ") {
+            let app = TargetMatrixArgs::parse_from(args.split(' '));
+            if app.target.is_empty() {
```
Alexhuszagh, line 12: We can just do if !app.target.is_empty() and skip the empty block.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+runners = "qemu-user qemu-system"
+
+[[target]]
+target = "mips64-unknown-linux-muslabi64"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 154: This is implicit, I know it was part of the original target spec but we can probably remove this.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "mips64el-unknown-linux-muslabi64"
+os = "ubuntu-latest"
+std = true
+run = true
```
Alexhuszagh, line 160: I think this might have had C++ and dynamic library support fixed, but that's for another PR. Let me add this to my TODO list.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "i586-unknown-linux-musl"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 278: Should we also allow a native runner here? It needs `qemu-user` to ensure we use the right CPU, but it can also run natively (for whatever that's worth). It's also the default when running via `cross`, so it's probably a good idea.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "i686-unknown-linux-musl"
+os = "ubuntu-latest"
+cpp = true
+dylib = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 287: Same as for `i586-unknown-linux-musl`.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+std = true
+run = true
+
+[[target]]
+target = "aarch64-linux-android"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 311: Implicit, all Android runners default to `qemu-user`.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "arm-linux-androideabi"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 319: Same as above.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "armv7-linux-androideabi"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 327: Same as above.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "thumbv7neon-linux-androideabi"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 335: Same as above.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "i686-linux-android"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 343: Same as above.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+
+[[target]]
+target = "x86_64-linux-android"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+runners = "qemu-user"
```
Alexhuszagh, line 351: Same as above.

targets.toml
```diff
@@ -0,0 +1,460 @@
...
+std = true
+run = true
+
+[[target]]
+target = "i686-pc-windows-gnu"
+os = "ubuntu-latest"
+cpp = true
+std = true
+run = true
+
```
Alexhuszagh, line 366: We're missing the commented out `asmjs-unknown-emscripten`, which we should document why it's not enabled.

```yaml
# Disabled for now, see https://github.com/rust-lang/rust/issues/98216
#-{ target: asmjs-unknown-emscripten,         os: ubuntu-latest,  cpp: 1,           std: 1, run: 1 }
```

Alexhuszagh:
Missing a few things, but otherwise, lgtm.

targets.toml
Emilgardis: added this, we're already testing it so nothing has been missed but good to be explicit imo

Emilgardis:
bors try --target *unknown-linux-musl

Emilgardis:
bors try --target *unknown-linux-musl

Emilgardis:
I was wrong, not tested :D

Emilgardis:
bors try --std 1 --cpp true

Emilgardis:
bors try --target *dragonfly*

xtask/src/ci/target_matrix.rs
```diff
@@ -37,6 +37,7 @@ pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Rep
 }
 
 #[derive(Serialize)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
```
Emilgardis, line 40: deny here is kinda pointless

Alexhuszagh: I think this is fine: I just tried it with `--build-std` and it was properly denied.

Alexhuszagh:
bors try --target i586-unknown-linux-musl

Emilgardis:
it should work because of 6713611

xtask/src/ci/target_matrix.rs
Emilgardis: that doesn't come from here though, it comes from the deny on `CiTarget`

`deny_unknown_fields` is only applied on deserialization

Alexhuszagh:
> it should work because of [6713611](https://github.com/cross-rs/cross/commit/67136116c3d873bc50fa05aa8c0ac648ad3581eb)

We can fix this in a separate PR then. lgtm, squash the commits and I'll do one last final review before merging.

xtask/src/ci/target_matrix.rs
```diff
@@ -0,0 +1,152 @@
...
+            std: target.std.map(|b| b as u8),
+        })
+        .collect::<Vec<_>>();
+    gha_output("matrix", &serde_json::to_string(&matrix)?);
+    Ok(())
+}
+
+#[derive(Serialize)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+struct Matrix<'a> {
```
Alexhuszagh, line 41: Can we change the naming on this maybe? `Matrix` and `MATRIX` was somewhat confusing early, since the former was a specific element and the `MATRIX` was the `Vec<Matrix>`. Normally the matrix runs every combination, and we just create a single set of combinations (which absolutely works). `MatrixElement` could work, even though it's not ideal, but it's much clearer than `Matrix`. That way we don't use matrix for both a list of elements and the element itself.

I understand for serialization why this differs.

xtask/src/ci/target_matrix.rs
```diff
@@ -0,0 +1,152 @@
...
+                        .any(|t| wildmatch::WildMatch::new(t).matches(&matrix_string));
+            })
+        };
+        if let Some(std) = self.std {
+            matrix.retain(|m| m.std.unwrap_or_default() == std)
+        }
+        if let Some(cpp) = self.cpp {
+            matrix.retain(|m| m.cpp.unwrap_or_default() == cpp)
+        }
+        if let Some(dylib) = self.dylib {
```
Alexhuszagh, line 104: We should probably add a single test to ensure `dylib` or at least one of these checks works, by checking an item in `dylib` (`aarch64-unknown-linux-gnu`) and an item not in `dylib` (`thumbv6m-none-eabi`) are in and not in it respectively. Not all of these code paths need to be tested, but I think at least one should be.

xtask/src/ci/target_matrix.rs
Alexhuszagh: I don't really see an issue then, because it's better to fail fast in general, so even if this isn't needed, it might be useful after refactoring.

Alexhuszagh:
Just those 2 minor changes and it's ready for merging.

Emilgardis:
added a new flag `--none` and hopefully made coverage happy

Alexhuszagh:
hors r+

PR #1011, opened 2022-09-06
Fix documentation issues

saschagrunert:
I came across this project and found possible enhancements to the documentation:

- mention that the release binaries can be used directly
- added the `pre-build` to Cross.toml `build` section, which also works
  for all targets
- fixed the architecture in the arm64 example in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,10 @@ default to `docker`.
 cargo install cross --git https://github.com/cross-rs/cross
 ```
 
+It's also possible to directly download the pre-compiled [release
+binaries](https://github.com/cross-rs/cross/releases) or using
+[cargo-binstall](https://github.com/cargo-bins/cargo-binstall).
+
 ## Usage
 
 `cross` has the exact same CLI as [Cargo](https://github.com/rust-lang/cargo)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -58,6 +58,9 @@ default to `docker`.
 cargo install cross --git https://github.com/cross-rs/cross
 ```
 
+It's also possible to directly download the pre-compiled [release
+binaries](https://github.com/cross-rs/cross/releases).
+
```
Emilgardis, line 63: can we have a little mention to https://github.com/cargo-bins/cargo-binstall here maybe?? what do you think?

Emilgardis (approved):
Awesome, thanks for finding these improvements :D

looks good, a small comment I think we could address but otherwise this is just fine

README.md
saschagrunert: Yes why not, let's mention it as well :)

Alexhuszagh (approved):
Great finds.
bors r=Emilgardis,Alexhuszagh

Emilgardis:
lets try again

bors r=Emilgardis,Alexhuszagh

PR #1015, opened 2022-09-08
Improve error diagnostics and other changes

Emilgardis:
- don't hide hostname querying in verbose
- output cross images as json
- associate `CROSS_DEBUG` with `MessageInfo`
image
- track caller when doing debug or error prints from `MessageInfo` and print the line it comes from

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+Dockerfile.* linguist-language=Dockerfile
--- a/src/bin/commands/images.rs
+++ b/src/bin/commands/images.rs
@@ -26,6 +26,9 @@ pub struct ListImages {
     /// Container engine (such as docker or podman).
     #[clap(long)]
     pub engine: Option<String>,
+    /// Output format
+    #[clap(long, default_value = "human")]
+    pub format: OutputFormat,
     /// Only list images for specific target(s). By default, list all targets.
     pub targets: Vec<String>,
 }
@@ -33,6 +36,25 @@ pub struct ListImages {
 impl ListImages {
     pub fn run(self, engine: docker::Engine, msg_info: &mut MessageInfo) -> cross::Result<()> {
         list_images(self, &engine, msg_info)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum OutputFormat {
+    Human,
+    Json,
+}
+
+impl clap::ValueEnum for OutputFormat {
+    fn value_variants<'a>() -> &'a [Self] {
+        &[Self::Human, Self::Json]
+    }
+
+    fn to_possible_value<'a>(&self) -> Option<clap::PossibleValue<'a>> {
+        match self {
+            OutputFormat::Human => Some(clap::PossibleValue::new("human")),
+            OutputFormat::Json => Some(clap::PossibleValue::new("json")),
+        }
     }
 }
 
@@ -118,7 +140,7 @@ impl Images {
     }
 }
 
-#[derive(Debug, PartialOrd, Ord, PartialEq, Eq)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, serde::Serialize)]
 struct Image {
     repository: String,
     tag: String,
@@ -258,7 +280,9 @@ fn get_image_target(
 }
 
 pub fn list_images(
-    ListImages { targets, .. }: ListImages,
+    ListImages {
+        targets, format, ..
+    }: ListImages,
     engine: &docker::Engine,
     msg_info: &mut MessageInfo,
 ) -> cross::Result<()> {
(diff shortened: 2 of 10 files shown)


src/bin/commands/images.rs
```diff
@@ -26,6 +26,9 @@ pub struct ListImages {
     /// Container engine (such as docker or podman).
     #[clap(long)]
     pub engine: Option<String>,
+    /// output format
+    #[clap(long, default_value = "human", value_parser = clap::builder::PossibleValuesParser::new(["human", "json"]))]
+    pub output_format: String,
```
Emilgardis, line 31: output is a externally tagged on target, should it maybe be internally tagged?

```json
{
  "aarch64-linux-android": [
    {
      "repository": "ghcr.io/cross-rs/aarch64-linux-android",
      "tag": "local",
      "id": "52c5946ea8ae"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-linux-android",
      "tag": "main",
      "id": "695f693bb38d"
    }
  ],
  "aarch64-unknown-linux-gnu": [
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "0.2.2",
      "id": "7ce08a4ab4f3"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "0.2.4",
      "id": "377343ffa1f4"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "<none>",
      "id": "0575ed604a64"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "local",
      "id": "119d741c8acd"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "local-centos",
      "id": "16f7395d709f"
    },
    {
      "repository": "ghcr.io/cross-rs/aarch64-unknown-linux-gnu",
      "tag": "main",
      "id": "4b4a3b05b155"
    },
    {
      "repository": "localhost/cross-rs/cross-custom-crossissue",
      "tag": "aarch64-unknown-linux-gnu-97a4c",
      "id": "8c4c09ff7f80"
    },
    {
      "repository": "localhost/cross-rs/cross-custom-ring-sqlx",
      "tag": "aarch64-unknown-linux-gnu-c4bb0",
      "id": "d4270573e82c"
    },
    {
      "repository": "localhost/cross-rs/cross-custom-ring-sqlx",
      "tag": "aarch64-unknown-linux-gnu-d5f53",
      "id": "bcd20ef4ea80"
    }
  ]
}
```

src/extensions.rs
```diff
@@ -85,19 +92,31 @@ impl CommandExt for Command {
     }
 
     fn fmt_message(&self, mut msg_info: &mut MessageInfo) -> String {
+        use std::fmt::Write;
         let msg_info = &mut msg_info;
+        let mut string = String::new();
+        if let Some(caller) = msg_info.caller() {
+            write!(string, "[{}] + ", caller).unwrap();
+        } else {
+            write!(string, "+ ",).unwrap();
```
Emilgardis, line 101: trailing ,

src/shell.rs
```diff
@@ -364,19 +396,19 @@ impl Default for MessageInfo {
...
     fn from(verbosity: Verbosity) -> MessageInfo {
-        MessageInfo::new(ColorChoice::Auto, verbosity)
+        MessageInfo::new(ColorChoice::Auto, verbosity).cross_debug_check()
     }
 }
 
 impl From<(ColorChoice, Verbosity)> for MessageInfo {
     fn from(info: (ColorChoice, Verbosity)) -> MessageInfo {
-        MessageInfo::new(info.0, info.1)
+        MessageInfo::new(info.0, info.1).cross_debug_check()
```
Emilgardis, line 411: is this really needed? I added it just to be safe and consistent

src/bin/commands/images.rs
Alexhuszagh: I think the command name should be changed to either of the following:
- `--message-format` (consistency with `cargo build`).
- `-f`, `--format` (consistency with `docker`).

src/shell.rs
```diff
@@ -148,7 +153,15 @@ impl MessageInfo {
             verbosity,
             stdout_needs_erase: false,
             stderr_needs_erase: false,
+            cross_debug: false,
+        }
+    }
+
+    pub fn cross_debug_check(mut self) -> Self {
```
Alexhuszagh, line 160: I don't particularly like this name. It should probably be `cross_debug` and then ensure it's always called rather than the field used (with a `&mut self` signature), put into the constructor, or something similar. This I don't feel properly describes what it does.

src/shell.rs
Alexhuszagh: I think we can move this into the `new` and `create` constructors, probably with `new` requiring the `cross_debug` bool, and `create` parsing the value. Maybe there should probably be a static function to detect it.

src/bin/commands/images.rs
Alexhuszagh, line 31: Make this an enum, since we have a value parser and a list of enumerated values. No reason this should be a string.

src/bin/commands/images.rs
```diff
@@ -279,45 +286,55 @@ pub fn list_images(
...
-            for _ in 0..spaces {
-                row.push(fill);
-            }
-            row.push('|');
-            row.push(fill);
-            row.push_str(col2);
-            let spaces = max_image_len.max(col2.len()) + 1 - col2.len();
-            for _ in 0..spaces {
-                row.push(fill);
+    match &*output_format {
```
Alexhuszagh, line 289: See above: this should match an enum.

src/extensions.rs
Alexhuszagh: Is there supposed to be a trailing comma? The previous `write!` doesn't have it, but this one does.

src/shell.rs
```diff
@@ -148,19 +154,41 @@ impl MessageInfo {
...
+            .as_deref()
+            .map(bool_from_envvar)
+        {
+            self.cross_debug = Some(b);
+            b
+        } else {
+            self.cross_debug = Some(false);
+            false
         }
     }
```
Emilgardis, line 175: I think this is a better interface, no way to do it wrong accidentally. Would be nice to use `OnceCell` or `LazyLock` though

src/extensions.rs
```diff
@@ -8,8 +8,10 @@ use crate::shell::MessageInfo;
 pub const STRIPPED_BINS: &[&str] = &[crate::docker::DOCKER, crate::docker::PODMAN, "cargo"];
 
 pub trait CommandExt {
+    #[track_caller]
```
Emilgardis, line 11: I don't think this `track_caller` is needed

src/extensions.rs
```diff
@@ -8,8 +8,10 @@ use crate::shell::MessageInfo;
...
     fn fmt_message(&self, msg_info: &mut MessageInfo) -> String;
```
Emilgardis, line 13: I don't think this `#[track_caller]` is needed

Emilgardis:
this might be a bug, it doesn't work on my mac at least.

```
Error: 
   0: could not run container
   1: could not complete remote run
   2: when copying project
   3: copy of files failed

Location:
   src/docker/remote.rs:710
```

Emilgardis:
bors cancel

bad test

Emilgardis:
bors try-

Emilgardis:

I can't test this rn, so ci is my tester :)

Emilgardis:
bors try-

Emilgardis:
yep, it's wrong and replicates, opening an issue

Emilgardis:
opened #1016

src/extensions.rs
Alexhuszagh: It's can't fail so I'd agree.

src/shell.rs
Alexhuszagh: If we go with a dedicated `cross_debug`, it should belong in the constructor, rather than being lazily evaluated. In that case, a static method which is then used by the constructor is a good idea, rather than have an option and doing the above every single time we print a message. I think this should be evaluated in `create`, and required as an argument in `new`. I do like decoupling the verbosity level from the debug info, since this was different behavior from `cargo`.

src/extensions.rs
Emilgardis: It's not that it cant fail, it's that the `#[track_caller]` is what makes `std::panic::Location` return a correct value, but since all calls to `fmt_message` are behind another track_caller, we should be fine

src/shell.rs
Emilgardis: the problem is that we call `MessageInfo::new` in a few places where it doesn't really make sense to check for the env var.

I can't think of a better way to solve this. I'll implement it in the constructor and we can see if it works out or if there is a better way

Emilgardis: Yeah no it doesn't really work out, with this I have to add the env getter to all invocations of `MessageInfo::new`

Emilgardis:
bors try --none

Emilgardis:
bors try --none

Alexhuszagh (approved):
Maybe change the commit title to be more descriptive and squash the commit history, and this lgtm. Maybe something like "Improve error diagnostics and other changes."

And then also add to the PR message the addition of the caller in `MessageInfo`, in addition to the other changes.

.gitattributes
```diff
@@ -0,0 +1 @@
+Dockerfile.* linguist-language=Dockerfile
```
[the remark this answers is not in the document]
Alexhuszagh: Would be nice for text editor support, such as Sublime Text which I use. I think as long as we keep it fairly simple and consistent for the config, it's not a big deal since it's effectively just an INI/TOML file.

Issue #1017, opened 2022-09-11
arm-unknown-linux-gnueabihf, gets Illegal Instruction when run on RPI Zero W


tmayoff:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

I'm trying to cross compile for my RPI Zero W, and just get Illegal Instruction when I try to run it. I cut the code down to a simple hello world, and still get Illegal Instruction

This is the backtrace produced under gdb.
```
#0  0x0054640c in __libc_csu_init ()
#1  0xb6da06cc in __libc_start_main (main=0xbefff654, argc=-1225945088, argv=0xb6da0690 <__libc_start_main+100>, init=0x54640d <__libc_csu_init>, fini=0x54644d <__libc_csu_fini>, rtld_fini=0xb6fdd510 <_dl_fini>, stack_end=0xbefff654)
    at libc-start.c:264
#2  0x00406444 in _start () at ../ports/sysdeps/arm/start.S:110
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```
Any help would be appreciated

### What target(s) are you cross-compiling for?

arm-unknown-linux-gnueabihf

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [X] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [ ] docker
- [X] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.4 (47df5c7 2022-08-30)

Alexhuszagh:
This should be linked to #426. I wonder if our update to fix targeting ARMv6 HF platforms was insufficient. Previously, we erroneously compiled to an ARMv7 target, which would have caused the bug.

This should have been changed in the v0.2.4 branch, but just to check: can you run `readelf -A $binary`, where `$binary` is the path to the cross-compiled hello world binary? It should output ARMv6 if it is, in fact, using the correct toolchain. If not, try updating the toolchain via `docker pull`. Let me know if you have any questions.

tmayoff:
```
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "7-A"
  Tag_CPU_arch: v7
  Tag_CPU_arch_profile: Application
[...]

PR #1018, opened 2022-09-12
Update deny list to all system packages for ARMv6 HF.

Alexhuszagh:
Currently, when using `arm-unknown-linux-gnueabihf`, users frequently try to install packages for the `armhf` Debian architecture, objects which are compiled for ARMv7-A, and therefore are not compatibility with ARMv6 platforms such as the Raspberry Pi 0. This differs from the soft-float targets, where packages for the `armel` architecture are compiled for the ARMv5TE architecture, and therefore are compatible with ARMv5TE, ARMv6, and ARMv7 soft-float targets.

Deny-listing all packages for the `armhf` architecture prevents issues such as in #1017, where the use of ARMv7 system packages are linked to the binary, so the resulting binary can no longer be executed on an ARMv6 platform. This is difficult to detect until the user attempts to use the generated objects (binaries, libraries) on native hardware.

Also linked is #426.

### DIFF
--- a/.changes/1018.json
+++ b/.changes/1018.json
@@ -1,0 +1,4 @@
+{
+    "description": "deny installation of armhf debian packages for the arm-unknown-linux-gnueabihf target.",
+    "type": "fixed"
+}
--- a/docker/Dockerfile.arm-unknown-linux-gnueabihf
+++ b/docker/Dockerfile.arm-unknown-linux-gnueabihf
@@ -17,6 +17,9 @@ RUN /crosstool-ng.sh arm-unknown-linux-gnueabihf.config 5
 
 ENV PATH /x-tools/arm-unknown-linux-gnueabihf/bin/:$PATH
 
+COPY deny-debian-packages.sh /
+RUN TARGET_ARCH=armhf /deny-debian-packages.sh
+
 COPY qemu.sh /
 RUN /qemu.sh arm
 
--- a/docker/deny-debian-packages.sh
+++ b/docker/deny-debian-packages.sh
@@ -3,15 +3,24 @@
 set -x
 set -euo pipefail
 
-main() {
-    local package
-
-    for package in "${@}"; do
-        echo "Package: ${package}:${TARGET_ARCH}
+deny_package() {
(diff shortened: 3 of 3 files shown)


Alexhuszagh:
Marking as a draft since I haven't tested this logic extensively, and denylists an entire architecture, so it's likely a good idea to be careful. Also, we might want to do this for a few musl targets in case if installing system packages can affect their dynamic linking (since this has been an issue with x86_64, see #902).

tmayoff:
This seems like a decent solution, as long as it's mentioned in docs somewhere (in case I forget about this and try installing a package that I know exists but it can't find), also I'm not sure if it is mentioned somewhere but a note that at least the `:armhf` packages are built for armv7 and that for armv6 might need to be compiled from source would also be good.

Alexhuszagh:
> This seems like a decent solution, as long as it's mentioned in docs somewhere (in case I forget about this and try installing a package that I know exists but it can't find), also I'm not sure if it is mentioned somewhere but a note that at least the `:armhf` packages are built for armv7 and that for armv6 might need to be compiled from source would also be good.

This needs to be added to the Wiki then as well, under the FAQ section then:
https://github.com/cross-rs/cross/wiki/FAQ

I'll do that later, thank you.

Emilgardis:
we don't have mentions for the other pins, probably a good idea to document it

Alexhuszagh:
I've added documentation on the pinned packages [here](https://github.com/cross-rs/cross/wiki/FAQ#cannot-find-package), and where we use them. I added the `arm-unknown-linux-gnueabihf` example as well since I'm assuming this will be merged: if not, we should still document this lack of compatibility somewhere.

Alexhuszagh:
The musl images don't need to deny the Debian system packages as expected since the musl sysroot is different, and the compiler doesn't try to link to those packages.

Emilgardis (approved):
awesome!

bors r+

Alexhuszagh:
bors r=Emilgardis

Emilgardis:
haha nice catch on the wrong commit message 😅

PR #1022, opened 2022-09-12
Support docker images with targets containing `.`s.

### LINKED ISSUE #1020: Future-Proofing Images with Subs
Rust triples are allowed to contain `.` characters, although I believe this is limited only to the architecture portion of the triple (and not the OS, vendor, or environment). For example, `thumbv8m.main-none-eabihf` is a valid LLVM triple. This means that splitting on a `.` character isn't necessarily going to give us our desired sub, and might produce erroneous results in the future.

We can get a bit more clarity looking at the comments in the [llvm/ADT/Triple.h](https://llvm.org/doxygen/Triple_8h_source.html) header. First, we want too look into the [parser](https://llvm.org/doxygen/Triple_8cpp_source.html#l00941), in `std::string Triple::normalize(StringRef Str)`. This isn't too bad, since it splits off into `parseArch`, `parseSubArch`, `parseVendor`, `parseOS`, and `parseEnvironment`.

We can see `parseArch` has some known values, and special cases for ARM and BPF-family architectures. The vendors are hard-coded and only accepted ASCII lowercase letters. The OSes and environments only expect strings **starting** with ASCII numbers and letters,  but are flexible for any possible characters (other than `-`) at the end of each identifier. The architecture field, however, will accept `.` characters for a few targets.

In short, the following code isn't generalizable enough: this will cause an invalid target/sub to be parsed from some valid Rust triples:

https://github.com/cross-rs/cross/blob/7c06ad9f8b195d2a1867b176bf9335721f3186cd/xtask/src/util.rs#L164-L180

Alexhuszagh:
Fix parsing of image targets for building docker images so it only uses subs containing only ASCII letters, so targets like `thumbv8m.main-none-eabihf` aren't mistaken for a target/sub combination.

Closes #1020.

### DIFF
--- a/xtask/src/util.rs
+++ b/xtask/src/util.rs
@@ -165,17 +165,23 @@ impl std::str::FromStr for ImageTarget {
     type Err = std::convert::Infallible;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if let Some((target, sub)) = s.split_once('.') {
-            Ok(ImageTarget {
-                name: target.to_string(),
-                sub: Some(sub.to_string()),
-            })
-        } else {
-            Ok(ImageTarget {
-                name: s.to_string(),
-                sub: None,
-            })
+        // we designate certain targets like `x86_64-unknown-linux-gnu.centos`,
+        // where `centos` is a subtype of `x86_64-unknown-linux-gnu`. however,
+        // LLVM triples can also contain `.` characters, such as with
+        // `thumbv8m.main-none-eabihf`, so we make sure it's only at the end.
+        if let Some((target, sub)) = s.rsplit_once('.') {
+            if sub.chars().all(|x| char::is_ascii_alphabetic(&x)) {
+                return Ok(ImageTarget {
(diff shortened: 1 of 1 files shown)


Emilgardis (approved):
seems fine, but I think there's a better way.

Instead make `target.none` be a substitute for `target`, and thus require troublesome target names to be named `target.withdot.none`

I'm happy with this either way

Alexhuszagh:
> seems fine, but I think there's a better way.
> 
> Instead make `target.none` be a substitute for `target`, and thus require troublesome target names to be named `target.withdot.none`
> 
> I'm happy with this either way

I tried to avoid that just because that would require `cargo build-docker-image "$target".none`, however, you are right that this may conflict if we have a triple where the environment might have a `.`. I don't ever see this occurring, however, since I've never seen a triple where this exists.

Emilgardis:
feel free to merge unless you want to implement the other solution

Alexhuszagh:
bors r=Emilgardis

PR #1023, opened 2022-09-15
Add support for Android versions 5-11.

Alexhuszagh:
**Configuration**

We provide the following arguments to configure the Android version:
- `ANDROID_NDK`
- `ANDROID_SDK` (note that this is dependent on `ANDROID_VERSION)
- `ANDROID_VERSION`
- `ANDROID_SYSTEM_COMPLETE`: do a complete Android build
- `ANDROID_SYSTEM_NONE`: do not build the Android system, disables runner support

**Version Support**

We now support NDK versions r10e-r25b, SDK versions 21-33, and Android versions 5.0, 5.1, 6.0, 7.0, 8.0, 8.1, 9.0, 10.0, and 11.0. We also validate that the NDK, SDK, and Android versions are compatible.

For Android 12+, we do support using the NDK with the proper API level, allowing cross-compilation for the desired target, however, running or testing the generated binary is currently not supported without a complete build. Note that support for non-complete builds with Android versions 12+ is unlikely to ever occur, due to issues with APEX in the build system.

**Implementation Details**

Next, we've improved the removal of unittests during the build process, to ensure fast builds while maintaining compatibility with various newer Android versions. To do this, we've implemented a Python library and a script. The Python library contains a complete parser (correctly parses all valid input) for Soong blueprint files, using an LALR grammar, and a rudimentary parser for Makefiles.

The Soong parser removes any `gtest` dependencies, as well as any subdirectories or scope names containing `test`. For example:

```go
cc_library {
    name: "lib",
    srcs: ["lib.cc",],
}
cc_test {
    name: "test",
    defaults: ["target"],
    srcs: ["test.cc"],
}
```

Will become:

```go
cc_library {
    name: "lib",
    srcs: ["lib.cc",],
}
```

The Makefile parser first splits the file based on conditional directives (`ifeq`, `endif`, etc.) to ensure any subsequent processing doesn't lead to unbalanced directives. Next, we split the text within each directive based on comment sections used in the Android source tree. For example:

```Makefile
test_tags := tests

include $(call subdir,$(LOCAL_PATH))

c_flags := \
  -g \
  -Wall
```

We can therefore remove the `Benchmarks` and `Unit tests` sections without removing the `Other section`.

The Python library is reasonably performant (it adds no noticeable overhead to the build process) and is compact (in total, < 60KB). Also, it is much more resilient than a series of `sed` scripts.

Finally, extensive unittests have been added for the Python library, for both code linting (`flake8`) and unittests (via `tox`). Since we cannot assume the presence of Python on the host machine, the tests can be optionally enabled via the `--python` flag (or `PYTHON` environment variable, to hook into the git hooks), and custom overrides for the `flake8` and `tox` commands are provided (since the user may wish to specify a specific Python version, such as `python3.7 -m flake8`).

**Linker**

For Android 10+, since we use a minimal Android build, we only support the bootstrap and ASAN linkers unless using a complete Android build. Supporting the APEX linker requires a nearly complete Android runtime, requiring 60+GB image sizes, slow builds, and other prohibitive factors.

**Complete Builds**

Complete builds are currently untested, for Androids version 12+ they exceed the default storage capacity of WSL2 images (250GB). They should work and properly install both the `/system` and `/apex` directories, however, the builds are slow and untested.


Alexhuszagh:
Currently only tested on `9.0.0_r1`, needs testing on a few more versions:
- [x] 9.0.0_r1
- [ ] 11.0.0_r48
- [ ] 12.1.0_r31
- [ ] 13.0.0_r6

docker/Dockerfile.aarch64-linux-android
```diff
@@ -20,7 +24,7 @@ RUN /android-system.sh arm64
 COPY qemu.sh /
 RUN /qemu.sh aarch64
```
Emilgardis, line 25: is it possible to move this back? That way when changing the args you don't have to rebuild qemu

docker/android-system.sh
```diff
@@ -8,12 +8,21 @@ set -euo pipefail
 
 main() {
     local arch="${1}"
+    local version="${ANDROID_VERSION}"
+    local gcc_version=4.9
+    local glibc_version=glibc2.15-4.8
```
Emilgardis, line 13: are these guaranteed for future versions?

docker/Dockerfile.aarch64-linux-android
Alexhuszagh: Sure, it is.

docker/android-system.sh
Alexhuszagh: I don't believe so. The GCC version likely is, but there's multiple glibc versions present, which possibly could be removed at some point in the future. Unlikely, but possible. I want to be able to version-gate it. We also might want to upgrade it at the very least: some newer versions include `prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.19-4.9`.

Alexhuszagh: Actually this is Android version-dependent. I keep forgot the platform meta repository lists repositories that have been removed. Updated to:

```bash
    case "${major_version}" in
        4)
            glibc_version=glibc2.7-4.6
            ;;
        5|6|7|8|9)
            glibc_version=glibc2.15-4.8
            ;;
        10|11|12|13)
            glibc_version=glibc2.17-4.8
            ;;
        *)
            # current master branch as of 9/15/2022
            glibc_version=glibc2.19-4.9
            ;;
    esac
```

Alexhuszagh:
bors try --target aarch64-linux-android

Alexhuszagh:
bors try --target aarch64-linux-android

Alexhuszagh:
I've restored using Python for the Android NDK install steps, since it's a bit more resilient than making all the symlinks outself manually. We could probably speed it up slightly making manual symlinks, but considering the major bottleneck is the Android build itself, and this adds no size to the resulting image, I don't see any reason to remove it. I've added `python3` since newer Android NDK versions use it by default and it also has effectively no change in the build times or performance.

Alexhuszagh:
bors try --target aarch64-linux-android

Alexhuszagh:
Android 10.x fails during runtime due to a broken symbolic link from `system/bin/linker64` to `/apex/com.android.runtime/bin/linker64`. This might require building `apex` in 10.x, which we normally only do in 11+, so this might be fixable.

Alexhuszagh:
bors try *-android*

docker/.dockerignore
```diff
@@ -0,0 +1,11 @@
+# don't copy any of the python artifacts to the docker context
+__pycache__/
```
Alexhuszagh, line 2: This is just to ensure any Python bytecode, eggs, distribution data, or anything else isn't copied over, as well as the test suite, since we need to copy over an entire Python module and run it in separate steps, so we want to minimize the image size.

Alexhuszagh:
bors try-
bors try --target *-android*

docker/Dockerfile.aarch64-linux-android
```diff
@@ -10,27 +10,51 @@ RUN /cmake.sh
...
     CXX_aarch64_linux_android=aarch64-linux-android-g++ \
+    LD_aarch64_linux_android=aarch64-linux-android-ld \
+    NM_aarch64_linux_android=aarch64-linux-android-nm \
+    OBJCOPY_aarch64_linux_android=aarch64-linux-android-objcopy \
+    OBJDUMP_aarch64_linux_android=aarch64-linux-android-objdump \
+    RANLIB_aarch64_linux_android=aarch64-linux-android-ranlib \
+    READELF_aarch64_linux_android=aarch64-linux-android-readelf \
+    SIZE_aarch64_linux_android=aarch64-linux-android-size \
+    STRINGS_aarch64_linux_android=aarch64-linux-android-strings \
+    STRIP_aarch64_linux_android=aarch64-linux-android-strip \
```
Alexhuszagh, line 57: I decided it was probably a good idea to hard-code these values specifically, since in recent versions the Android NDK has switched to using `llvm-$tool` for everything besides GCC instead, and it's better to ensure we have consistent naming and any future changes don't implicitly cause issues.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -10,27 +10,51 @@ RUN /cmake.sh
...
+COPY android $PYTHON_TMPDIR
 RUN /android-system.sh arm64
 
-COPY qemu.sh /
-RUN /qemu.sh aarch64
-
 ENV CROSS_SYSROOT=/android-ndk/sysroot
-RUN cp $CROSS_SYSROOT/usr/lib/aarch64-linux-android/28/libz.so /system/lib/
+COPY android-symlink.sh /
+RUN /android-symlink.sh aarch64 aarch64-linux-android
```
Alexhuszagh, line 37: The logic for creating symlinks and also copying over files to ensure they're in the right locations has gotten more complex, so it's been moved to a dedicated script.

docker/Dockerfile.arm-linux-androideabi
```diff
@@ -10,27 +10,49 @@ RUN /cmake.sh
...
+COPY validate-android-args.sh /
+RUN /validate-android-args.sh arm
+
 COPY android-ndk.sh /
-RUN /android-ndk.sh arm 28
+RUN /android-ndk.sh arm
 ENV PATH=$PATH:/android-ndk/bin
 
-COPY android-system.sh /
+COPY android-system.sh remove_android_tests.py /
```
Alexhuszagh, line 30: This is a bug and needs to be fixed.

Alexhuszagh:
bors try-

docker/android-ndk.sh
```diff
@@ -6,26 +6,59 @@ set -euo pipefail
...
 . lib.sh
 
-NDK_URL=https://dl.google.com/android/repository/android-ndk-r21d-linux-x86_64.zip
-
 main() {
-    local arch="${1}" \
-        api="${2}"
+    local arch="${1}"
+
+    if [[ "${ANDROID_SDK}" -eq 25 ]]; then
```
Alexhuszagh, line 12: This isn't required anymore since it's checked in `validate-android-args.sh`.

docker/android-ndk.sh
```diff
@@ -6,26 +6,59 @@ set -euo pipefail
...
+        exit 1
+    fi
 
-    install_packages curl unzip python
+    # python3 is still needed for newer NDK versions, just since it
+    # simplifies making symlinks even though the toolchain is prebuilt
+    install_packages curl python python3
+    get_ndk_info
+    if [[ "${NDK_VERSION}" -le 9 ]]; then
+        install_packages bzip2
```
Alexhuszagh, line 22: Very old versions package the binaries as a `tar.bz2` format, while anything Android NDK 10+ ships as a zipfile. As the NDK version 9 cannot be used with Android 5, in theory someone might not want to have an Android system and play around with very old versions, so I've left the logic here.

docker/android-ndk.sh
```diff
@@ -6,26 +6,59 @@ set -euo pipefail
...
+        tar -xjf "${NDK_FILENAME}"
+    else
+        unzip -q "${NDK_FILENAME}"
+    fi
+    rm "${NDK_FILENAME}"
+    pushd "android-ndk-${ANDROID_NDK}"
+    # android NDK versions <= 13 error without the verbose flag
+    local build_cmd=
+    local api=
+    if [[ "${NDK_VERSION}" -le 11 ]]; then
```
Alexhuszagh, line 42: Newer NDK versions migrated from a shell script to a Python module. They don't do any builds, however, so the dependencies are the same.

docker/android-system.sh
```diff
@@ -1,4 +1,30 @@
...
+#   5.0: 21 (tested at NDK 10e and r13b, 5.0.0_r1)
+#   5.1: 22 (tested at NDK r21d, 5.1.1_r38, unused DT)
+#   6.0: 23 (tested at NDK r21dm 6.0.1_r81)
+#   7.0: 24 (tested at NDK r21d, 7.0.0_r36)
+#   7.1: 25 (tested at NDK r21d, 7.1.2_r39, not supported)
+#   8.0: 26 (tested at NDK r21d, 8.0.0_r51)
+#   8.1: 27 (tested at NDK r21d, 8.1.0_r81)
+#   9.0: 28 (tested at NDK r21d and r25b, 9.0.0_r1)
+#   10.0: 29 (tested at NDK r25b, 10.0.0_r47)
+#   11.0: 30
```
Alexhuszagh, line 22: This needs to document the version testing.

docker/android-system.sh
```diff
@@ -7,32 +33,39 @@ set -euo pipefail
...
 
 main() {
-    local arch="${1}"
-    local td
-    td="$(mktemp -d)"
-    pushd "${td}"
+    export ARCH="${1}"
+    MAJOR_VERSION=$(echo "${ANDROID_VERSION}" | cut -d '.' -f 1)
+    MINOR_VERSION=$(echo "${ANDROID_VERSION}" | cut -d '.' -f 2)
+    TAG="android-${ANDROID_VERSION}"
```
Alexhuszagh, line 39: This is to avoid having to pass all these variables to the called functions, since we need to break down our logic extensively.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
         python \
         python3 \
         xz-utils
-    purge_list+=(default-jre)
```
Alexhuszagh, line 49: I need to remove the two files (`java` and `javac`) manually.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+        5|6)
+            java_type=java
+            java_version=1.7.0
+            jre_info="IcedTea 2.6.9"
+            build_info="build 24.131-b00, mixed mode"
+            ;;
+        *)
+            java_type=openjdk
+            java_version=1.8.0_342
+            jre_info="build 1.8.0_342-8u342-b07-0ubuntu1~20.04-b07"
```
Alexhuszagh, line 160: Android 7+ changed the required Java version and forces an OpenJDK runtime environment. I've therefore changed it to use Java 1.8 and an OpenJDK runtime using the standard install on a Ubuntu-based system.

docker/android-system.sh
```diff
@@ -7,32 +33,39 @@ set -euo pipefail
...
 
-    # more faking
-    export ANDROID_JAVA_HOME=/tmp
-    mkdir /tmp/lib/
-    touch /tmp/lib/tools.jar
+    local td
+    td="$(mktemp -d)"
+    pushd "${td}"
+
+    fake_java
```
Alexhuszagh, line 68: Complete Android builds might actually require a true Java installation. I haven't tested this, so YMMV.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+        if [[ ! -f "/system/bin/${tool}" ]]; then
+            ln -s /system/bin/toolbox "/system/bin/${tool}"
+        fi
+    done
+
+    echo "127.0.0.1 localhost" > /system/etc/hosts
+}
+
+symlink_bootstrap() {
+    # for Android 10+, we need to use the bootstrap linker rather than
```
Alexhuszagh, line 238: All the linkers are now installed in `/system/bin/bootstrap/` for `linker`, `linker64`, `linker_asan`, etc. The ones in `/system/bin/` are broken symlinks to `/apex/com.android.runtime/...`, which means we have to replace these symlinks with the bootstrap ones. This works perfectly fine, however.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+            for lib in "${libdir}/bootstrap/"*; do
+                file=$(basename "${lib}")
+                unlink "${libdir}/${file}"
+                ln -s "${libdir}/bootstrap/${file}" "${libdir}/${file}"
+            done
+        fi
+    done
+}
+
+# this are the minimum set of modules that are need to build bionic
```
Alexhuszagh, line 266: I've separated the logic for each Android version, rather than do a series of if chains, because it makes it much less clear what the specific dependencies for each version are. This simplifies maintenance a lot, in case something breaks for a specific Android version but not another patch or minor version.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+    # avoid build tests
+    rm bionic/linker/tests/Android.mk
+    rm bionic/tests/Android.mk
+    rm bionic/benchmarks/Android.mk
+    rm prebuilts/misc/common/android-support-test/Android.mk
+    # we don't need the relocation packer, and removing
+    # the unittests from it is a bit of work.
+    rm bionic/tools/relocation_packer/Android.mk
+
+    remove_tests
```
Alexhuszagh, line 421: At this point, the Makefiles have test targets inside the same file as the actual products. This means we can't simply remove `Android.mk` files since this might lead to broken inclusions, but worse at the targets and objects declared inside non-test files. This means simple `sed` scripts are no longer viable in all cases.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+}
+
+# tested on 8.0.0_r51 (SDK 26)
+# tested on 8.1.0_r81 (SDK 27)
+android_repo_v8() {
+    # need to build LLVM components, or libLLVM is disabled.
+    export FORCE_BUILD_LLVM_COMPONENTS=true
```
Alexhuszagh, line 428: Android 8 is a pain because it now requires the LLVM components, but by default turns off building them from source, and using system packages doesn't work. We therefore need to force their compilation, making these builds quite sluggish but overall, work fine.

docker/android-system.sh
Alexhuszagh: See the comments in the Python module for `android/make.py` for a more detailed description of this.

docker/android-system.sh
```diff
@@ -46,64 +79,474 @@ EOF
...
+    rm system/core/libsysutils/Android.bp
+    local path=
+    find prebuilts/misc/ -name 'Android.mk' | while IFS= read -r path; do
+        rm "${path}"
+    done
+
+    # avoid java dependencies
+    rm external/lzma/Java/Tukaani/Android.mk
+
+    remove_tests
```
Alexhuszagh, line 498: Due to the introduction of the Soong build system in Android 8+, we need much more complex build system modification. This is described in detail in `android/soong.py` below.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
     rm bionic/tests/headers/Android.bp
     rm bionic/tests/headers/posix/Android.bp
 
-    sed -i -z -e 's/cc_test {.*}//g' bionic/libc/malloc_debug/Android.bp
-    sed -i -z -e 's/cc_test {.*}//g' bionic/libc/malloc_hooks/Android.bp
-    sed -i -z -e 's/cc_test_host {.*}//g' bionic/tools/relocation_packer/Android.bp
```
Alexhuszagh, line 121: These are all handled by the Python Soong parser below, and have the added benefit of actually handling matching braces. Before, the sed script would either need to be greedy (so these would remove contents until the end of the file), or non-greedy, in which case no inner objects could exist.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
...
+    sync prebuilts/go/linux-x86
+    sync prebuilts/ndk
+    sync prebuilts/sdk
+    sync prebuilts/vndk/v28
+    sync system/core
+    sync system/sepolicy
 
-    # patch the linker to avoid the error
-    # FATAL: kernel did not supply AT_SECURE
-    #sed -i -e 's/if (!kernel_supplied_AT_SECURE)/if (false)/g' bionic/linker/linker_environ.cpp
```
Alexhuszagh, line 127: This is only relevant for Android 5, so it's been removed from more recent versions. Since it's documented there, there's no reason to duplicate in every other Android level.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
...
 
-    popd
+    # make sure we don't build benchmarks or apex
+    rm -r bionic/apex
+    rm -r bionic/benchmarks/
 
-    rm -rf "${td}"
-    rm "${0}"
+    # libziparchive has tests in the header, remove them
+    local libziparchive_h="system/core/libziparchive/include/ziparchive/zip_writer.h"
```
Alexhuszagh, line 694: This is a very simple fix and I'd rather not write a parser for generating an AST for C and then be able to serialize it back to well-formatted, human readable source code. Luckily, `sed` works perfectly fine here.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
...
+    sed -i -e 's/#include <gtest\/gtest_prod.h>//g' "${libziparchive_h}"
+    sed -i -e 's/FRIEND_TEST(zipwriter, WriteToUnseekableFile);//g' "${libziparchive_h}"
+
+    remove_tests
+}
+
+android_repo_v12() {
+    # FIXME: this is a work in progress, and is unlikely to ever
```
Alexhuszagh, line 702: I've kept this in since it's a work-in-progress. It might never finish, but it's a good starting point if anyone can ever do this work in the future.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
...
+    rm -r system/linkerconfig/testmodules
+
+    remove_tests
+}
+
+remove_tests() {
+    install_packages python3-pip
+
+    local version=
+    version=$(python3 -c 'import sys
```
Alexhuszagh, line 802: The actual Python version may change with future Ubuntu updates, or if we upgrade to 22.04, etc. Since the Python path will differ accordingly for the installed packages, we need to detect the major and minor versions reliably. Luckily, this is very simple with `python -c $source_code`.

docker/android-system.sh
```diff
@@ -116,49 +559,264 @@ EOF
...
+major = sys.version_info.major
+minor = sys.version_info.minor
+print(f"{major}.{minor}")')
+    set +u
+    export PYTHONPATH="${PYTHON_TMPDIR}/lib/python${version}/site-packages/:${PYTHONPATH}"
+    set -u
+    mkdir -p "${PYTHON_TMPDIR}"
+    python3 -m pip install sly==0.4.0 --prefix "${PYTHON_TMPDIR}"
+    python3 -m pip install google-re2==1.0 --prefix "${PYTHON_TMPDIR}"
```
Alexhuszagh, line 811: Install our dependencies to a temporary directory `/tmp/android/lib/python3.8/site-packages` so removing the temp directory automatically cleans these dependencies up. Since we insert them in `PYTHONPATH`, they're automatically found.

docker/android/android/__init__.py
```diff
@@ -0,0 +1,15 @@
+import sys
+
+# we run this script once every build, and we'd rather
+# have much smaller image sizes, so copying without
+# any bytecode is a better idea.
+sys.dont_write_bytecode = True
```
Alexhuszagh, line 6: There's no reason not to have this, since we're almost entirely running this script **inside** the Docker image and only running it once, so any nearly insignificant performance improvements during our git hooks (which won't test Python unless you explicitly opt-in) won't matter.

docker/android/android/util.py
```diff
@@ -0,0 +1,28 @@
...
+
+def flatten(lst):
+    return [i for sublist in lst for i in sublist]
+
+
+def _is_match(pattern, string):
+    return re.search(pattern, string) is not None
+
+
+def is_test(string):
```
Alexhuszagh, line 17: Determining if something is a test isn't exactly trivial. Examples of tests include:
- `libgtest`
- `libgtest_main`
- `art-test-defaults`
- `cc_test`
- `cc_test_host`
- `py2-c-module_test_capi`.

While example of non-tests include:
- `latest`
- `fmtlib-non-test-defaults`.

So I've written a few false positives, and then used the fact that anything besides `gtest` if it is prefixed by a letter or digit is not a test. Any other printable character, however, is.

docker/android/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "android"
+version = "0.0.0-dev.0"
+license = { text = "MIT OR Apache-2.0" }
+dependencies = ["sly==0.4"]
```
Alexhuszagh, line 5: This needs to include `google-re2==1.0`.

docker/android/scripts/build-system.py
```diff
@@ -0,0 +1,207 @@
...
+    git_glob = f'{root}/**/.git'
+    for path in glob.iglob(git_glob, recursive=True):
+        os.chdir(os.path.dirname(path))
+        subprocess.check_call(['git', 'stash'])
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    action_group = parser.add_mutually_exclusive_group(required=True)
+    action_group.add_argument(
```
Alexhuszagh, line 130: These options are quite handy when debugging: we can use `--backup` to restore from build-system only files, `--stash` allows us to simply stash all changes and start from scratch (rather than have to use `git stash` in every repository ourselves, and `--remove-tests` is our main, desired functionality.

docker/android/scripts/build-system.py
```diff
@@ -0,0 +1,207 @@
...
+        action='store_true',
+    )
+    parser.add_argument(
+        '--disable-backup',
+        help='disable automatic backup of build files during processing.',
+        action='store_false',
+    )
+    flags_group = parser.add_mutually_exclusive_group()
+    flags_group.add_argument(
+        '--soong-only',
```
Alexhuszagh, line 157: These are quite handy if in the future something breaks with our Makefile parser, but not our Soong parser, or the inverse, and tests do not need to be removed from the other. This might be a reality since the Makefiles are slowly being deprecated and we have an incomplete parser for them: only our Soong parser is correct.

docker/android/android/make.py
```diff
@@ -0,0 +1,513 @@
...
+              CommentBlock: Section 2
+                Directive Block:
+                  start=ifneq ($(USE_B),)
+                  end=endif
+                  children:
+                    Block: `SOURCES=b.cc\nelse\nSOURCES=a.cc`
+              Block: `else\nSOURCES=c.cc`
+'''
+
+import re2 as re
```
Alexhuszagh, line 71: We desperately need `re2` since some specific edge-cases work abysmally with backtracking regular expressions. This can mean it will take **hours** to match a header with Python's built-in `re` library for what takes milliseconds or less with `re2`.

docker/android/android/make.py
```diff
@@ -0,0 +1,513 @@
+'''
```
Alexhuszagh, line 1: Because of how the Android Makefiles are designated, we don't need a complete or correct parser in earlier versions to separate the tests from the actual non-test targets. This is because all tests are designated by something like:

```Makefile
##################################
# test executable
```

Or:

```Makefile
# Unit tests.
# -------------------------------------
```

The valid section headers are `#########` (repeated more), `# -----------`, or `# =========`, and they can sandwich the header, be on top, or on bottom. The header title can also be multiple lines. We handle all these cases, including empty headers like:

```Makefile
#######################
#


```

docker/android/android/soong.py
```diff
@@ -0,0 +1,654 @@
+'''
```
Alexhuszagh, line 1: This is a complete and **mostly** correct Soong parser. It will accept all valid input, and reject most invalid input. It however, converts it to an AST, but does not do validation of that AST, only the initial grammar, so identifiers of incorrect types may not be properly processed. This is irrelevant here, since we only want to remove tests from the file and then re-serialize it: not validate and re-implement the entire build system.

docker/android/android/soong.py
```diff
@@ -0,0 +1,654 @@
...
+
+    this uses LALR parsing since it makes the grammar very easy to define and
+    the parsing simple. since the build step and repository synchronization
+    is much slower, the performance here is practically irrelevant.
+'''
+
+import json
+import sys
+
+import sly
```
Alexhuszagh, line 63: We use an LALR parser since it makes implementing the parser very easy with a simple and concise grammar. It's basically a user-friendly version of the older tools `lex` and `yacc`:
https://sly.readthedocs.io/en/latest/sly.html#writing-a-lexer

docker/android/android/soong.py
```diff
@@ -0,0 +1,654 @@
...
+        COMMA,
+        EQUALS,
+        PLUS,
+    }
+    ignore = ' \t'
+    ignore_comment = _COMMENT
+
+    # Tokens
+    # this uses a string regex based on the CSS2.1 grammar
+    STRING = fr'"([^\n\r\f\\"]|\\{_NL}|{_ESCAPE})*"'
```
Alexhuszagh, line 117: Our complex grammars are mostly borrowed for the CSS2 specifications (before they swapped to railroad diagrams), which are great because it fully conforms to the Soong grammar.

docker/android/android/soong.py
```diff
@@ -0,0 +1,654 @@
...
+
+# dictionaries got insertion order in 3.6, guaranteed in 3.7
+assert sys.version_info >= (3, 6)
+
+# base character defs
+_H = r'[0-9a-f]'
+_NL = r'\n|\r\n|\r|\f'
+_UNICODE = fr'\\{_H}{1,6}(\r\n|[ \t\r\n\f])?'
+_ESCAPE = r'{_UNICODE}|\\[^\r\n\f0-9a-f]'
+_SINGLELINE_COMMENT = r'\/\/.*'
```
Alexhuszagh, line 75: Soong supports both C and C++ style comments. Luckily, there is no nesting, etc, so regular expression removal of comments is quite simple. We really don't care about the number of match steps, so unrolling this isn't a concern: readability is much more important.

docker/android/tests/Addition.bp
```diff
@@ -0,0 +1,62 @@
+// special file testing `+` and `+=` operators
```
Alexhuszagh, line 1: We have very comprehensive tests for the Soong grammar since it's not trivial. It may not be enormously complex, but there's enough edge-cases to be careful.

docker/android/tests/Android.mk
```diff
@@ -0,0 +1,101 @@
+LOCAL_PATH := $(call my-dir)
```
Alexhuszagh, line 1: Likewise, we want to check our Makefile "parsers" work, especially if we have conditional directives that could be removed by a comment. We don't want to generate unbalanced directives, so we have testing to ensure this.

src/rustc.rs
```diff
@@ -265,7 +265,7 @@ impl QualifiedToolchain {
             Ok(_) | Err(_) if config.custom_toolchain() => {
                 QualifiedToolchain::custom(toolchain, &sysroot, config, msg_info)
             }
-            Ok(_) => return Err(eyre::eyre!("toolchain is not fully qualified")
```
Alexhuszagh, line 268: Just a clippy lint that was detected with the latest nighties, and completely unrelated to the PR.

xtask/src/hooks.rs
```diff
@@ -1,12 +1,14 @@
 use std::fs::File;
```
Alexhuszagh, line 1: Add our Python checks and tests if we opt-in. Since we can't guarantee Python on the host machine, nor the presence of tox or flake8, we don't enable these by default.

Alexhuszagh:
bors try --target *-android*

Alexhuszagh:
There might be other things I'm missing: I've tried to run through this numerous times, but at 4k lines of code, there may be something I've overlooked. I've re-run the tests for the entire suite of Android versions on `aarch64-linux-android` on the latest version of the build scripts, so it should work for the tested versions for all targets, but there could be subtle logic errors.

Emilgardis:
since these scripts are very android specific and most of it would not be used in CI, would it make sense to move this to cross-toolchain?

Alexhuszagh:
> since these scripts are very android specific and most of it would not be used in CI, would it make sense to move this to cross-toolchain?

Sure, that could work. We can then also cherry pick pieces if we say, want to upgrade to Android 11.

Alexhuszagh:
If we do migrate this to cross-toolchains, it would be nice if we could have some basic CI to test the additions, however:
https://github.com/cross-rs/cross-toolchains/pull/24

src/rustc.rs
Alexhuszagh: If we move this to cross-toolchains, then this needs to be included in a separate PR.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -10,27 +10,51 @@ RUN /cmake.sh
 COPY xargo.sh /
 RUN /xargo.sh
 
+COPY qemu.sh /
+RUN /qemu.sh aarch64
+
+ARG ANDROID_NDK=r21d
```
Emilgardis, line 16: with https://github.com/rust-lang/rust/issues/103673 / https://github.com/rust-lang/rust/pull/102332

we should probably target 25, however it seems it may be reverted and incrementally be deployed, beginning with 21

xtask/src/hooks.rs
```diff
@@ -122,27 +136,125 @@ fn shellcheck(all: bool, msg_info: &mut MessageInfo) -> cross::Result<()> {
...
+    }
+    let join_to = tests.len() - 1;
+    let mut joined = tests[0..join_to].join(", ");
+    if join_to > 1 {
+        joined.push(',');
+    }
+    msg_info.info(format_args!(
+        "Running {joined} and {} tests.",
+        tests[join_to]
+    ))?;
```
Emilgardis, line 250: ditto
```suggestion
    msg_info.info(format_args!(
        "Running {} tests.",
        tests.join(',')
    ))?;
```

xtask/src/hooks.rs
```diff
@@ -122,27 +136,125 @@ fn shellcheck(all: bool, msg_info: &mut MessageInfo) -> cross::Result<()> {
...
+    }
+    let join_to = checks.len() - 1;
+    let mut joined = checks[0..join_to].join(", ");
+    if join_to > 1 {
+        joined.push(',');
+    }
+    msg_info.info(format_args!(
+        "Running {joined} and {} checks.",
+        checks[join_to]
+    ))?;
```
Emilgardis, line 220: I think it's fine to drop the `, and`

I think it's fine to drop the `, and` here

```suggestion
    let mut checks = vec!["rustfmt", "clippy", "shellcheck"];
    if python {
        checks.push("python");
    }
    msg_info.info(format_args!(
        "Running {} checks.",
        checks.join(",")
    ))?;
```

docker/android-system.sh
```diff
@@ -7,32 +34,39 @@ set -euo pipefail
...
 
-    chmod +x /usr/bin/java
-    chmod +x /usr/bin/javac
+    if [[ "${ANDROID_SYSTEM_COMPLETE}" != "1" ]] && [[ "${MAJOR_VERSION}" -ge 12 ]]; then
+        echo "Android versions 12 and higher couple APEX tightly into the build system." 1>&2
+        echo "These are currently unsupported, and are unlikely to ever be supported." 1>&2
+        echo "Try using a complete Android system build or disable building Android system." 1>&2
+        echo "Note that a complete Android system build is slow and creates massive images." 1>&2
+        echo "Disabling the Android system build will prevent the use of Android runners." 1>&2
+        echo "If you like support for newer versions, help is always wanted." 1>&2
```
Emilgardis, line 58:
```suggestion
        echo "If you want support for newer versions, contributions are always appreciated." 1>&2
```

Emilgardis (approved):
looks amazing, some minor comments

I'm not super familiar with the android build system so can't comment to much on the approach, but this seems great!

This might now be the most complete containerized build system for c/c++/rust projects in android available anywhere 😅 

i'm fine with merging this into this repo

Alexhuszagh:
Currently waiting on the fix for `-ar` and `-gcc-ar` in #1105, and also we had a spurious [network issue](https://github.com/cross-rs/cross/actions/runs/3369842159/jobs/5590119009#step:10:3177) in the merge.

Emilgardis:
bors retry

Alexhuszagh:
bors r=Emilgardis

PR #1028, opened 2022-09-24
Fix tests, linkers, and specifications for various targets.

Alexhuszagh:
- Fixes linking to libgcc for `armv5te-unknown-linux-musleabi`.
- Adds C++ support for FreeBSD targets by linking to `libstdc++`.
- Tests dynamic library support for Android targets in CI.
- Tests C++ support for `mips64el-unknown-linux-muslabi64` in CI.
- Convert `mips64el-unknown-linux-muslabi64` to a hard-float toolchain to match the rust target.
- Convert `mips64el-unknown-linux-muslabi64` to use the `mips64r2` architecture.
- Convert `mips-unknown-linux-musl` and `mipsel-unknown-linux-musl` to use the `mips32r2` architecture, identical to the rust targets.
- Document whether MIPS musl targets are hard or soft float.

### DIFF
--- a/.changes/1028.json
+++ b/.changes/1028.json
@@ -26,10 +26,6 @@
         "breaking": true
     },
     {
-        "description": "document whether MIPS musl targets are hard or soft float.",
-        "type": "internal"
-    },
-    {
         "description": "convert mips-unknown-linux-musl and mipsel-unknown-linux-musl to use the mips32r2 architecture, identical to the rust targets.",
         "type": "changed",
         "breaking": true
--- a/.changes/1028.json
+++ b/.changes/1028.json
@@ -1,0 +1,37 @@
+[
+    {
+        "description": "link to libgcc for armv5te-unknown-linux-musleabi.",
+        "type": "fixed"
+    },
+    {
+        "description": "add C++ support for FreeBSD targets.",
+        "type": "added"
+    },
+    {
+        "description": "test dynamic library support for Android targets in CI.",
+        "type": "internal"
+    },
+    {
+        "description": "test partial C++ support for mips64el-unknown-linux-muslabi64 in CI.",
+        "type": "internal"
+    },
+    {
+        "description": "convert mips64el-unknown-linux-muslabi64 to a hard-float toolchain to match the rust target.",
+        "type": "changed",
+        "breaking": true
+    },
+    {
+        "description": "convert mips64el-unknown-linux-muslabi64 to use the mips64r2 architecture, identical to the rust target.",
+        "type": "changed",
+        "breaking": true
+    },
+    {
+        "description": "document whether MIPS musl targets are hard or soft float.",
+        "type": "internal"
+    },
+    {
+        "description": "convert mips-unknown-linux-musl and mipsel-unknown-linux-musl to use the mips32r2 architecture, identical to the rust targets.",
+        "type": "changed",
+        "breaking": true
+    }
+]
--- a/docker/Dockerfile.mips64el-unknown-linux-muslabi64
+++ b/docker/Dockerfile.mips64el-unknown-linux-muslabi64
@@ -12,21 +12,29 @@ RUN /xargo.sh
 COPY qemu.sh /
 RUN /qemu.sh mips64el
 
+# this is a hard-float target for the mips64r2 architecture
+# https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mips64el_unknown_linux_muslabi64.rs#L6
 COPY musl.sh /
 RUN /musl.sh \
-    TARGET=mips64el-linux-muslsf \
-    "COMMON_CONFIG += -with-arch=mips64"
+    TARGET=mips64el-linux-musl \
+    "COMMON_CONFIG += -with-arch=mips64r2"
 
-ENV CROSS_MUSL_SYSROOT=/usr/local/mips64el-linux-muslsf
+ENV CROSS_MUSL_SYSROOT=/usr/local/mips64el-linux-musl
 COPY musl-symlink.sh /
-RUN /musl-symlink.sh $CROSS_MUSL_SYSROOT mips64el-sf
+RUN /musl-symlink.sh $CROSS_MUSL_SYSROOT mips64el
+RUN mkdir -p $CROSS_MUSL_SYSROOT/usr/lib64
+# needed for the C/C++ runners
+RUN ln -s $CROSS_MUSL_SYSROOT/usr/lib/libc.so $CROSS_MUSL_SYSROOT/usr/lib64/libc.so
+RUN ln -s $CROSS_MUSL_SYSROOT/usr/lib/libc.so.1 $CROSS_MUSL_SYSROOT/usr/lib64/libc.so.1
+
+COPY mips64el-linux-musl-gcc.sh /usr/bin/
 
 COPY qemu-runner base-runner.sh /
 
-ENV CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_LINKER=mips64el-linux-muslsf-gcc \
+ENV CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_LINKER=mips64el-linux-musl-gcc.sh \
     CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_RUNNER="/qemu-runner mips64el" \
-    CC_mips64el_unknown_linux_muslabi64=mips64el-linux-muslsf-gcc \
-    CXX_mips64el_unknown_linux_muslabi64=mips64el-linux-muslsf-g++ \
+    CC_mips64el_unknown_linux_muslabi64=mips64el-linux-musl-gcc \
+    CXX_mips64el_unknown_linux_muslabi64=mips64el-linux-musl-g++ \
     BINDGEN_EXTRA_CLANG_ARGS_mips64el_unknown_linux_muslabi64="--sysroot=$CROSS_MUSL_SYSROOT" \
     QEMU_LD_PREFIX=$CROSS_MUSL_SYSROOT \
     RUST_TEST_THREADS=1
--- a/docker/mips64el-linux-musl-gcc.sh
+++ b/docker/mips64el-linux-musl-gcc.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# this linker wrapper works around the missing soft-fp routine __trunctfsf2
+# this affects rust versions with compiler-builtins <= 0.1.77,
+# which has not yet been merged into stable. this requires the `-lgcc`
+# linker flag to provide the missing builtin.
+# https://github.com/rust-lang/compiler-builtins/pull/483
+
+set -x
+set -euo pipefail
+
+main() {
+    if [[ $# -eq 0 ]]; then
+        exec mips64el-linux-musl-gcc "${@}"
+    else
+        exec mips64el-linux-musl-gcc "${@}" -lgcc -static-libgcc
+    fi
+}
+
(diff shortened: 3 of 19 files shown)


Alexhuszagh:
Blocking this until #1008 is merged, so we can use the new TOML format for the matrix.

Alexhuszagh:
bors try

Alexhuszagh:
We might be able to fix the partial C++ support for `mips64el-unknown-linux-muslabi64` by checking the Rustc version in newer releases. If not, we can always link to `libgcc` like we do for `mips64-unknown-linux-muslabi64`. This might be the preferable solution.

Alexhuszagh:
bors try --target mips64el-unknown-linux-muslabi64

Alexhuszagh:
This is technically a breaking change for `mips64el-unknown-linux-muslabi64`, however, this changes our build to match the Rust target itself, and since we're moving to a 0.3.0 release, we're allowed to have breaking changes here:
https://github.com/cross-rs/cross/pull/1028/files#diff-8541a3a4775c5542e352ca560a66b3bb939d2dc6112554657dfd8930b87e60ff

Alexhuszagh:
We should clearly document the target support, since we know:
- `mips-unknown-linux-musl`: soft-float ([mips_unknown_linux_musl.rs#L7](https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mips_unknown_linux_musl.rs#L7))
- `mipsel-unknown-linux-musl`: soft-float ([mipsel_unknown_linux_musl.rs#L6](https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mipsel_unknown_linux_musl.rs#L6))
- `mips64-unknown-linux-muslabi64`: hard-float ([mips64_unknown_linux_muslabi64.rs#L7](https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mips64_unknown_linux_muslabi64.rs#L7))
- `mips64el-unknown-linux-muslabi64`: hard-float ([mips64el_unknown_linux_muslabi64.rs#L6](https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mips64el_unknown_linux_muslabi64.rs#L6))

So with this change, all our targets are now correct for soft or hard-float.

docker/Dockerfile.mips64el-unknown-linux-muslabi64
```diff
@@ -14,19 +14,25 @@ RUN /qemu.sh mips64el
 
 COPY musl.sh /
 RUN /musl.sh \
-    TARGET=mips64el-linux-muslsf \
+    TARGET=mips64el-linux-musl \
     "COMMON_CONFIG += -with-arch=mips64"
```
Alexhuszagh, line 18: This should probably also be `mips64r2` like the big-endian variant. See [mips64el_unknown_linux_muslabi64.rs](https://github.com/rust-lang/rust/blob/75d3027fb5ce1af6712e4503c9574802212101bd/compiler/rustc_target/src/spec/mips64el_unknown_linux_muslabi64.rs#L6).

```rust
base.features = "+mips64r2".into();
```

Alexhuszagh:
We can probably get away with building the musl toolchain for a generic `mips32` and `mips64` architecture, however, since the Rust target will use a more specific architecture, I doubt this will have any changes in practice. This will (minorly) affect external C or C++ dependencies, but shouldn't affect any Rust code.

Alexhuszagh:
bors try --target mips*musl*

.changes/1028.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+        "description": "convert mips64el-unknown-linux-muslabi64 to use the mips64r2 architecture, identical to the rust target.",
+        "type": "changed",
+        "breaking": true
+    },
+    {
+        "description": "document whether MIPS musl targets are hard or soft float.",
+        "type": "internal"
+    },
```
Emilgardis, line 31: This isn't really documented, only commented in the respective dockerfiles. I think this entry is not needed

docker/Dockerfile.mips64el-unknown-linux-muslabi64
```diff
@@ -12,21 +12,29 @@ RUN /xargo.sh
...
 
 COPY qemu-runner base-runner.sh /
 
-ENV CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_LINKER=mips64el-linux-muslsf-gcc \
+ENV CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_LINKER=mips64el-linux-musl-gcc.sh \
     CARGO_TARGET_MIPS64EL_UNKNOWN_LINUX_MUSLABI64_RUNNER="/qemu-runner mips64el" \
-    CC_mips64el_unknown_linux_muslabi64=mips64el-linux-muslsf-gcc \
-    CXX_mips64el_unknown_linux_muslabi64=mips64el-linux-muslsf-g++ \
+    CC_mips64el_unknown_linux_muslabi64=mips64el-linux-musl-gcc \
+    CXX_mips64el_unknown_linux_muslabi64=mips64el-linux-musl-g++ \
```
Emilgardis, line 37: ping @abcfy2 

does this change work for you?
You can build the image with

```
cargo xtask build-docker-image mips64el-unknown-linux-muslabi64
```

this is a hard-float target according to rustc

docker/mips64el-linux-musl-gcc.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# this linker wrapper works around the missing soft-fp routine __trunctfsf2
+# this affects rust versions with compiler-builtins <= 0.1.77,
+# which has not yet been merged into stable. this requires the `-lgcc`
+# linker flag to provide the missing builtin.
+# https://github.com/rust-lang/compiler-builtins/pull/483
```
Emilgardis, line 7: 0.1.78 is present in 1.65

Emilgardis (changes requested):

docker/mips64el-linux-musl-gcc.sh
Alexhuszagh: This can be applied then to both mips64 and mips64le then.

Alexhuszagh: Since this logic would now be present in 3 files, but i don't really want to add it to `lib.sh` for full rebuilds, should i just make `rust.sh` which would have `rustc_version` and `rustc_minor`? Something like:

```bash
#!/bin/bash

rustc_version() {
    rustc -Vv | grep '^release:' | cut -d ':' -f2
}

rustc_minor() {
    rustc_version | cut -d '.' -f2
}
```

And then just `rustc_minor` in `aarch64-linux-musl-gcc.sh`, `mips64-linux-musl-gcc.sh`, and `mips64el-linux-musl-gcc.sh` after sourcing `rustc.sh`?

Emilgardis: we could have cross inform about the rustc version also. An env var like `_RUSTC_MINOR_VERSION` or similar. Technically we should also report the major version, but there are no plans for a rust 2.0 :)

Alexhuszagh: We prefix every other one with `CROSS` so `CROSS_RUSTC_MAJOR_VERSION`, `CROSS_RUSTC_MINOR_VERSION`, and `CROSS_RUSTC_PATCH_VERSION` could work nicely. That way, we have no conflicts.

docker/Dockerfile.mips64el-unknown-linux-muslabi64
Alexhuszagh: It should be hard-float, we're just using a soft float toolchain which is bound to cause problems at some point. A quick test is:

```bash
$ cargo init --bin hello
$ cross build --target mips64el-unknown-linux-muslabi64
$ file target/mips64el-unknown-linux-muslabi64/debug/hello
target/mips64el-unknown-linux-muslabi64/debug/hello: ELF 64-bit LSB executable, MIPS, MIPS64 rel2 version 1 (SYSV), statically linked, with debug_info, not stripped
$ readelf -A target/mips64el-unknown-linux-muslabi64/debug/hello | grep Tag_GNU_MIPS_ABI_FP
  Tag_GNU_MIPS_ABI_FP: Hard float (double precision)
$ docker run -it --rm ghcr.io/cross-rs/mips64el-unknown-linux-muslabi64:main bash
# still using the soft-float toolchain
$ echo $CROSS_MUSL_SYSROOT/
/usr/local/mips64el-linux-muslsf/

# show 32-bit is soft-float
$ cross build --target mipsel-unknown-linux-musl
$ file target/mipsel-unknown-linux-musl/debug/hello
target/mipsel-unknown-linux-musl/debug/hello: ELF 32-bit LSB pie executable, MIPS, MIPS32 rel2 version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-mipsel-sf.so.1, with debug_info, not stripped
$ readelf -A target/mipsel-unknown-linux-musl/debug/hello | grep Tag_GNU_MIPS_ABI_FP
  Tag_GNU_MIPS_ABI_FP: Soft float
```

So we're linking a hard-float executable with a soft-float toolchain, which isn't exactly ideal, and bound to cause problems later down the line.

docker/mips64el-linux-musl-gcc.sh
Alexhuszagh: Does the latest commit fix this? I've added `rustc_version` to `DockerOptions`, and changed `docker_envvars` to take the `DockerOptions`, and then export the version info via:

```rust
    let (major, minor, patch) = match options.rustc_version.as_ref() {
        Some(version) => (version.major, version.minor, version.patch),
        // no toolchain version available, always provide older
        // compiler available. this isn't a major issue because
        // linking will libgcc will not include symbols found in
        // the builtins.
        None => (1, 0, 0),
    };
    docker.args(["-e", &format!("CROSS_RUSTC_MAJOR_VERSION={}", major)]);
    docker.args(["-e", &format!("CROSS_RUSTC_MINOR_VERSION={}", minor)]);
    docker.args(["-e", &format!("CROSS_RUSTC_PATCH_VERSION={}", patch)]);
```

This also fixes an issue if the `rustc` version doesn't provide the version info, which might have caused a bug in `aarch64-unknown-linux-musl`.

Alexhuszagh:
bors try --target mips*musl* --target aarch64-unknown-linux-musl

src/docker/local.rs
```diff
@@ -38,14 +38,7 @@ pub(crate) fn run(
...
         .specify_platform(&options.engine, &mut docker);
-    docker_envvars(
-        &mut docker,
-        &options.config,
-        dirs,
-        &options.target,
-        options.cargo_variant,
-        msg_info,
-    )?;
+    docker_envvars(&mut docker, &options, dirs, msg_info)?;
```
Emilgardis, line 41: very nice

src/docker/shared.rs
```diff
@@ -564,6 +572,18 @@ pub(crate) fn docker_envvars(
         docker.args(&parse_docker_opts(&value)?);
     };
 
+    let (major, minor, patch) = match options.rustc_version.as_ref() {
+        Some(version) => (version.major, version.minor, version.patch),
+        // no toolchain version available, always provide the oldest
+        // compiler available. this isn't a major issue because
+        // linking will libgcc will not include symbols found in
```
Emilgardis, line 579:
```suggestion
        // linking with libgcc will not include symbols found in
```

Emilgardis (commented):
looks good!

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #997, opened 2022-08-20
fix wrong usage of bool_from_envvar

Emilgardis:
fixes #996

PR #1032, opened 2022-09-27
Allow disabling buildkit for engines lacking support.

Alexhuszagh:


.changes/1036.json
```diff
@@ -0,0 +1,4 @@
+{
```
Alexhuszagh, line 1: this needs to change to `1032.json`.

Alexhuszagh:
This could also likely be fixed by either using an environment variable (to manually disable) or by checking if `$engine buildx --help` fails, since we've already checked `$engine --version` works at this point. This could use a slightly more comprehensive solution.

Emilgardis (approved):
this looks good to me for now

adding more comprehensive checks for automatically doing this could be done in the client/server checks for platform I think, only issue is that the fields are not stable between implementations

Alexhuszagh:
bors r=Emilgardis

Emilgardis:
bors retry

possibly spurious error on winehq

src/docker/engine.rs
```diff
@@ -108,6 +108,11 @@ impl Engine {
             .map(|s| bool_from_envvar(&s))
             .unwrap_or_default()
     }
+
+    #[must_use]
+    pub fn has_buildkit() -> bool {
+        !bool_from_envvar("CROSS_CONTAINER_ENGINE_NO_BUILDKIT")
```
Emilgardis, line 114: aha... the classic

this functions takes the value of the envvar, not the key...

Emilgardis (changes requested):

src/docker/engine.rs
Emilgardis: we should look into making this harder to missuse

Alexhuszagh: I thought for consistency we should keep it as `0`, `1`, `true`, `false`, etc.

Alexhuszagh:
> bors retry
> 
> possible spurious error on winehq

It looks like a signing key changed, I'll try to fix this and submit a PR:
https://github.com/cross-rs/cross/actions/runs/3145747924/jobs/5113458177#step:10:1640

Emilgardis:
> It looks like a signing key changed, I'll try to fix this and submit a PR

sounds good!

The zig error is due to not using `buildx` anymore (because `Engine::has_buildkit` always returns false), and thus `TARGETPLATFORM` is not set correctly

https://github.com/cross-rs/cross/actions/runs/3149116522/jobs/5120408064#step:10:1398

Alexhuszagh:
> > It looks like a signing key changed, I'll try to fix this and submit a PR
> 
> sounds good!
> 
> The zig error is due to not using `buildx` anymore (because `Engine::has_buildkit` always returns false), and thus `TARGETPLATFORM` is not set correctly
> 
> https://github.com/cross-rs/cross/actions/runs/3149116522/jobs/5120408064#step:10:1398

Ah ok, so I need to fix that logic then.

src/docker/engine.rs
Emilgardis: happened before here: https://github.com/cross-rs/cross/pull/997

Alexhuszagh: Ahhhhh. That is confusing, and I was the one who added that function.

Alexhuszagh: I've fixed the invocation and also added documentation for `bool_from_envvar`:

```rust
// this takes the value of the environment variable,
// so you should call `bool_from_envvar(env::var("FOO"))`
pub fn bool_from_envvar(envvar: &str) -> bool {
    if let Ok(value) = bool::from_str(envvar) {
        value
    } else if let Ok(value) = i32::from_str(envvar) {
        value != 0
    } else {
        !envvar.is_empty()
    }
}
```

```rust
    #[must_use]
    pub fn has_buildkit() -> bool {
        !env::var("CROSS_CONTAINER_ENGINE_NO_BUILDKIT")
            .map(|x| bool_from_envvar(&x))
            .unwrap_or_default()
    }
```

src/config.rs
```diff
@@ -186,6 +186,8 @@ fn split_to_cloned_by_ws(string: &str) -> Vec<String> {
     string.split_whitespace().map(String::from).collect()
 }
 
+// this takes the value of the environment variable,
+// so you should call `bool_from_envvar(env::var("FOO"))`
```
Emilgardis, line 190: make this a doc-comment and this should be good to go!

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #1033, opened 2022-09-27
Fix image build errors with podman and nerdctl.

### LINKED ISSUE #1031: Image Build Fails on Podman
### Checklist

### Describe your issue

Currently, attempting to build an image for Podman fails, producing the following error:

```bash
$ cargo build-docker-image $target --engine podman                                                                                                                                                  
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/xtask build-docker-image aarch64-linux-android --engine podman`
[cross] note: Build aarch64-linux-android for x86_64-unknown-linux-gnu
Error: unable to parse value provided `` to --cache-from: invalid repo "type=registry,ref=ghcr.io/cross-rs/aarch64-linux-android:main": must contain registry and repository: invalid reference format
Error: 
   0: `podman buildx build --platform linux/amd64 --load --pull --cache-from 'type=registry,ref=ghcr.io/cross-rs/aarch64-linux-android:main' --tag ghcr.io/cross-rs/aarch64-linux-android:local --label 'org.cross-rs.for-cross-target=aarch64-linux-android' --label 'org.cross-rs.runs-with=x86_64-unknown-linux-gnu' -f /home/ahuszagh/git/cross/docker/Dockerfile.aarch64-linux-android --progress auto .` failed with exit status: 125

Location:
   /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/convert/mod.rs:55
```

aarch64-linux-android

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [X] podman

### cross version

cross 0.2.4 (d92e563cdded42aa5468746fc821f9b795880ce2)

Removing the `--cache-from` step fixes the issue. We should probably also add a build image step for `podman`, using the fast image build (`x86_64-unknown-linux-gnu`) to avoid this issue in the future. The podman version is `4.2.0`.

Alexhuszagh:
Ensures that docker and nerdctl, which support the custom `--output` and `--cache-from` flags use them, while podman and unknown container engines use the strict, minimal subset podman supports. This is because podman only supports a registry/repository, without a tag, for `--cache-from`, which means it synchronizes poorly with our target subs, like `centos`. Therefore, unless unsupported, we should always use the features available for our container engine.

This also fixes the `--pull` flag on nerdctl, which is unsupported. Engines without BuildKit extensions by default, such as nerdctl, now do not use `buildx` unless explicitly enabled with `CROSS_CONTAINER_ENGINE_NO_BUILDKIT=0`.

Closes #1031.

### DIFF
--- a/src/docker/engine.rs
+++ b/src/docker/engine.rs
@@ -17,20 +17,44 @@ pub enum EngineType {
     Docker,
     Podman,
     PodmanRemote,
+    Nerdctl,
     Other,
 }
 
 impl EngineType {
     /// Returns `true` if the engine type is [`Podman`](Self::Podman) or [`PodmanRemote`](Self::PodmanRemote).
     #[must_use]
-    pub fn is_podman(&self) -> bool {
+    pub const fn is_podman(&self) -> bool {
         matches!(self, Self::Podman | Self::PodmanRemote)
     }
 
     /// Returns `true` if the engine type is [`Docker`](EngineType::Docker).
     #[must_use]
-    pub fn is_docker(&self) -> bool {
+    pub const fn is_docker(&self) -> bool {
         matches!(self, Self::Docker)
+    }
+
+    /// Returns `true` if the build command supports the `--output` flag.
+    #[must_use]
+    pub const fn supports_output_flag(&self) -> bool {
+        !matches!(self, Self::Other)
+    }
+
+    /// Returns `true` if the build command supports the `--pull` flag.
+    #[must_use]
+    pub const fn supports_pull_flag(&self) -> bool {
+        !matches!(self, Self::Nerdctl | Self::Other)
+    }
+
+    /// Returns `true` if the build command supports the `--cache-from type=` key.
+    ///
+    /// Some container engines, especially podman, do not support the `type`
+    /// key of `--cache-from` during the image build steps. They also do
+    /// not support any tags for the `--cache-from` steps either. See:
+    /// https://docs.podman.io/en/latest/markdown/podman-build.1.html#cache-from
+    #[must_use]
+    pub const fn supports_cache_from_type(&self) -> bool {
+        matches!(self, Self::Docker | Self::Nerdctl)
     }
 }
 
@@ -132,6 +156,8 @@ fn get_engine_info(
         EngineType::PodmanRemote
     } else if stdout_help.contains("podman") {
         EngineType::Podman
+    } else if stdout_help.contains("nerdctl") {
+        EngineType::Nerdctl
     } else if stdout_help.contains("docker") && !stdout_help.contains("emulate") {
         EngineType::Docker
     } else {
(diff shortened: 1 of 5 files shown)


Alexhuszagh:

Alexhuszagh:
bors try-

Alexhuszagh:

Alexhuszagh:
bors try-

Alexhuszagh:

Alexhuszagh:

I've added a custom image build to the Podman test (it's a trivial pre-build step, so it just tests the image build steps work), so we can ensure building and testing images work.

Alexhuszagh:
I haven't added tests for Lima (`nerdctl`), since it requires a Homebrew installation which takes a long, long time and sometimes errors due to too many file descriptors being open. I have, however, tested this locally and confirmed the above flags work. We're currently waiting on https://github.com/lima-vm/lima/discussions/1075 to see if we can get BuildKit working with Lima.

Alexhuszagh:
We might also want to default to no `buildx` with Lima, since it doesn't support it.

src/docker/engine.rs
```diff
@@ -32,6 +33,29 @@ impl EngineType {
     pub fn is_docker(&self) -> bool {
         matches!(self, Self::Docker)
     }
+
+    /// Returns `true` if the engine build supports the `--output` flag.
```
Emilgardis, line 37: I think the wording here is a bit awkward, maybe
```suggestion
    /// Returns `true` if the build command supports the `--output` flag.
```

xtask/src/build_docker_image.rs
```diff
@@ -187,7 +187,7 @@ pub fn build_docker_image(
 
         if push {
             docker_build.arg("--push");
-        } else if engine.kind.is_docker() && no_output {
+        } else if engine.kind.supports_output_flag() && no_output {
```
Emilgardis, line 190: I feel like this should be an error, i.e it should not be possible to specify no output and then not have it happen

src/docker/engine.rs
```diff
@@ -32,6 +33,29 @@ impl EngineType {
...
+    #[must_use]
+    pub fn supports_output_flag(&self) -> bool {
```
Emilgardis, line 39: Can be const, same goes for the other fns

Emilgardis (changes requested):

src/docker/engine.rs
Emilgardis: Same for the other fns

xtask/src/build_docker_image.rs
Alexhuszagh: I've changed it so we error if it doesn't support the output flag, with another branch:

```rust
if push {
    docker_build.arg("--push");
} else if engine.kind.supports_output_flag() && no_output {
    docker_build.args(["--output", "type=tar,dest=/dev/null"]);
} else if no_output {
    msg_info.fatal("cannot specify `--no-output` with engine that does not support the `--output` flag", 1);
} else if has_buildkit {
    docker_build.arg("--load");
}
```

Emilgardis (approved):
lgtm!

bors r+

PR #1038, opened 2022-09-29
upgrade to clap 4.0

Emilgardis:


Alexhuszagh:
bors try --dylib 1 --std 0
Just trying with 0 targets hopefully without failing to see if the crates.io index has been updated.

Emilgardis:
this is super weird, there is no cache hit, and it still fails.

Emilgardis:
aha, I think this is why

https://github.com/clap-rs/clap/blob/39fe46e650e478627a3df72fd3ca7fdeac507032/Cargo.toml#L26

I've just never seen this kind of error message, is that how it's supposed to work?

Alexhuszagh:
> aha, I think this is why
> 
> https://github.com/clap-rs/clap/blob/v4.0.9/Cargo.toml#L26
> 
> I've just never seen this kind of error message, is that how it's supposed to work?

I think that's a new feature, which it would give a better diagnostic message. Is it worth updating the MSRV here? Or should we wait for 1 more major release?

Emilgardis:
> Is it worth updating the MSRV here? Or should we wait for 1 more major release?

our msrv policy is:

> This crate is guaranteed to compile on stable Rust 1.58.1 and up. It might compile with older versions but that may change in any new patch release.

I say lets do it :)

Alexhuszagh:
I know this sounds silly but can we wait 3 days? Also, then we need a README update? That way, 1.60 will have been 6 months old by the time we update the MSRV.

Emilgardis:
> Also, then we need a README update?

What do you mean? I've changed the mentioned msrv in the readme 👍🏼

Alexhuszagh (approved):
lgtm, squash the commit history and merge when ready

Emilgardis:
Id prefer leaving them separated, can do though

Alexhuszagh:
> Id prefer leaving them separated, can do though

that's fine, just r me when ready

Emilgardis:
bors r=Alexhuszagh

PR #1039, opened 2022-09-29
Add support for other overlay filesystems.

Alexhuszagh:
This allows other storage engines, such as `overlay` and `fuse-overlayfs` to be used with `CROSS_CONTAINER_IN_CONTAINER`. Any overlay filesystem can be supported, and this has been tested with both `overlay` and `overlay2`.

### DIFF
--- a/.changes/1039.json
+++ b/.changes/1039.json
@@ -1,0 +1,4 @@
+{
+    "description": "support overlay and fuse-overlayfs storage drivers",
+    "type": "added"
+}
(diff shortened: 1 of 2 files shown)


Alexhuszagh:
The rationale for why this works for any overlay filesystem is documented [here](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt). All overlay filesystems should have an upper, a lower, and a merged directory.

> An overlay filesystem combines two filesystems - an 'upper' filesystem and a 'lower' filesystem.  When a name exists in both filesystems, the object in the 'upper' filesystem is visible while the object in the 'lower' filesystem is either hidden or, in the case of directories, merged with the 'upper' object.

> Where both upper and lower objects are directories, a merged directory is formed.

Emilgardis (approved):

PR #1042, opened 2022-10-02
fix `cargo llvm-cov` to use new `report` subcommand

Emilgardis:


Emilgardis:
bors try --none

Alexhuszagh:
This will always install the latest version of `cargo-llvm-cov`, correct?

Emilgardis:
> This will always install the latest version of `cargo-llvm-cov`, correct?

Correct, the way to circumvent that is documented here:
https://github.com/taiki-e/install-action#example-workflow

Alexhuszagh:
We might want to lock it to a specific version of `cargo-llvm-cov` then just in case, since breaking versions changes might break our CI. `tool: cargo-llvm-cov@0.5` so we only match [minor versions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions) of `0.5` might be a good idea here.

Alexhuszagh:
> ## try
> 
> Build failed:
> 
>     * [conclusion](https://github.com/cross-rs/cross/actions/runs/3169232460/jobs/5161096346)

It worked, it just seems `--none` will always fail due to an empty matrix.

Emilgardis:
> It worked, it just seems --none will always fail due to an empty matrix.

should be easy to fix by inspecting the needs, will fix eventually

Alexhuszagh:
My mistake: apparently `install-action` doesn't follow Github Actions version syntax, it requires an exact version for getting the binary and doesn't match based on the semver. It's probably easier to just remove the version on `cargo-llvm-cov` and fix it later if we need to:

https://github.com/cross-rs/cross/actions/runs/3169293623/jobs/5161102311#step:4:31

Alexhuszagh (approved):
lgtm, just squash the commit history and merge when ready

Emilgardis:
bors r=Alexhuszagh

PR #1044, opened 2022-10-02
Document remote container termination safety guarantees.

Alexhuszagh:
Updates the comment documentation for the async-signal safety guarantees in stopping a running container.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -87,6 +87,21 @@ unsafe fn termination_handler() {
     //
     // on windows, a non-reentrant static mutex is used, so it is
     // definitely not thread safe, but this should not matter.
+    //
+    // NOTE: there is one major function that is not async-signal safe here:
+    // memory allocation and deallocation, which is not async-signal safe.
+    // this should only be run once without deadlocking since any
+    // atomics are guaranteed to be lock-free. we cannot easily avoid
+    // allocation/deallocation, since we would need static global muts
+    // for basically everything. `Command::arg` and `Command::new` will
+    // internally allocate, and freeing it will deallocate any arguments
+    // it was provided. even if we used a static global `Command`, the
+    // `io::Result` requires a `Box` or `io::Error`, which would allocate.
+    // the alternative would be to use `posix_spawnp` or `CreateProcess`
+    // directly, which are async-signal safe and thread-safe, respectively,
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
I double checked the entirety of `Command` for both Windows and UNIX processes to determine if there was anything potentially missing here for how we run this. There's a few minor issues that weren't documented:

**Async-Signal Unsafe**
- `Engine::drop`: `Engine` contains a `PathBuf`, which is allocated.
- `Command::new`, which creates a `CString` from an `OsStr`.
- `Command::arg`, which creates a `CString` from an `OsStr`.
- `Command::args`, which does the same as `Command::arg` for each argument.
- `DeleteContainer::drop`: Contains `Engine` and `String`, which are allocated. Also invokes `container_stop` and `container_rm`, which allocate since they create a `Command`.

**Async-Signal Safe**
- `Command::capture_envp`: since we don't modify the environment, this won't allocate.
- `Command::spawn`: as long as we don't modify the environment, `Command::capture_envp` won't allocate, and therefore neither will `spawn` as long as we're not using pipes (which we aren't).

As documented, it would be possible to make this entirely async-signal safe even without the atomic guard, however, this shouldn't be necessary.

src/errors.rs
```diff
@@ -87,6 +87,21 @@ unsafe fn termination_handler() {
...
+    // allocation/deallocation, since we would need static global muts
+    // for basically everything. `Command::arg` and `Command::new` will
+    // internally allocate, and freeing it will deallocate any arguments
+    // it was provided. even if we used a static global `Command`, the
+    // `io::Result` requires a `Box` or `io::Error`, which would allocate.
+    // the alternative would be to use `posix_spawnp` or `CreateProcess`
+    // directly, which are async-signal safe and thread-safe, respectively,
+    // however, we'd need to store the engine path and the argument list as
+    // a global CString and `Vec<CString>`, respectively. since we have an
+    // atomic guard here, this should be safe regardless.
```
Emilgardis, line 104:
```suggestion
    // a global CString and `Vec<CString>`, respectively. This atomic guard makes this safe regardless.
```

Emilgardis (approved):
r=me on change

Alexhuszagh:
bors r=Emilgardis

Issue #1046, opened 2022-10-03
aarch64-unknown-linux-musl build failed due to "CROSS_RUSTC_MINOR_VERSION: unbound variable" error


taiki-e:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

https://github.com/cross-rs/cross/commit/9311417f7d251103c21f547aa341e643f08a5fd8 added `CROSS_RUSTC_MINOR_VERSION` environment variables without fallback, so the current docker image of the `main` tag is incompatible with the released version of cross (v0.2.4).

log: https://github.com/taiki-e/cargo-llvm-cov/actions/runs/3170903254/jobs/5163844829

```
error: linking with `aarch64-linux-musl-gcc.sh` failed: exit status: 1
  |
[...]

PR #1047, opened 2022-10-03
Provide fallback for rustc version when envvars not present.

Alexhuszagh:
Fixes an issue where `CROSS_RUSTC_MINOR_VERSION` is not present when using older versions of cross but newer images. This only uses the fallback if the minor version is not provided as an environment variable by cross itself.

### DIFF
--- a/docker/rustc_info.sh
+++ b/docker/rustc_info.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# FIXME: remove this file further on, especially after
+# 0.2.5 has been released and we've had a weeks of
+# releases on main so people can update images.
+# people may use newer versions of cross with older
+# images for backwards compatibility, but we don't
+# need to guarantee newer images work with older cross
+# versions.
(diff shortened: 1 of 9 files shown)


Alexhuszagh:
This is a temporary fallback to try to fix #1046 and cases where newer images are used with an older version of `cross` (we should support the inverse, but normally using newer images with an older version of `cross` is unsupported).

This does not fix the much broader issue: `bininstall` and the pre-compiled releases are not using the correct version tags for images.

Alexhuszagh:
bors try --target aarch64-unknown-linux-musl

docker/rustc_info.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
Alexhuszagh, line 1: This file is intentionally non-executable and named differently than `rustc.sh` just to avoid any conflicts or confusion with `rustc` itself.

Emilgardis (approved):
bors r+

PR #1052, opened 2022-10-03
Update cross-util program name and version.

Alexhuszagh:
Updates to output of `cross-util --version` to use the program name `cross-util`, and in development builds, add the commit hash (so, for example, the output might be `cross-util 0.2.4 (8d8d84b 2022-09-08)`).


src/bin/cross-util.rs
```diff
@@ -6,8 +6,24 @@ use cross::{docker, rustc::Toolchain};
 
 mod commands;
 
+const APP_NAME: &str = "cross-util";
+static mut VERSION: String = String::new();
+
+fn version() -> &'static str {
+    // safe since single-threaded execution
+    unsafe {
```
Alexhuszagh, line 14: I've used to avoid the `once_cell` dependency in the main crate. I can change it if desired.

Emilgardis: yes please, let's use once_cell, soon it's in stable also so that's nice :)

however, is it really needed? Think it would be fine to do `name = "cross-util", version = format!("{}{}", env!("CARGO_PKG_VERSION"), cross::commit_info())`

Emilgardis (changes requested):

Emilgardis:
Even better, on clap 4.0, with the `string` feature enabled this is possible

```
#[clap(about, long_about = None, name = APP_NAME, version = format!(
    concat!(env!("CARGO_PKG_VERSION"), "{}"),
    cross::commit_info()
))]
```

Alexhuszagh:
Nice, the clap 4.0 migration will be very handy. I've updated it to move `once_cell` back to a dev dependency, removed the unnecessary `format_args` since we can just use `concat!` with `commit_info!` as a macro directly.

Emilgardis (approved):
bors r+

PR #1059, opened 2022-10-05
Ensure successful CI conclusion with empty matrix.

### LINKED ISSUE #1043: CI With Empty Matrix Caused Failed Conclusion
Running a build with an empty matrix, such as [bors try --none](https://github.com/cross-rs/cross/pull/1042#issuecomment-1264681923) causes the conclusion to fail, even if all other steps pass. Since we can always inspect manually the failure, there's little reason to prioritize this.

Alexhuszagh:
Closes #1043.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-rust
+
       - name: Generate matrix
         id: generate-matrix
         run: cargo xtask ci-job target-matrix
@@ -102,7 +103,7 @@ jobs:
     name: target (${{ matrix.pretty }},${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     needs: [shellcheck, test, generate-matrix, check]
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && needs.generate-matrix.outputs.matrix != '{}' && needs.generate-matrix.outputs.matrix != '[]' && needs.generate-matrix.outputs.matrix != ''
     strategy:
       fail-fast: false
       matrix:
--- a/xtask/src/ci/target_matrix.rs
+++ b/xtask/src/ci/target_matrix.rs
@@ -32,7 +32,9 @@ pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Rep
             std: target.std.map(|b| b as u8),
         })
         .collect::<Vec<_>>();
-    gha_output("matrix", &serde_json::to_string(&matrix)?);
+    let json = serde_json::to_string(&matrix)?;
+    gha_print(&json);
+    gha_output("matrix", &json);
     Ok(())
 }
 


Alexhuszagh:
bors try --none

Alexhuszagh:
bors try-
bors try --none

Alexhuszagh:
bors try --none

Alexhuszagh:
bors try --none

.github/workflows/ci.yml
```diff
@@ -102,7 +103,8 @@ jobs:
     name: target (${{ matrix.pretty }},${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     needs: [shellcheck, test, generate-matrix, check]
-    if: github.event_name == 'push'
+    if: false
+    #if: github.event_name == 'push' && ${{ needs.generate-matrix.outputs.matrix != '{}' && needs.generate-matrix.outputs.matrix != '[]' && needs.generate-matrix.outputs.matrix != '' }}
```
Alexhuszagh, line 107: I think my logic is invalid here, since just skipping the `build` step altogether ensures the empty matrix doesn't occur.

Emilgardis: the problem I think is that

`conclusions` is `if: always()`, and in its needs, `build` is included. if `build` is skipped, but has elements in it's matrix, then it will run, if its matrix is empty and not skipped, it will fail (for some reason)

best way to solve this is to make conclusion see if `needs.generate-matrix.outputs.matrix` is empty, and if it is mark `build` as succeeded (on try only)

Alexhuszagh: That's why my logic is off: the build step is supposed to be skipped if the matrix is empty.

Alexhuszagh:
bors try --target none

.github/workflows/ci.yml
```diff
@@ -102,7 +103,7 @@ jobs:
     name: target (${{ matrix.pretty }},${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     needs: [shellcheck, test, generate-matrix, check]
-    if: github.event_name == 'push'
+    if: ${{ github.event_name == 'push' && needs.generate-matrix.outputs.matrix != '{}' && needs.generate-matrix.outputs.matrix != '[]' && needs.generate-matrix.outputs.matrix != '' }}
```
Emilgardis, line 106:
```suggestion
    if: github.event_name == 'push' && needs.generate-matrix.outputs.matrix != '{}' && needs.generate-matrix.outputs.matrix != '[]' && needs.generate-matrix.outputs.matrix != ''
```

Emilgardis, line 106: I think this should take into account if we are in a try or not, and only allow empty on a try

Emilgardis (changes requested):
nice!

.github/workflows/ci.yml
Alexhuszagh: I have no idea how to do that in an `if` statement (without a run branch).

Emilgardis:
```yml
if: |
  github.event_name == 'push' && 
  (needs.generate-matrix.outputs.matrix != '{}' && needs.generate-matrix.outputs.matrix != '[]' && needs.generate-matrix.outputs.matrix != '' && (github.ref == 'refs/heads/trying' || true))
```

seems correct?

if push, and matrix not empty or not in trying

Alexhuszagh: We could probably have the preceeding step store an envvar and check `env.MATRIX_IS_EMPTY` or something? Since `|| true` will always be true. Would make the statement easier then too.

Emilgardis: yes, or even simpler, have generate-matrix output an object `{"empty": true}` on empty when not in try

Emilgardis: even even simpler, have generate-matrix fail when it's empty on a non-trying branch

Emilgardis:
```
diff --git a/xtask/src/ci.rs b/xtask/src/ci.rs
index 012be66..ec348f5 100644
--- a/xtask/src/ci.rs
+++ b/xtask/src/ci.rs
@@ -27,6 +27,9 @@ pub enum CiJob {
         ref_name: String,
     },
     TargetMatrix {
+        // main, v0.1.0
+        #[clap(long, env = "GITHUB_REF_NAME")]
+        ref_name: String,
         #[clap(long, env = "COMMIT_MESSAGE")]
         message: String,
         #[clap(long, env = "COMMIT_AUTHOR")]
@@ -114,8 +117,12 @@ pub fn ci(args: CiJob, metadata: CargoMetadata) -> cross::Result<()> {
                 }
             }
         }
-        CiJob::TargetMatrix { message, author } => {
-            target_matrix::run(message, author)?;
+        CiJob::TargetMatrix {
+            ref_name,
+            message,
+            author,
+        } => {
+            target_matrix::run(ref_name, message, author)?;
         }
     }
     Ok(())
diff --git a/xtask/src/ci/target_matrix.rs b/xtask/src/ci/target_matrix.rs
index e8c87dd..6db8446 100644
--- a/xtask/src/ci/target_matrix.rs
+++ b/xtask/src/ci/target_matrix.rs
@@ -4,7 +4,11 @@ use serde::Serialize;
 
 use crate::util::{get_matrix, gha_output, gha_print, CiTarget, ImageTarget};
 
-pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Report> {
+pub(crate) fn run(
+    ref_name: String,
+    message: String,
+    author: String,
+) -> Result<(), color_eyre::Report> {
     let mut matrix: Vec<CiTarget> = get_matrix().clone();
     if author == "bors[bot]" && message.starts_with("Try #") {
         if let Some((_, args)) = message.split_once(": ") {
@@ -32,7 +36,11 @@ pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Rep
             std: target.std.map(|b| b as u8),
         })
         .collect::<Vec<_>>();
-    let json = serde_json::to_string(&matrix)?;
+    let json = if !matrix.is_empty() || ref_name == "trying" {
+        serde_json::to_string(&matrix)?
+    } else {
+        eyre::bail!("matrix can't be empty")
+    };
     gha_print(&json);
     gha_output("matrix", &json);
     Ok(())
```

Emilgardis: actually though, this point seems kinda moot, if it is empty then `targets.toml` is empty, which seems not likely

disregard this :)

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #1070, opened 2022-10-08
Warn when mounted volume likely uses a path rather than an environment variable.

### LINKED ISSUE #1069: warn when env.volumes is probably a path and not a `var = path`
Related to #1068

https://github.com/cross-rs/cross/blob/683f1dbb23607e273e303840e74bd955cf12af37/src/docker/shared.rs#L571-L584

should warn when the env var is not defined/looks like a path

Alexhuszagh:
Closes #1069.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -568,11 +568,29 @@ pub(crate) fn register(engine: &Engine, target: &Target, msg_info: &mut MessageI
     docker.run(msg_info, false).map_err(Into::into)
 }
 
-fn validate_env_var(var: &str) -> Result<(&str, Option<&str>)> {
+fn validate_env_var<'a>(
+    var: &'a str,
+    warned: &mut bool,
+    var_type: &'static str,
+    var_syntax: &'static str,
+    msg_info: &mut MessageInfo,
+) -> Result<(&'a str, Option<&'a str>)> {
     let (key, value) = match var.split_once('=') {
         Some((key, value)) => (key, Some(value)),
         _ => (var, None),
     };
+
+    if value.is_none()
+        && !*warned
+        && !var
+            .chars()
+            .all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '_' ))
+    {
+        msg_info.warn(format_args!(
+            "got {var_type} of \"{var}\" which is not a valid environment variable name. the proper syntax is {var_syntax}"
+        ))?;
+        *warned = true;
+    }
 
     if key == "CROSS_RUNNER" {
         eyre::bail!(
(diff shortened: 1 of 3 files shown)


src/docker/shared.rs
```diff
@@ -714,13 +714,24 @@ pub(crate) fn docker_mount(
     paths: &DockerPaths,
     mount_cb: impl Fn(&mut Command, &Path, &Path) -> Result<()>,
     mut store_cb: impl FnMut((String, String)),
+    msg_info: &mut MessageInfo,
 ) -> Result<()> {
+    let mut warned = false;
```
Alexhuszagh, line 719: Avoid multiple warnings, since a single warning should be sufficient to notify the user that their syntax is invalid.

src/docker/shared.rs
```diff
@@ -714,13 +714,24 @@ pub(crate) fn docker_mount(
...
         .unwrap_or_default()
     {
         let (var, value) = validate_env_var(var)?;
+        if value.is_none()
+            && !warned
+            && !var
+                .chars()
+                .all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '_' ))
+        {
+            msg_info.warn(format_args!("got volume of \"{var}\" which is not an environment variable name. the proper syntax is `volumes = [\"ENVVAR=/path/to/directory\"]`"))?;
```
Emilgardis, line 732: should this warn or error?

im fine with either

Emilgardis (commented):
I feel like this should live in `validate_env_var` given its name

Alexhuszagh:
> I feel like this should live in `validate_env_var` given its name

We'd probably need to have separate warning messages however for each, since a warning for a volume shouldn't be the same as a passthrough environment variable.

src/docker/shared.rs
Alexhuszagh: I think it should warn. It's technically valid input to the Docker container, it just won't behave as expected.

Alexhuszagh:
The new version produces the following diagnostics:

```bash
$ cross build --target x86_64-unknown-linux-gnu
[cross] warning: got environment variable of "v:y" which is not a valid environment variable name. the proper syntax is `passthrough = ["ENVVAR=value"]`
[cross] warning: got volume of "../rust-c-hello-world" which is not a valid environment variable name. the proper syntax is `volumes = ["ENVVAR=/path/to/directory"]`
```

The check has been moved into `validate_env_var`, which now requires a `var_type` and `var_syntax` argument to properly format the warning.

Emilgardis (approved):
lgtm!

bors r+

PR #964, opened 2022-07-19
Do not keep stdin open when running containers.

Alexhuszagh:

PR #1071, opened 2022-10-08
Terminate child containers on program termination.

### LINKED ISSUE #1041: CTRL-C does not kill cross
### Checklist

### Describe your issue

Hitting CTRL-C on my macbook fails to kill a `cross build`. Instead, I must kill this process with `kill <PID>` or `docker kill <CONTAINER>`

x86_64-unknown-linux-gnu

- [X] macOS
- [ ] Windows
- [ ] Linux / BSD

- [ ] x86_64 / AMD64
- [ ] arm32
- [X] arm64 (including Mac M1)

- [X] docker
- [ ] podman

### cross version

cross 0.2.4,  cargo 1.66.0-nightly (73ba3f35e 2022-09-18)

### Example

```sh
cross build --target x86_64-unknown-linux-gnu 
```

Everything works correctly when the `aarch64-apple-darwin` (host) target is used instead

Alexhuszagh:
Upon receiving a termination signal, ensuring any spawned containers are also terminated. This is required since we intentional do not attach stdint to the container. This means, however, that interrupts such as Ctrl+C did not stop the build process. Likewise, termination signals such as SIGTERM merely exited cross, and did not terminate the container.

This also refactors the global container state to store whether the container exists and current logic to terminate the container if it exists. By coupling the two, it simplifies safe handling and modification of this global state. Since we have a signal handler, these functions may be called asynchronously, therefore any global state may be modified, rendering the current state invalid if not appropriately guarded. Coupling the two ensures that our logic is sound.

Fixes the regression in #964.
Closes #1041.

### DIFF
--- a/src/docker/local.rs
+++ b/src/docker/local.rs
@@ -1,6 +1,7 @@
 use std::io;
 use std::path::Path;
 use std::process::{Command, ExitStatus};
+use std::sync::atomic::Ordering;
 
 use super::shared::*;
 use crate::errors::Result;
@@ -50,6 +51,8 @@ pub(crate) fn run(
         msg_info,
     )?;
 
+    let container = toolchain_dirs.unique_container_identifier(options.target.target())?;
+    docker.args(["--name", &container]);
     docker.arg("--rm");
 
     docker_seccomp(&mut docker, engine.kind, &options.target, &paths.metadata)
@@ -121,9 +124,22 @@ pub(crate) fn run(
             .wrap_err("when building custom image")?;
     }
 
-    docker
+    Container::create(engine.clone(), container)?;
+    let status = docker
         .arg(&image_name)
         .args(["sh", "-c", &build_command(toolchain_dirs, &cmd)])
         .run_and_get_status(msg_info, false)
-        .map_err(Into::into)
+        .map_err(Into::into);
+
+    // `cargo` generally returns 0 or 101 on completion, but isn't guaranteed
+    // to. `ExitStatus::code()` may be None if a signal caused the process to
+    // terminate or it may be a known interrupt return status (130, 137, 143).
+    // simpler: just test if the program termination handler was called.
+    // SAFETY: an atomic load.
+    let is_terminated = unsafe { crate::errors::TERMINATED.load(Ordering::SeqCst) };
+    if !is_terminated {
+        Container::exit_static();
+    }
+
+    status
 }
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -549,6 +549,19 @@ impl Container {
         unsafe { CONTAINER.exists() }
     }
 
+    // when the `docker run` command finished.
+    // the container has already exited, so no cleanup required.
+    pub fn exit(&mut self) {
+        self.exists.store(false, Ordering::SeqCst);
+    }
+
+    pub fn exit_static() {
+        // SAFETY: an atomic store.
+        unsafe {
+            CONTAINER.exit();
+        }
+    }
+
     // when the `docker exec` command finished.
     pub fn finish(&mut self, is_tty: bool, msg_info: &mut MessageInfo) {
         // relax the no-timeout and lack of output
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -1,6 +1,7 @@
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, ExitStatus, Output};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{env, fs, time};
 
 use super::custom::{Dockerfile, PreBuild};
@@ -15,8 +16,8 @@ use crate::extensions::{CommandExt, SafeCommand};
 use crate::file::{self, write_file, PathExt, ToUtf8};
 use crate::id;
 use crate::rustc::QualifiedToolchain;
-use crate::shell::{MessageInfo, Verbosity};
-use crate::{CargoVariant, Target, TargetTriple};
+use crate::shell::{ColorChoice, MessageInfo, Verbosity};
+use crate::{CargoVariant, OutputExt, Target, TargetTriple};
 
 use rustc_version::Version as RustcVersion;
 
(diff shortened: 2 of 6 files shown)


Emilgardis:
#964 isn't in a published version, therefor this changelog entry should not be added as a fix

src/docker/shared.rs
```diff
@@ -488,6 +489,284 @@ impl Directories {
...
+impl Container {
+    pub const fn new() -> Container {
+        Container {
+            deleter: None,
+            exists: AtomicBool::new(false),
+        }
+    }
+
+    pub fn create(engine: Engine, name: String) -> Result<()> {
+        // SAFETY: safe, since guarded by an atomic swap.
```
Emilgardis, line 523: safety comments should explain why things are safe, not that it is safe

```suggestion
        // SAFETY: guarded by an atomic swap
```

i know this is how it was previously but lets change it now :D

src/docker/shared.rs
```diff
@@ -488,6 +489,284 @@ impl Directories {
...
+
+    // the static functions have been placed by the internal functions to
+    // verify the internal functions are wrapped in atomic load/stores.
+
+    pub fn exists(&self) -> bool {
+        self.exists.load(Ordering::Relaxed)
+    }
+
+    pub fn exists_static() -> bool {
+        // SAFETY: safe, since it's an atomic load.
```
Emilgardis, line 548: ditto

src/docker/shared.rs
```diff
@@ -488,6 +489,284 @@ impl Directories {
...
+    }
+
+    // when the `docker run` command finished.
+    // the container has already exited, so no cleanup required.
+    pub fn exit(&mut self) {
+        self.exists.store(false, Ordering::SeqCst);
+    }
+
+    pub fn exit_static() {
+        // SAFETY: safe, an atomic store on a static variable is fine.
```
Emilgardis, line 559: ditto

src/docker/shared.rs
```diff
@@ -488,6 +489,284 @@ impl Directories {
...
+            }
+            deleter.color_choice = msg_info.color_choice;
+            deleter.verbosity = msg_info.verbosity;
+        }
+
+        self.terminate();
+    }
+
+    pub fn finish_static(is_tty: bool, msg_info: &mut MessageInfo) {
+        // SAFETY: safe, since internally guarded by an atomic.
```
Emilgardis, line 582: ditto

Emilgardis (commented):
having a hard time seeing what the actual change is, could you split out the move into `shared` and the fix into two commits?

Alexhuszagh:
> having a hard time seeing what the actual change is, could you split out the move into `shared` and the fix into two commits?

Definitely. Sorry, this was made more confusing than it needed to be.

Alexhuszagh:
I've separated out the two commits, one for the refactoring and one for the change. The refactoring makes this quite easy to do, since all we need is to attach a name to the container, and if there's an interrupt, we handle the container termination like we do when using remote docker. If the docker command exits successful, IE, with no interrupts, then we simply mark that the command has completed.

src/docker/local.rs
```diff
@@ -121,9 +123,16 @@ pub(crate) fn run(
...
-    docker
+    Container::create(engine.clone(), container)?;
+    let status = docker
         .arg(&image_name)
         .args(["sh", "-c", &build_command(toolchain_dirs, &cmd)])
         .run_and_get_status(msg_info, false)
-        .map_err(Into::into)
+        .map_err(Into::into);
+
+    if status.as_ref().map(|x| x.success()).unwrap_or_default() {
```
Alexhuszagh, line 133: This actually should probably only ignore the specific termination signals handled by `ctrlc`, so anything other than those codes is consider "success" (IE, a termination handler wasn't called).

src/docker/shared.rs
```diff
@@ -488,6 +489,271 @@ impl Directories {
...
+    }
+
+    // the static functions have been placed by the internal functions to
+    // verify the internal functions are wrapped in atomic load/stores.
+
+    pub fn exists(&self) -> bool {
+        self.exists.load(Ordering::Relaxed)
+    }
+
+    pub fn exists_static() -> bool {
```
Alexhuszagh, line 547: I named these `foo` and `foo_static`, but there's almost certainly a better naming convention.

src/docker/shared.rs
```diff
@@ -488,6 +489,271 @@ impl Directories {
...
+
+    #[must_use]
+    pub fn exists(&self) -> bool {
+        !matches!(self, Self::DoesNotExist)
+    }
+}
+
+// it's unlikely that we ever need to erase a line in the destructors,
+// and it's better than keep global state everywhere, or keeping a ref
+// cell which could have already deleted a line
```
Emilgardis, line 653: this comment doesn't make sense without context of how `cross::shell` works, could be removed or just mention the reason for no `MessageInfo` is because it needs mutability

src/docker/shared.rs
```diff
@@ -488,6 +489,271 @@ impl Directories {
...
+            }
+            deleter.color_choice = msg_info.color_choice;
+            deleter.verbosity = msg_info.verbosity;
+        }
+
+        self.terminate();
+    }
+
+    pub fn finish_static(is_tty: bool, msg_info: &mut MessageInfo) {
+        // SAFETY: internally guarded by an atomic.
```
Emilgardis, line 582:
```suggestion
        // SAFETY: internally guarded by an atomic swap.
```

src/docker/shared.rs
```diff
@@ -488,6 +489,271 @@ impl Directories {
...
+    msg_info: &mut MessageInfo,
+) -> Result<ExitStatus> {
+    // we want a faster timeout, since this might happen in signal
+    // handler. our containers normally clean up pretty fast, it's
+    // only without a pseudo-tty that they don't.
+    container_stop(engine, container, DEFAULT_TIMEOUT, msg_info)
+}
+
+// if stop succeeds without a timeout, this can have a spurious error
+// that is, if the container no longer exists. just silence this.
```
Emilgardis, line 705: another one I missed in the previous pr

this comment feels misplaced

move it to a doccomment or at the call site

src/docker/shared.rs
```diff
@@ -541,14 +541,27 @@ impl Container {
     // verify the internal functions are wrapped in atomic load/stores.
 
     pub fn exists(&self) -> bool {
-        self.exists.load(Ordering::Relaxed)
+        self.exists.load(Ordering::SeqCst)
```
Emilgardis, line 544: nice

Emilgardis (changes requested):

src/docker/shared.rs
Alexhuszagh: It's actually an atomic load.

src/docker/shared.rs
Alexhuszagh: Is the new comment better? I've changed it to:

```rust
// we need to specify drops for the containers, but we
// also need to ensure the drops are called on a
// termination handler. we use an atomic bool to ensure
// that the drop only gets called once, even if we have
// the signal handle invoked multiple times or it fails.
```

src/docker/shared.rs
Alexhuszagh: Since we call `container_rm` from multiple locations, I've updated the comment to better explain why and made it a doc comment:

```rust
/// if stopping a container succeeds without a timeout, this command
/// can fail because the container no longer exists. however, if
/// the container was killed, we need to cleanup the exited container.
/// just silence any warnings.
```

src/docker/shared.rs
Emilgardis: but the mechanic that makes it safe is that the operations are gated by a bool that is swapped, no?

Alexhuszagh: I guess a load and swap? The swap is for `terminate`. The load is for modifying the timeout and the verbosity/color choice.

src/docker/shared.rs
```diff
@@ -488,6 +489,287 @@ impl Directories {
...
+    #[must_use]
+    pub fn exists(&self) -> bool {
+        !matches!(self, Self::DoesNotExist)
+    }
+}
+
+// the lack of MessageInfo is because we need a mutable reference to
+// store if we need to erase a line from stdout or stderr. since our
+// cleanup will not need to erase a line, we can just store the basic
+// MessageInfo configurations.
```
Emilgardis, line 654:
```suggestion
// the lack of [MessageInfo] is because it'd require a mutable reference,
// since we don't need the functionality behind the [MessageInfo], we can just store the basic
// MessageInfo configurations.
```

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #1072, opened 2022-10-10
Fix persistent data volumes with remote cross.

### LINKED ISSUE #1016: `CROSS_REMOTE=1` with persistent volume doesn't change project files
### Checklist

### Describe your issue

Using `CROSS_REMOTE=1` with persistent volume, changing files in the project does not carry over across the first remote build.

_No response_

- [X] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32
- [X] arm64 (including Mac M1)

- [X] docker
- [ ] podman

### cross version

8d8d84b

### Example

```
$ cross-util volumes create
$ cargo new --bin issue1016
$ cd issue1016
$ CROSS_REMOTE=1 cross run
...
   Compiling issue1008 v0.1.0 (/cross/Users/emil/workspace/issue1016)
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `/linux-runner x86_64 target/x86_64-unknown-linux-gnu/debug/issue1016`
Hello, world!
$ # edit src/main.rs to say something else
$ CROSS_REMOTE=1 cross run
...
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `/linux-runner x86_64 target/x86_64-unknown-linux-gnu/debug/issue1016`
Hello, world!
```

tested in ci in https://github.com/cross-rs/cross/runs/8285127194?check_suite_focus=true

Alexhuszagh:
This fixes copying files into a directory inside a docker data volume, where previously we copied the entire temp directory into the volume itself. This meant that `tmp.tmpLkX8M3` would be copied to `"${volume}/tmp.tmpLkX8M3"`, rather than its contents being copied into `"${volume}"`. This also contains numerous other bugs fixes and changes internally, which have been submitted each as separate commits.

This fixes a regression in #1054 to re-enable fingerprinting to work, since previously the fingerprint file name was based off the container name. Since #1054 changes the container names to rely on the system time, this ensures the fingerprints are specific for only the toolchain and the path of directory.

This also simplifies the internals of the docker module substantially. Most free functions have been modified to become methods of structs to simplify their function signatures. Some structs have also been renamed to make it clearer what their purpose is.

Closes #1016.

### DIFF
--- a/src/bin/commands/containers.rs
+++ b/src/bin/commands/containers.rs
@@ -443,29 +443,10 @@ pub fn create_persistent_volume(
     docker::Container::create(engine.clone(), container.clone())?;
     docker.run_and_get_status(msg_info, false)?;
 
-    docker::remote::copy_volume_container_xargo(
-        engine,
-        &container,
-        &dirs,
-        mount_prefix.as_ref(),
-        msg_info,
-    )?;
-    docker::remote::copy_volume_container_cargo(
-        engine,
-        &container,
-        &dirs,
-        mount_prefix.as_ref(),
-        copy_registry,
-        msg_info,
-    )?;
-    docker::remote::copy_volume_container_rust(
-        engine,
-        &container,
-        &dirs,
-        None,
-        mount_prefix.as_ref(),
-        msg_info,
-    )?;
+    let data_volume = docker::remote::DataVolume::new(engine, &container, &dirs);
+    data_volume.copy_xargo(mount_prefix.as_ref(), msg_info)?;
+    data_volume.copy_cargo(mount_prefix.as_ref(), copy_registry, msg_info)?;
+    data_volume.copy_rust(None, mount_prefix.as_ref(), msg_info)?;
 
     docker::Container::finish_static(is_tty, msg_info);
 
--- a/src/bin/commands/containers.rs
+++ b/src/bin/commands/containers.rs
@@ -320,7 +320,8 @@ fn get_cross_volumes(
     msg_info: &mut MessageInfo,
 ) -> cross::Result<Vec<String>> {
     use cross::docker::VOLUME_PREFIX;
-    let stdout = docker::subcommand(engine, "volume")
+    let stdout = engine
+        .subcommand("volume")
         .arg("list")
         .args(["--format", "{{.Name}}"])
         // handles simple regex: ^ for start of line.
@@ -348,7 +349,7 @@ pub fn remove_all_volumes(
 ) -> cross::Result<()> {
     let volumes = get_cross_volumes(engine, msg_info)?;
 
-    let mut command = docker::subcommand(engine, "volume");
+    let mut command = engine.subcommand("volume");
     command.arg("rm");
     if force {
         command.arg("--force");
@@ -370,7 +371,7 @@ pub fn prune_volumes(
     engine: &docker::Engine,
     msg_info: &mut MessageInfo,
 ) -> cross::Result<()> {
-    let mut command = docker::subcommand(engine, "volume");
+    let mut command = engine.subcommand("volume");
     command.args(["prune", "--force"]);
     if execute {
         command.run(msg_info, false).map_err(Into::into)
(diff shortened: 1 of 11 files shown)


Alexhuszagh:
I still need a few things to do:
- [ ] Structify the `Engine`

src/docker/remote.rs
```diff
@@ -561,7 +564,7 @@ impl QualifiedToolchain {
             .file_name()
             .expect("should be able to get toolchain name")
             .to_utf8()?;
-        let toolchain_hash = path_hash(self.get_sysroot())?;
+        let toolchain_hash = path_hash(self.get_sysroot(), 5)?;
```
Emilgardis, line 570: could we make the path static variables?

src/bin/commands/containers.rs
```diff
@@ -441,7 +441,7 @@ pub fn create_persistent_volume(
     }
     // store first, since failing to non-existing container is fine
     docker::ChildContainer::create(engine.clone(), container_id.clone())?;
-    docker.run_and_get_status(msg_info, false)?;
+    docker.run_and_get_status(msg_info, true)?;
```
Emilgardis, line 445: nice, I've been annoyed by this :3

src/docker/remote.rs
```diff
@@ -92,7 +92,9 @@ impl<'a, 'b, 'c> ContainerDataVolume<'a, 'b, 'c> {
             file::create_dir_all(dst_path.parent().expect("must have parent"))?;
             fs::copy(&src_path, &dst_path)?;
         }
-        self.copy_files(temppath, dst, msg_info)
+        // if copying from a src directory to dst directory with docker, to
+        // copy the contents from `src` into `dst`, `src` must end with `/.`
+        self.copy_files(&temppath.join("."), dst, msg_info)
```
Emilgardis, line 97: should this maybe be checked and done in `copy_files` ?

i.e the functions does it automagically for you

Emilgardis (changes requested):
great!

src/docker/remote.rs
Alexhuszagh: Sure. I was going to add a comment on the PR that I made the existing ones all 5, but for the mount identifier, I made it require 10 hexdigits. 

For the old hashes, since they either are for the toolchain path (very few actual paths, especially with rustup), a unique ID for the container (which is unique because of the system time, so a small hash is fine), the package path (which is also prefixed by the package name, so small number of possibilities in real cases), 16^5 or ~1million is plenty unique. 

But we run into a birthday problem pretty fast (we're using SHA1 so it's all trusted data) if we consider a user mounting many directories or a large number of users mounting a non-trivial number of directories each. If we have 16^5 options, asuming everything is perfectly distributed, that means we have a 50% chance of collision at ~1200 directories. If we up that to 10 hexdigits, then we have no issue.

We shouldn't modify the toolchain ID, since it would mean existing data volumes would no longer be used, and there's no reason to modify them. But for arbitrary volume mounts, it's probably a good idea to ensure there's no accidental collision.

src/docker/remote.rs
Alexhuszagh: I'll have to check: it's quite possible `copy_cargo` or `copy_rust` relies on this, but it would be trivial to change it to copy-in rather than just copy the entire directory. But this might also need patching elsewhere (looking at `copy_files_nocache`).

Alexhuszagh: So we can't do this entirely (see `rust_base`, where we want to copy into the folder), but we can do it for `copy_files_nocache`, so everything for the project files should be quite doable.

```rust
        let dst = mount_prefix.join(&dirs.sysroot_mount_path_relative()?);
        let rustlib = Path::new("lib").join("rustlib");
        self.create_dir(&dst.join(&rustlib), msg_info)?;
        for basename in ["bin", "libexec", "etc"] {
            let file = dirs.get_sysroot().join(basename);
            self.copy_files(&file, &dst, msg_info)?;
        }
```

See here, we want to copy `bin`, `libexec`, `etc` into `rustlib`, so to `rustlib/bin`, etc.

src/docker/remote.rs
Alexhuszagh: I've added the following global const variables, so it's clear why we're using each in a given scenario.

```rust
/// Short hash for identifiers with minimal risk of collision.
pub const PATH_HASH_SHORT: usize = 5;

/// Longer hash to minimize risk of random collisions
/// Collision chance is ~10^-6
pub const PATH_HASH_UNIQUE: usize = 10;
```

Emilgardis (approved):
lgtm!

bors r+

PR #1073, opened 2022-10-11
Passthrough cross environment variables by default.

Alexhuszagh:


Alexhuszagh:

This passes through any cross configuration environment variables to the container, allowing custom configurations in the future to be forwarded to the container. An initial example is https://github.com/cross-rs/cross-toolchains/pull/15, however, this also applies to `CROSS_DEBUG` and allows us to deny known environment variables cross uses (such as `CROSS_RUNNER` and `CROSS_RUSTC_MINOR_VERSION`).

.changes/1073.json
```diff
@@ -0,0 +1,4 @@
+{
+    "type": "added",
+    "description": "allow passthrough of cross environment variables"
```
Emilgardis, line 3: this makes it seems as if this was somehow not possible before

Emilgardis (approved):
Lgtm!

Alexhuszagh:
bors r=Emilgardis

PR #1078, opened 2022-10-14
Fix custom image names for images with a trailing '-' character.

### LINKED ISSUE #1077: Custom docker images: invalid tag format
### Checklist

### Describe your issue

When cross-compiling a crate from a folder named `foo-123`, Podman will complain with an error:

`Error: tag localhost/cross-rs/cross-custom-foo-:armv7-unknown-linux-gnueabihf-3bda0: invalid reference format`

A quick look into the code shows that there are actually two bugs:

1. Digits are not accepted as valid characters, although they should be (missing match arm [here](https://github.com/cross-rs/cross/blob/dd760cfc548512cb6c22c1e9fe0952c7bfbcf7a9/src/docker/custom.rs#L261))
2. The mangled name needs to be checked that it doesn't end on a dash, as that's considered invalid. This could be avoided by appending one arbitrary valid character to the tag name in those cases.

_No response_

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [X] podman

### cross version

cross 0.2.4

### Example

_No response_

_No response_

Alexhuszagh:
Closes #1077.

### DIFF
--- a/.changes/1078.json
+++ b/.changes/1078.json
@@ -1,0 +1,5 @@
+{
+    "type": "fixed",
+    "description": "Fix custom image names for images with a trailing '-' character.\ncustom images in packages with the `^[^0-9-].*([^A-Za-z_]*-)|(-[^A-Za-z_]*)$` package name pattern are now supported.",
+    "issues": [1077]
+}
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -259,7 +259,7 @@ fn docker_tag_name(file_name: &str) -> String {
     let mut consecutive_underscores = 0;
     for c in file_name.chars() {
         match c {
-            'a'..='z' | '.' | '-' => {
+            'a'..='z' | '0'..='9' | '.' | '-' => {
                 consecutive_underscores = 0;
                 result.push(c);
             }
@@ -278,7 +278,11 @@ fn docker_tag_name(file_name: &str) -> String {
         }
     }
 
-    // in case all characters were invalid, use a non-empty filename
+    // in case all characters were invalid or we had all non-ASCII
+    // characters followed by a `-`, we use a non-empty filename
+    if result.ends_with('-') {
+        result.pop();
+    }
     if result.is_empty() {
         result = "empty".to_owned();
     }
@@ -312,5 +316,8 @@ mod tests {
             docker_tag_name("pAcKaGe---test.name"),
             s!("package---test.name")
         );
+
+        assert_eq!(docker_tag_name("foo-123"), s!("foo-123"));
+        assert_eq!(docker_tag_name("foo-123-"), s!("foo-123"));
     }
 }


src/docker/custom.rs
```diff
@@ -281,6 +281,8 @@ fn docker_tag_name(file_name: &str) -> String {
     // in case all characters were invalid, use a non-empty filename
     if result.is_empty() {
         result = "empty".to_owned();
+    } else if result.ends_with('-') {
```
Alexhuszagh, line 284: `-` isn't a valid Rust package name, so this will never be empty.

.changes/1078.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "fixed",
+    "description": "support custom images in packages with the `.*\\d*-$` package name pattern.",
```
Emilgardis, line 3: this is a bit more general than I think is needed, maybe

```suggestion
    "description": "Fix custom image names for images with a trailing '-' character.\ncustom images in packages with the `.*\\d*-$` package name pattern are now supported.",

**Review by Emilgardis — APPROVED** [review]
lgtm with changelog adjusted a bit

File: .changes/1078.json (hunk 2 above)
Comment (Emilgardis) [reply .changes/1078.json:3]: also maybe mention that digits are now taken into account too

Comment (Alexhuszagh) [reply .changes/1078.json]: Technically digits were accepted before, they were just ignored. If they were used internally, such as `foo123bar`, it would produce a custom image tag of `foobar`, and at the end for `foo123`, it would be `foo`. So the custom image name has changed, but only packages with the `.*\\d*-$` name pattern were affected.

Comment (Alexhuszagh) [reply .changes/1078.json]: The correct regex would for what's now accepted but what would previously be considered invalid is `^.*([^A-Za-z_]*-)|(-[^A-Za-z_]*)$`. I can update the changelog entry with that.

Comment (Alexhuszagh) [reply .changes/1078.json]: Is the changelog entry better? Feel free to merge when ready.

File: src/docker/custom.rs [hunk 3]
```diff
@@ -278,7 +278,11 @@ fn docker_tag_name(file_name: &str) -> String {
         }
     }
 
-    // in case all characters were invalid, use a non-empty filename
+    // in case all characters were invalid or we had all non-ASCII
+    // characters followed by a `-`, we use a non-empty filename
+    if result.ends_with('-') {
```
Alexhuszagh, line 283: Actually this can now be empty, if we have something like `한글-`, which is available on nightly.

Emilgardis: maybe converting to ascii via `let file_name = file_name.as_bytes().escape_ascii()` would work. Making it into a punycode-ish non-unique string

Alexhuszagh: I think it's fine already since the custom image name depends on the target and the path hash. The name is just to make it user-friendly: if we've got Unicode characters, punycode-ish isn't very [friendly](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=11679b7219273da5d8000a9eb99427f5).

```rust
pub fn main() {    
    let x = "cargo한글-";
    println!("escape_ascii={:?}", x.as_bytes().escape_ascii().to_string());
}
```

```bash
$ rustc main.rs
$ ./main
escape_ascii="cargo\\xed\\x95\\x9c\\xea\\xb8\\x80-"
```

After creating the tag name, we'd get something like `cargoxedx95x9cxeaxb8x80`, which I don't think is very helpful. The Docker image would already be named `localhost/cross-rs/cross-custom-empty:armv7-unknown-linux-gnueabihf-ec99f-pre-build`, where `ec99f` is the path hash. If we were worried about collisions (~1/1000)`, extending the path hash to 10 characters might be better.

Emilgardis (approved):
bors r+

Alexhuszagh:
bors retry

PR #1080, opened 2022-10-18
Add `CROSS_SYSROOT` envvar to specify target sysroot.

### LINKED ISSUE #1079: Add `CROSS_SYSROOT` for our Cross-Toolchains
### Checklist

### Describe your request

When copying over or installing external dependencies in custom docker images, adding them to the sysroot so they can be part of the compiler include path or linker search path is non-trivial. Having it so they can use `COPY ext/lib $CROSS_SYSROOT/lib` would make this much easier, or `PREFIX=$CROSS_SYSROOTmake install`, etc.

For certain targets, we already provide information on the sysroot either via bindgen [args](https://github.com/cross-rs/cross/blob/79eeb91c8031b0a02448a7c73fd07ef57b3e2ea8/docker/Dockerfile.aarch64-unknown-linux-gnu#L37) or via [CROSS_MUSL_SYSROOT](https://github.com/cross-rs/cross/blob/79eeb91c8031b0a02448a7c73fd07ef57b3e2ea8/docker/Dockerfile.aarch64-unknown-linux-musl#L19). However, this isn't very user-friendly, isn't meant to be relied upon, nor is it guaranteed to be stable. A better approach: just use `CROSS_SYSROOT` for all targets that support it, and then provide our bindgen and other arguments via that variable.

This can be particularly useful like `arm-unknown-linux-gnueabihf`, where the sysroot is at `/x-tools/arm-unknown-linux-gnueabihf/arm-unknown-linux-gnueabihf/sysroot/`, which is not exactly obvious.

This would make using cross with external dependencies that cannot be installed via a package manager much more user friendly, and requires minimal work on our end.

Alexhuszagh:
Closes #1079.


Alexhuszagh:
Does this need a changelog entry? I don't feel it merits it.

docker/Dockerfile.mips64el-unknown-linux-muslabi64
```diff
@@ -19,13 +19,13 @@ RUN /musl.sh \
     TARGET=mips64el-linux-musl \
     "COMMON_CONFIG += -with-arch=mips64r2"
 
-ENV CROSS_MUSL_SYSROOT=/usr/local/mips64el-linux-musl
+ENV CROSS_SYSROOT=/usr/local/mips64el-linux-musl
```
Emilgardis, line 22: this would be a breaking change

I think having both is fine

Emilgardis (approved):
only one concern about breaking musl, otherwise this is really nice

docker/Dockerfile.mips64el-unknown-linux-muslabi64
Alexhuszagh: It might be, but we never documented and just used it internally, so I'm not sure if it actually affects anyone. `CROSS_MUSL_SYSROOT` was done in #905, which didn't document it, just made our own logic simpler.

Alexhuszagh:
bors r=Emilgardis

docker/Dockerfile.arm-unknown-linux-musleabihf
```diff
@@ -21,16 +21,16 @@ RUN /musl.sh \
...
 COPY qemu-runner base-runner.sh /
 
 ENV CARGO_TARGET_ARM_UNKNOWN_LINUX_MUSLEABIHF_LINKER=arm-linux-musleabihf-gcc \
     CARGO_TARGET_ARM_UNKNOWN_LINUX_MUSLEABIHF_RUNNER="/qemu-runner armhf" \
     CC_arm_unknown_linux_musleabihf=arm-linux-musleabihf-gcc \
     CXX_arm_unknown_linux_musleabihf=arm-linux-musleabihf-g++ \
-    BINDGEN_EXTRA_CLANG_ARGS_arm_unknown_linux_musleabihf="--sysroot=$CROSS_MUSL_SYSROOT" \
-    QEMU_LD_PREFIX=$CROSS_MUSL_SYSROOT \
+    BINDGEN_EXTRA_CLANG_ARGS_arm_unknown_linux_musleabihf="--sysroot=$CROSS_SYSROOT" \
+    QEMU_LD_PREFIX="$CROSS_SYSROO"T \
```
Alexhuszagh, line 35: Ah...

Alexhuszagh:
bors r=Emilgardis

PR #1083, opened 2022-10-18
Avoid installing toolchain if is_custom is true

otavio:
Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -578,15 +578,14 @@ To override the toolchain mounted in the image, set `target.{}.image.toolchain =
             // set the sysroot explicitly to the toolchain
             let mut is_nightly = toolchain.channel.contains("nightly");
 
-            let installed_toolchains = rustup::installed_toolchains(msg_info)?;
-
-            if !installed_toolchains
-                .into_iter()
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -578,15 +578,14 @@ To override the toolchain mounted in the image, set `target.{}.image.toolchain =
             // set the sysroot explicitly to the toolchain
             let mut is_nightly = toolchain.channel.contains("nightly");
 
-            let installed_toolchains = rustup::installed_toolchains(msg_info)?;
```
Alexhuszagh, line 581: This is removed entirely, which is needed further below.

Alexhuszagh (changes requested):
I'm not sure if this is enough as well: I'm testing with a system Rust install on Fedora 36 (it will only be able to compile for `x86_64-unknown-linux-gnu`), and we currently have issues also with identifying the proper toolchain path even if rustup is disabled.

src/lib.rs
otavio: I reworked the code; it is not used now.

PR #1085, opened 2022-10-19
Fixes custom toolchain support without rustup.

Alexhuszagh:
Avoids checking for installed toolchains, and also prevents renaming the sysroot based on the toolchain full name, since this may be in a directory completely different than the target name.

### DIFF
--- a/src/rustc.rs
+++ b/src/rustc.rs
@@ -92,12 +92,18 @@ pub struct QualifiedToolchain {
 }
 
 impl QualifiedToolchain {
-    pub fn new(channel: &str, date: &Option<String>, host: &ImagePlatform, sysroot: &Path) -> Self {
+    pub fn new(
+        channel: &str,
+        date: &Option<String>,
+        host: &ImagePlatform,
+        sysroot: &Path,
+        is_custom: bool,
+    ) -> Self {
         let mut this = Self {
             channel: channel.to_owned(),
             date: date.clone(),
             host: host.clone(),
-            is_custom: false,
+            is_custom,
             full: if let Some(date) = date {
                 format!("{}-{}-{}", channel, date, host.target)
             } else {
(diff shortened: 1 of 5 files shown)


Alexhuszagh:
@otavio Can you test if this works for you? It only checks if the toolchain is installed if not using a custom toolchain, but also avoids naming the sysroot in that case. This means that the above example where we had a sysroot of `/usr` being renamed to `/stable-x86_64-unknown-linux-gnu` won't happen with custom toolchains.

This should supersede #1083.

otavio (approved):

otavio:
It worked just fine.

Emilgardis:
bors try --none

just want some artifacts :)

Emilgardis:
bors try-

bors try --target x86_64-unknown-linux-gnu x86_64-pc-windows-msvc

Emilgardis:
I think we can improve this story a bit.

When using a non-rustup toolchain installed by e.g a package manager, you'll get the following error message


```
emil@debian-gnu-linux-10:/home/emil/dev_space/aaaaaa-2018$ CROSS_CUSTOM=1 cross build
Error: 
   0: toolchain is not fully qualified

Location:
   src/rustc.rs:268

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

we could add a `with_suggestion` here, that suggestion could link to a wiki entry, or just say `set CROSS_CUSTOM_TOOLCHAIN=1 or install rust via rustup`

Emilgardis:
We already have a similar message here, but that only works if the sysroot looks like a rustc triple

https://github.com/Emilgardis/cross/blob/2a210f69e6f43ec781c96e6e1a0d3ed18c643d96/src/rustup.rs#L97

Alexhuszagh:
I've changed the error message to be:

```rust
Ok(_) => return Err(eyre::eyre!("toolchain is not fully qualified")
    .with_suggestion(|| {
        "if using a custom toolchain try set `CROSS_CUSTOM_TOOLCHAIN=1` or install rust via rustup"
})),
```

src/rustc.rs
```diff
@@ -256,7 +265,10 @@ impl QualifiedToolchain {
             Ok(_) | Err(_) if config.custom_toolchain() => {
                 QualifiedToolchain::custom(toolchain, &sysroot, config, msg_info)
             }
-            Ok(_) => eyre::bail!("toolchain is not fully qualified"),
+            Ok(_) => return Err(eyre::eyre!("toolchain is not fully qualified")
+                .with_suggestion(|| {
+                    "if using a custom toolchain try set `CROSS_CUSTOM_TOOLCHAIN=1` or install rust via rustup"
+            })),
```
Emilgardis, line 271:
```suggestion
            Ok(_) => return Err(eyre::eyre!("toolchain is not fully qualified")
                .with_note(|| "cross expects the toolchain to be a rustup installed toolchain")
                .with_suggestion(|| {
                    "if you're using a custom toolchain try setting `CROSS_CUSTOM_TOOLCHAIN=1` or install rust via rustup"
            })),
```

Emilgardis (approved):
bors r+

nice :D

PR #1087, opened 2022-10-19
Support crosstool for uclibc, musl, and newlib.

Alexhuszagh:


xtask/src/crosstool.rs
```diff
@@ -124,84 +303,105 @@ CT_GCC_VERSION=\"{gcc_major}.{gcc_minor}.{gcc_patch}\""
...
+        .replacen(ct_libc_v_placeholder, &ct_libc_v, 1)
+        .replacen(ct_libc_placeholder, &ct_libc, 1);
 
     // configure the `CT_LINUX` values
-    let linux_major = linux_versions[0].parse::<u32>()?;
-    let linux_minor = linux_versions[1].parse::<u32>()?;
-    let linux_patch = linux_versions.get(2).unwrap_or(&"0").parse::<u32>()?;
-    let ct_linux_v = format!(
-        "CT_LINUX_V_{linux_major}_{linux_minor}=y
+    if file_name.contains("linux") {
```
Alexhuszagh, line 353: Since we now support newlib targets, Linux needs to be optional.

Emilgardis (approved):
minor nit, otherwise looks good

Alexhuszagh:
> minor nit

What's the nit?

Alexhuszagh:
bors r-
This actually needs a mirror config option for pre-1.0 uclibc, which is required for some embedded systems.

Alexhuszagh:
Pre-1.0 uclibc support, at least from the configuration standpoint has been fixed. I'm diagnosing some other errors, but those are limited to compiler and Linux versions, and not the build process itself.

Emilgardis (approved):
bors r+

PR #1090, opened 2022-10-21
Add editorconfig file.

### LINKED ISSUE #1089: add a `.editorconfig`
maybe we should add a [`.editorconfig`](https://editorconfig.org/) too while I'm at it and lint on it? Unsure how to lint a `.editorconfig` in a simple way though

_Originally posted by @Emilgardis in https://github.com/cross-rs/cross/pull/1015#discussion_r967640904_

Alexhuszagh:
Closes #1089.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{rs,py,sh,md,toml,yml,js}]
+charset = utf-8
+
+[*.{rs,py,sh,yml,js}]
+trim_trailing_whitespace = true
+
+[*.{rs,py,sh}]
+indent_style = space
+indent_size = 4
+
+[*.yml]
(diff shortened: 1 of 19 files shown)


Alexhuszagh:
The code can be linted via [eclint](https://www.npmjs.com/package/eclint), however, this produces a lot of false positives, especially in inline diff files, which means linting in CI is a waste of time. These are suggestions: not hard rules.

Emilgardis (approved):
awesome!

bors r+

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
I think some of these might be now an unrelated issue and due to the failing being the same one in the non-cached build: `max_kernel_version` isn't picking up anything because the glob expansion doesn't work:
https://github.com/cross-rs/cross/actions/runs/3293850267

Alexhuszagh:
bors r=Emilgardis

PR #1021, opened 2022-09-12
Allow CI without requiring targets to be built/tested.

Alexhuszagh:
Co-authored-by: Emil Gardström <emil.gardstrom@gmail.com>

PR #1091, opened 2022-10-22
improve weekly checks and push the resulting images

Emilgardis:
since the maximum timeout for workflows is not enough, we need to use the same strategy as our normal ci flow

also adds the ability to filter targets to run

image

this pr also pushes the results so that the cache is renewed, making the images not stagnate

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -45,6 +45,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
       - name: Build xtask
         run: cargo build -p xtask
+
       - name: Prepare Meta
         id: prepare-meta
         timeout-minutes: 60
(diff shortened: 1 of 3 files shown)


Emilgardis:
ideally, we could use a

`uses: .github/workflows/ci.yml`

and use inputs to specify not using cache, etc etc, but I'm not entirely sure how to do that

Emilgardis:
seems to work
https://github.com/Emilgardis/cross/actions/runs/3302954301/jobs/5450406715

Emilgardis:
this would conflict with #1021

I think we can merge that one

xtask/src/ci/target_matrix.rs
```diff
@@ -11,6 +11,24 @@ pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Rep
     let mut matrix: Vec<CiTarget> = get_matrix().clone();
     let (prs, mut app) = if author == "bors[bot]" {
         process_bors_message(&message)?
+    } else if author == "[gha]" {
```
Emilgardis, line 14: im not sure this is the best approach

Alexhuszagh: Why not just use an environment variable here? We don't need to hack the author if we set an envvar upon this workflow, or a CLI argument. Something like:

```rust
    TargetMatrix {
        #[clap(long, env = "COMMIT_MESSAGE")]
        message: String,
        #[clap(long, env = "COMMIT_AUTHOR")]
        author: String,
        // if the check is being run as part of a weekly check
        #[clap(long, env = "WEEKLY_CI", value_parser = BoolishValueParser::new())]
        weekly: Option<bool>,
    },
```

xtask/src/ci/target_matrix.rs
```diff
@@ -158,6 +178,9 @@ impl TargetMatrixArgs {
             std::mem::take(matrix);
             return;
         }
+        if self.has_image {
```
Alexhuszagh, line 181: I don't believe this is used, ever, since it's never called from either `weekly.yml` or `ci.yml`.

.github/workflows/weekly.yml
```diff
@@ -10,21 +14,53 @@ env:
...
+    runs-on: ${{ matrix.os }}
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-rust
       - name: Set up Docker Buildx
```
Alexhuszagh, line 43: We currently only use Linux but this should probably use `if: runner.os == 'Linux'` since we aren't hard-coding `runs-on: ubuntu-latest` anymore.

Alexhuszagh (changes requested):
This is mostly good, however, I think we can fix a few things:

1. Avoid hijacking the author by using a `--weekly` flag to ` cargo xtask ci-job target-matrix`
2. Ensure we only use the `buildx` action on Linux, since we no longer hard-code the matrix OS.

The diff I have with my [branch](https://github.com/Alexhuszagh/cross/tree/better-weekly) is:

```diff
diff --git a/.github/workflows/weekly.yml b/.github/workflows/weekly.yml
index 5bf7eb3..3038927 100644
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Generate matrix
         id: generate-matrix
         run: |
-          cargo xtask ci-job target-matrix --author "[gha]" --message "check"
+          cargo xtask ci-job target-matrix --author "[gha]" --message "check" --weekly
         env:
           TARGETS: ${{ inputs.targets }}
   weekly:
@@ -41,6 +41,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-rust
       - name: Set up Docker Buildx
+        if: runner.os == 'Linux'
         uses: docker/setup-buildx-action@v1
       - name: Build xtask
         run: cargo build -p xtask
diff --git a/xtask/src/ci.rs b/xtask/src/ci.rs
index 012be66..f4b4d6b 100644
--- a/xtask/src/ci.rs
+++ b/xtask/src/ci.rs
@@ -31,6 +31,9 @@ pub enum CiJob {
         message: String,
         #[clap(long, env = "COMMIT_AUTHOR")]
         author: String,
+        // if the check is being run as part of a weekly check
+        #[clap(long, env = "WEEKLY_CI")]
+        weekly: bool,
     },
 }
 
@@ -114,8 +117,12 @@ pub fn ci(args: CiJob, metadata: CargoMetadata) -> cross::Result<()> {
                 }
             }
         }
-        CiJob::TargetMatrix { message, author } => {
-            target_matrix::run(message, author)?;
+        CiJob::TargetMatrix {
+            message,
+            author,
+            weekly,
+        } => {
+            target_matrix::run(message, author, weekly)?;
         }
     }
     Ok(())
diff --git a/xtask/src/ci/target_matrix.rs b/xtask/src/ci/target_matrix.rs
index a84db0b..42ab40d 100644
--- a/xtask/src/ci/target_matrix.rs
+++ b/xtask/src/ci/target_matrix.rs
@@ -7,11 +7,11 @@ use serde::{Deserialize, Serialize};
 
 use crate::util::{get_matrix, gha_output, gha_print, CiTarget, ImageTarget};
 
-pub(crate) fn run(message: String, author: String) -> Result<(), color_eyre::Report> {
+pub(crate) fn run(message: String, author: String, weekly: bool) -> Result<(), color_eyre::Report> {
     let mut matrix: Vec<CiTarget> = get_matrix().clone();
     let (prs, mut app) = if author == "bors[bot]" {
         process_bors_message(&message)?
-    } else if author == "[gha]" {
+    } else if weekly {
         let app = TargetMatrixArgs {
             target: std::env::var("TARGETS")
                 .unwrap_or_default()

```

xtask/src/ci/target_matrix.rs
Emilgardis: it is, via https://github.com/Emilgardis/cross/blob/0a4ff656c7881425aef62552d9e4ed13efd49d2e/xtask/src/ci/target_matrix.rs#L29

Alexhuszagh (approved):
lgtm, just fix the trivial merge conflict and merge when ready

Alexhuszagh:
Squash the commit (I just trivially fixed the merge conflict) and this is ready.

Emilgardis:
I think I want to keep the two commits, they are a bit distinct

Alexhuszagh:
Ok merge whenever ready.

Emilgardis:
bors r=Alexhuszagh

Emilgardis:
bors retry

Emilgardis:
bors cancel

bors r=Alexhuszagh

PR #1093, opened 2022-10-24
Add note for custom images and arm-unknown-linux-gnueabihf.

Alexhuszagh:


Alexhuszagh:
This produces the below warning message. This is needed because this has been repeatedly been an issue with people attempting to install such packages, and confused when we added them to a deny list by pinning all system installation candidates to a negative value (blocked).

Screenshot_20221024_115911

src/docker/shared.rs
```diff
@@ -86,6 +86,9 @@ impl DockerOptions {
         msg_info: &mut MessageInfo,
     ) -> Result<String> {
         let mut image = self.image.clone();
+        if self.target.triple() == "arm-unknown-linux-gnueabihf" {
+            msg_info.warn("cannot install armhf system packages via apt for `arm-unknown-linux-gnueabihf`, since they are for ARMv7a targets but this target is ARMv6. installation of all packages for the armhf architecture has been blocked.")?;
```
Emilgardis, line 90: I don't think this should be a "warning", instead maybe a "note" or "info" ?

Emilgardis (changes requested):
I think

Emilgardis (approved):
bors r+

PR #1105, opened 2022-11-01
Explicitly prefer `-ar` to `-gcc-ar`.

### LINKED ISSUE #1100: archiver in freebsd images appears to be broken
### Checklist

### Describe your issue

The toolchain used in the freebsd images appears to be broken. I managed to reduce the test case basically down to just:

```shell
$ docker pull ghcr.io/cross-rs/x86_64-unknown-freebsd:0.2.4
$ docker run -it ghcr.io/cross-rs/x86_64-unknown-freebsd:0.2.4 bash
$ echo "int banana(void) { return 0; }" | x86_64-unknown-freebsd12-gcc -xc - -c -o chaenomeles.o
$ x86_64-unknown-freebsd12-gcc-ar cq mango.ar chaenomeles.o; echo $?
x86_64-unknown-freebsd12-gcc-ar: Cannot find plugin 'liblto_plugin.so'
1
```

This in practice causes crates such as `cc` fail to package the built objects into an `.a` in crates’ build.rs scripts.

i686-unknown-freebsd, x86_64-unknown-freebsd

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [X] docker
- [ ] podman

### cross version

cross 0.2.4

### Example

_No response_

The closest relevant thing I can find on the internet is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81326. Curiously we are already building the toolchain with `--disable-lto` flag, so it might be a bug in gcc itself?

Alexhuszagh:
Fixes #1100.

### DIFF
--- a/.changes/1105.json
+++ b/.changes/1105.json
@@ -1,0 +1,5 @@
+{
+    "type": "changed",
+    "description": "explicitly prefer `-ar` to `-gcc-ar`.",
+    "issues": [1100]
+}
--- a/docker/Dockerfile.i686-unknown-freebsd
+++ b/docker/Dockerfile.i686-unknown-freebsd
@@ -23,6 +23,7 @@ COPY i686-unknown-freebsd12-gcc.sh /usr/bin
 
 ENV CROSS_SYSROOT=/usr/local/i686-unknown-freebsd12
 ENV CARGO_TARGET_I686_UNKNOWN_FREEBSD_LINKER=i686-unknown-freebsd12-gcc.sh \
+    AR_i686_unknown_freebsd=i686-unknown-freebsd12-ar \
     CC_i686_unknown_freebsd=i686-unknown-freebsd12-gcc \
     CXX_i686_unknown_freebsd=i686-unknown-freebsd12-g++ \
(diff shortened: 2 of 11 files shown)


Alexhuszagh:
bors try --target *-none-* --target *-freebsd --target *-dragonfly --target *-solaris --target *-illumos *-netbsd

Alexhuszagh:
Note I didn't provide `AR_$target` for the Android images, since these are provided in #1023.

Emilgardis (approved):
this change is parallel to https://github.com/rust-lang/cc-rs/pull/741, but this is fine standalone and resolves issues for people locked to `cc="=1.0.74"`

bors r+

svenstaro:
Can we get a release containing this? Otherwise this is currently broken by default.

Alexhuszagh:
I don't think 0.3.0 is quite ready for release, especially not in the next week or so, but this is a trivial patch so we could backport it and do a quick release for 0.2.5?

Emilgardis:
cc 1.0.75 has been released. https://github.com/rust-lang/cc-rs/pull/741#issuecomment-1307514624

hopefully this solves the problem enough

Issue #659, opened 2022-03-17
error/panic on fallback to host `cargo` or when in CI environment


Emilgardis:
https://github.com/cross-rs/cross/blob/eaeff5b3fed37c4762a82b27870c1543acdab72b/src/main.rs#L344-L350

jpmckinney:
I would also like an option to error/panic on fallback. This is especially useful in CI. I guess for now I can do:

```
      - id: cross
        run: echo "stderr=$(cross build --release --target ${{ matrix.target }})" 2>> $GITHUB_OUTPUT
      - if: steps.cross.outputs.stderr != ''
        run: |
          echo "${{ steps.cross.outputs.stderr }}"
          exit 1
```

Emilgardis:
thats something I hadn't considered, maybe providing a way to not allow any warnings would be helpful.

We definitely want this included in the next release

PR #1108, opened 2022-11-02
Add progress config options to custom image builds.

### LINKED ISSUE #1102: Add Progress Config Options for Custom Image Builds
Currently, when building custom images, the container engine always uses the `auto` progress for the build, meaning that debugging any builds is difficult from a terminal. It would be handy to allow environment variables to configure these options.

Alexhuszagh:
Closes #1102.

### DIFF
--- a/src/docker/build.rs
+++ b/src/docker/build.rs
@@ -1,0 +1,121 @@
+use std::env;
+use std::process::Command;
+use std::str::FromStr;
+
+use super::engine::Engine;
+use crate::errors::*;
+use crate::shell::Verbosity;
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Progress {
+    Plain,
+    Auto,
+    Tty,
+}
+
+impl FromStr for Progress {
+    type Err = eyre::ErrReport;
+
+    fn from_str(progress: &str) -> Result<Self> {
+        Ok(match progress {
+            "plain" => Progress::Plain,
+            "auto" => Progress::Auto,
+            "tty" => Progress::Tty,
+            s => eyre::bail!("unexpect progress type: expected plain, auto, or tty and got {s}"),
+        })
+    }
+}
+
+impl From<Progress> for &str {
+    fn from(progress: Progress) -> Self {
+        match progress {
+            Progress::Plain => "plain",
+            Progress::Auto => "auto",
+            Progress::Tty => "tty",
+        }
+    }
+}
+
+pub trait BuildCommandExt {
+    fn invoke_build_command(&mut self) -> &mut Self;
+    fn progress(&mut self, progress: Option<Progress>) -> Result<&mut Self>;
+    fn verbose(&mut self, verbosity: Verbosity) -> &mut Self;
+    fn disable_scan_suggest(&mut self) -> &mut Self;
+    fn cross_labels(&mut self, target: &str, platform: &str) -> &mut Self;
+}
+
+impl BuildCommandExt for Command {
+    fn invoke_build_command(&mut self) -> &mut Self {
+        match Engine::has_buildkit() {
+            true => self.args(["buildx", "build"]),
+            false => self.arg("build"),
+        }
+    }
+
+    fn progress(&mut self, progress: Option<Progress>) -> Result<&mut Self> {
+        let progress: Progress = match progress {
+            None => env::var("CROSS_BUILD_PROGRESS")
+                .as_deref()
+                .unwrap_or("auto")
+                .parse()?,
+            Some(progress) => progress,
+        };
+        Ok(self.args(["--progress", progress.into()]))
+    }
+
+    fn verbose(&mut self, verbosity: Verbosity) -> &mut Self {
+        match verbosity {
+            Verbosity::Verbose(2..) => self.args(["--build-arg", "VERBOSE=1"]),
+            _ => self,
+        }
+    }
+
+    fn disable_scan_suggest(&mut self) -> &mut Self {
+        self.env("DOCKER_SCAN_SUGGEST", "false")
+    }
+
+    fn cross_labels(&mut self, target: &str, platform: &str) -> &mut Self {
+        self.args([
+            "--label",
+            &format!("{}.for-cross-target={target}", crate::CROSS_LABEL_DOMAIN,),
+        ]);
+        self.args([
+            "--label",
+            &format!("{}.runs-with={platform}", crate::CROSS_LABEL_DOMAIN,),
+        ])
+    }
+}
+
+pub trait BuildResultExt {
+    fn engine_warning(self, engine: &Engine) -> Result<()>;
+    fn buildkit_warning(self) -> Result<()>;
+}
+
+impl BuildResultExt for Result<()> {
+    fn engine_warning(self, engine: &Engine) -> Result<()> {
+        self.with_warning(|| {
+            format!(
+                "call to {} failed",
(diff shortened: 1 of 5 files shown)


src/docker/build.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn build(&mut self) -> &mut Self {
+        match Engine::has_buildkit() {
+            true => self.args(["buildx", "build"]),
+            false => self.arg("build"),
+        }
+    }
+
+    fn progress(&mut self, progress: Option<&str>) -> Result<&mut Self> {
+        let progress: Progress = match progress {
+            None => env::var("CROSS_BUILD_PROGRESS")
```
Alexhuszagh, line 57: I don't like parsing from `Progress` and then re-serializing it, but this also needs good error messages with clap. This needs to be changed to use the parser from clap with auto-derive parsers and serializers.

src/docker/build.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+pub trait DockerResultExt {
+    fn engine_warning(self, engine: &Engine) -> Result<()>;
+    fn buildkit_warning(self) -> Result<()>;
+}
+
+impl DockerResultExt for Result<()> {
+    fn engine_warning(self, engine: &Engine) -> Result<()> {
```
Alexhuszagh, line 99: Allow the BuildKit and engine warnings when doing local image builds as well, and not just custom ones. I'm not sure if this would interfere with GHA?

xtask/src/main.rs
```diff
@@ -98,7 +98,7 @@ pub fn main() -> cross::Result<()> {
             target_info::target_info(args, &engine, &mut msg_info)?;
         }
         Commands::BuildDockerImage(args) => {
-            let mut msg_info = get_msg_info!(args, args.verbose != 0)?;
```
Alexhuszagh, line 101: We now want to add the `VERBOSE` flag with `-vv` builds, I believe, so we should re-enable this to accept the levels, unless there's a reason I'm forgetting.

src/docker/build.rs
Alexhuszagh: I think changing it take `Option<Progress>` otherwise parsing the environment variable is the best option, and letting clap use the default parser. Otherwise, we'd need to use `clap` or `serde` logic internally.

src/docker/build.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+pub trait BuildCommandExt {
+    fn build(&mut self) -> &mut Self;
+    fn progress(&mut self, progress: Option<Progress>) -> Result<&mut Self>;
+    fn verbose(&mut self, verbosity: Verbosity) -> &mut Self;
+    fn scan_suggest(&mut self, suggest: bool) -> &mut Self;
+    fn cross_labels(&mut self, target: &str, platform: &str) -> &mut Self;
+}
```
Emilgardis, line 45: I'm not sure I see the benefit of doing this this way.

src/docker/build.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    fn scan_suggest(&mut self, suggest: bool) -> &mut Self {
+        let value = match suggest {
+            true => "true",
+            false => "false",
+        };
+        self.env("DOCKER_SCAN_SUGGEST", value)
+    }
```
Emilgardis, line 79: when would you need this to be false?

Emilgardis (changes requested):

src/docker/build.rs
Alexhuszagh: We use it false already in custom images, so I decided to keep that behavior.

src/docker/build.rs
Alexhuszagh: Have another suggestion for de-duplicating some of the code so we don't redundantly add cross labels and build commands?

src/docker/build.rs
Emilgardis: my bad! I meant when would you want this to be `true`

Alexhuszagh: Ah probably never, we could just have `disable_scan_suggest` I guess?

src/docker/build.rs
Emilgardis: I misunderstood the intent of this. My concern was with making it hard to decouple xtask and custom build in cross, but as it is now it's fine.

xtask/src/main.rs
Emilgardis: I don't think there is a reason, just an artifact when -vv was enabled

xtask/src/build_docker_image.rs
```diff
@@ -249,15 +243,13 @@ pub fn build_docker_image(
         if engine.kind.supports_pull_flag() {
             docker_build.arg("--pull");
         }
+        let untagged_name = format!("{repository}/{}", target.name);
```
Emilgardis, line 246: what does `untagged` mean?

Emilgardis (changes requested):

src/docker/build.rs
Emilgardis: should be fine

Emilgardis: only important to make sure we don't read `CI` or other ci envs when doing custom builds unless the intent is to actually do that, like https://github.com/cross-rs/cross/issues/659

xtask/src/build_docker_image.rs
Emilgardis: I think this could be renamed to just `image_name`

src/docker/build.rs
Emilgardis: sounds good!

xtask/src/build_docker_image.rs
Alexhuszagh: It's for podman: the name without the tag. `ghcr.io/cross-rs/x86_64-unknown-linux-gnu` is the untagged name, and `ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main` is the tagged name. Podman doesn't accept tags for the `--cache-from` (sigh).

Alexhuszagh: > I think this could be renamed to just `image_name`

I intentionally avoided that because `target.image_name` adds the tag to the image, which is then confusing.

Emilgardis: maybe `base_name`?

Alexhuszagh:
bors try --none

src/docker/build.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn cross_labels(&mut self, target: &str, platform: &str) -> &mut Self;
+}
+
+impl BuildCommandExt for Command {
+    fn build(&mut self) -> &mut Self {
+        match Engine::has_buildkit() {
+            true => self.args(["buildx", "build"]),
+            false => self.arg("build"),
+        }
+    }
```
Emilgardis, line 53: I think `BuildCommandExt` should be renamed to make clear what it does, e.g invoke the build subcommand

Alexhuszagh: Such as `DockerBuildCommentExt`?

Emilgardis: ... I can't think straight :)

the comment was for `BuildCommandExt:build` but I left out `::build` ...

Emilgardis: maybe

`invoke_build_command`

Emilgardis (approved):
bors r+

PR #1112, opened 2022-11-04
Fix CMake Support and Enhance Newlib Targets

### LINKED ISSUE #1110: CMake cannot find NDK or toolchain file for Android
### Checklist

### Describe your issue

I want to compile a dependency which is written in C and build it with CMake. The target is Android aarch64.
In order to do that, I specify in the build.rs file the Android (ANDROID_PLATFORM) and ABI (ANDROID_ABI) version, as requested from the official documentation: https://developer.android.com/ndk/guides/cmake#file
The CMAKE_TOOLCHAIN_FILE variable is not needed anymore, as CMake > 3.21 is used.
When I try to compile it with `cross build --target aarch64-linux-android` CMake cannot find the NDK or a standalone toolchain.

> --- stderr
> CMake Error at /usr/local/share/cmake-3.23/Modules/Platform/Android-Determine.cmake:209 (message):
>   Android: Neither the NDK or a standalone toolchain was found.

Looking at the environment variables I see that the needed variables for CMake are not set as requested here: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android

My question is now, what is the best approach in order to built C/C++ dependencies with CMake for Android?

aarch64-linux-android

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [X] docker
- [ ] podman

### cross version

cross 0.2.4 (a8082de 2022-10-26)

### Example

_No response_

_No response_

### LINKED ISSUE #1113: Newlib Targets Incorrectly Report `std` Support in `targets.toml`.
Looking at [Rust Platform Support](https://doc.rust-lang.org/rustc/platform-support.html), we see:

target | std | notes
-- | -- | --
thumbv6m-none-eabi | * | Bare Cortex-M0, M0+, M1
thumbv7em-none-eabi | * | Bare Cortex-M4, M7
thumbv7em-none-eabihf | * | Bare Cortex-M4F, M7F, FPU, hardfloat
thumbv7m-none-eabi | * | Bare Cortex-M3

where:

> \* indicates the target only supports [no_std](https://rust-embedded.github.io/book/intro/no-std.html) development.

However, all of our targets for newlib report `std = true`.

https://github.com/cross-rs/cross/blob/35a1e174aeb7f4a81d26e7c2efba2aee63a42b6f/targets.toml#L440-L458

Alexhuszagh:
Fix CMake Support and Enhance Newlib Targets

Fix CMake support for Android targets by providing a toolchain file that specifies the API level and the standalone toolchain path, and ensuring that compiler detection is forcibly disabled. Fix CMake support for newlib targets by providing a toolchain file that signals to CMake it is compiling for an embedded target, providing correct C, C++, and ASM flags even when invoked externally, and ensuring CMake skips testing if the C and C++ compilers work by running a binary file. Also add C++ support to bare-metal newlib targets, by also installing `libstdc++-arm-none-eabi-newlib`. Finally, `targets.toml` was updated to ensure newlib targets do not have `std` support.

Other minor changes including using the Cortex-M1 CPU for `thumbv6m-none-eabi` rather than the Cortex-M3 (an ARMv7-M CPU) for the Qemu runner.

To ensure these work, test suites for CMake using C+11 features, CMake's `try_run` feature (checking [CMAKE_CROSSCOMPILING_EMULATOR](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING_EMULATOR.html) is set), and a `no_std` package with a C++ dependency have been added.

Closes #1110.
Closes #1113.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -68,12 +68,10 @@ main() {
         retry cargo fetch
         # don't use xargo: should have native support just from rustc
         rustup toolchain add nightly
-        "${CROSS[@]}" build --lib --target "${TARGET}" ${CROSS_FLAGS}
+        cross_build --lib --target "${TARGET}"
         popd
 
         rm -rf "${td}"
-
-        return
     fi
 
     # `cross build` test for the other targets
@@ -83,7 +81,7 @@ main() {
         pushd "${td}"
         cargo init --lib --name foo .
         retry cargo fetch
-        "${CROSS[@]}" build --target "${TARGET}" ${CROSS_FLAGS}
+        cross_build --target "${TARGET}"
         popd
 
         rm -rf "${td}"
@@ -94,7 +92,7 @@ main() {
         # test that linking works
         cargo init --bin --name hello .
         retry cargo fetch
-        "${CROSS[@]}" build --target "${TARGET}" ${CROSS_FLAGS}
+        cross_build --target "${TARGET}"
         popd
 
         rm -rf "${td}"
@@ -166,8 +164,22 @@ main() {
 
     fi
 
-    # Test C++ support
+    # Test C++ support in a no_std context
     if (( ${CPP:-0} )); then
+        td="$(mkcargotemp -d)"
+
+        git clone --depth 1 https://github.com/cross-rs/rust-cpp-accumulate "${td}"
+
+        pushd "${td}"
+        retry cargo fetch
+        cross_build --target "${TARGET}"
+        popd
+
+        rm -rf "${td}"
+    fi
+
+    # Test C++ support
+    if (( ${STD:-0} )) && (( ${CPP:-0} )); then
         td="$(mkcargotemp -d)"
 
         git clone --depth 1 https://github.com/cross-rs/rust-cpp-hello-word "${td}"
@@ -177,7 +189,7 @@ main() {
         if (( ${RUN:-0} )); then
             cross_run --target "${TARGET}"
         else
-            "${CROSS[@]}" build --target "${TARGET}" ${CROSS_FLAGS}
+            cross_build --target "${TARGET}"
         fi
         popd
 
@@ -193,11 +205,44 @@ main() {
         cargo init --bin --name hello .
         retry cargo fetch
         RUSTFLAGS="-C target-feature=-crt-static" \
-            "${CROSS[@]}" build --target "${TARGET}" ${CROSS_FLAGS}
+            cross_build --target "${TARGET}"
         popd
 
         rm -rf "${td}"
     fi
+
+    # test cmake support
+    td="$(mkcargotemp -d)"
+
+    git clone \
+        --recursive \
+        --depth 1 \
+        https://github.com/cross-rs/rust-cmake-hello-world "${td}"
+
+    pushd "${td}"
+    retry cargo fetch
+    if [[ "${TARGET}" == "arm-linux-androideabi" ]]; then
+        # ARMv5te isn't supported anymore by Android, which produces missing
+        # symbol errors with re2 like `__libcpp_signed_lock_free`.
+        cross_run --target "${TARGET}" --features=tryrun
+    elif (( ${STD:-0} )) && (( ${RUN:-0} )) && (( ${CPP:-0} )); then
+        cross_run --target "${TARGET}" --features=re2,tryrun
+    elif (( ${STD:-0} )) && (( ${CPP:-0} )); then
+        cross_build --target "${TARGET}" --features=re2
+    elif (( ${STD:-0} )) && (( ${RUN:-0} )); then
+        cross_run --target "${TARGET}" --features=tryrun
+    elif (( ${STD:-0} )); then
+        cross_build --target "${TARGET}" --features=tryrun
+    else
+        cross_build --lib --target "${TARGET}"
+    fi
+    popd
+
+    rm -rf "${td}"
+}
+
+cross_build() {
+    "${CROSS[@]}" build "$@" ${CROSS_FLAGS}
 }
 
 cross_run() {
--- a/docker/Dockerfile.aarch64-linux-android
+++ b/docker/Dockerfile.aarch64-linux-android
@@ -34,15 +34,18 @@ RUN /android-system.sh arm64
 
 ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-android-
 ENV CROSS_SYSROOT=/android-ndk/sysroot
+ENV CROSS_ANDROID_SDK=$ANDROID_SDK
 COPY android-symlink.sh /
 RUN /android-symlink.sh aarch64 aarch64-linux-android
 
 COPY android-runner /
+COPY android.cmake /opt/toolchain.cmake
 
 # Libz is distributed in the android ndk, but for some unknown reason it is not
 # found in the build process of some crates, so we explicit set the DEP_Z_ROOT
+ENV CROSS_TARGET_RUNNER="/android-runner aarch64"
 ENV CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
-    CARGO_TARGET_AARCH64_LINUX_ANDROID_RUNNER="/android-runner aarch64" \
+    CARGO_TARGET_AARCH64_LINUX_ANDROID_RUNNER="$CROSS_TARGET_RUNNER" \
     AR_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"ar \
     AS_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"as \
     CC_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"gcc \
@@ -56,6 +59,7 @@ ENV CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     SIZE_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"size \
     STRINGS_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"strings \
     STRIP_aarch64_linux_android="$CROSS_TOOLCHAIN_PREFIX"strip \
+    CMAKE_TOOLCHAIN_FILE_aarch64_linux_android=/opt/toolchain.cmake \
     BINDGEN_EXTRA_CLANG_ARGS_aarch64_linux_android="--sysroot=$CROSS_SYSROOT" \
     DEP_Z_INCLUDE="$CROSS_SYSROOT/usr/include"/ \
     RUST_TEST_THREADS=1 \
@@ -63,4 +67,8 @@ ENV CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     TMPDIR=/tmp/ \
     ANDROID_DATA=/ \
     ANDROID_DNS_MODE=local \
-    ANDROID_ROOT=/system
+    ANDROID_ROOT=/system \
+    CROSS_CMAKE_SYSTEM_NAME=Android \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=aarch64 \
+    CROSS_CMAKE_CRT=android \
+    CROSS_CMAKE_OBJECT_FLAGS="-DANDROID -ffunction-sections -fdata-sections -fPIC"
--- a/docker/Dockerfile.aarch64-unknown-linux-gnu
+++ b/docker/Dockerfile.aarch64-unknown-linux-gnu
@@ -29,15 +29,22 @@ COPY linux-image.sh /
 RUN /linux-image.sh aarch64
 
 COPY linux-runner base-runner.sh /
+COPY toolchain.cmake /opt/toolchain.cmake
 
 ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-gnu-
 ENV CROSS_SYSROOT=/usr/aarch64-linux-gnu
+ENV CROSS_TARGET_RUNNER="/linux-runner aarch64"
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
-    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="/linux-runner aarch64" \
+    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="$CROSS_TARGET_RUNNER" \
     AR_aarch64_unknown_linux_gnu="$CROSS_TOOLCHAIN_PREFIX"ar \
     CC_aarch64_unknown_linux_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CXX_aarch64_unknown_linux_gnu="$CROSS_TOOLCHAIN_PREFIX"g++ \
+    CMAKE_TOOLCHAIN_FILE_aarch64_unknown_linux_gnu=/opt/toolchain.cmake \
     BINDGEN_EXTRA_CLANG_ARGS_aarch64_unknown_linux_gnu="--sysroot=$CROSS_SYSROOT" \
     QEMU_LD_PREFIX="$CROSS_SYSROOT" \
     RUST_TEST_THREADS=1 \
-    PKG_CONFIG_PATH="/usr/lib/aarch64-linux-gnu/pkgconfig/:${PKG_CONFIG_PATH}"
+    PKG_CONFIG_PATH="/usr/lib/aarch64-linux-gnu/pkgconfig/:${PKG_CONFIG_PATH}" \
+    CROSS_CMAKE_SYSTEM_NAME=Linux \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=aarch64 \
+    CROSS_CMAKE_CRT=gnu \
+    CROSS_CMAKE_OBJECT_FLAGS="-ffunction-sections -fdata-sections -fPIC"
--- a/docker/Dockerfile.armv7-linux-androideabi
+++ b/docker/Dockerfile.armv7-linux-androideabi
@@ -34,15 +34,18 @@ RUN /android-system.sh arm
 
 ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-androideabi-
 ENV CROSS_SYSROOT=/android-ndk/sysroot
+ENV CROSS_ANDROID_SDK=$ANDROID_SDK
 COPY android-symlink.sh /
 RUN /android-symlink.sh arm arm-linux-androideabi
 
 COPY android-runner /
+COPY android.cmake /opt/toolchain.cmake
 
 # Libz is distributed in the android ndk, but for some unknown reason it is not
 # found in the build process of some crates, so we explicit set the DEP_Z_ROOT
+ENV CROSS_TARGET_RUNNER="/android-runner arm"
 ENV CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
-    CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_RUNNER="/android-runner arm" \
+    CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_RUNNER="$CROSS_TARGET_RUNNER" \
     AR_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"ar \
     AS_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"as \
     CC_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"gcc \
@@ -56,6 +59,7 @@ ENV CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     SIZE_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"size \
     STRINGS_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"strings \
     STRIP_armv7_linux_androideabi="$CROSS_TOOLCHAIN_PREFIX"strip \
+    CMAKE_TOOLCHAIN_FILE_armv7_linux_androideabi=/opt/toolchain.cmake \
     BINDGEN_EXTRA_CLANG_ARGS_armv7_linux_androideabi="--sysroot=$CROSS_SYSROOT" \
     DEP_Z_INCLUDE="$CROSS_SYSROOT/usr/include/" \
     RUST_TEST_THREADS=1 \
@@ -63,4 +67,8 @@ ENV CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     TMPDIR=/tmp/ \
     ANDROID_DATA=/ \
     ANDROID_DNS_MODE=local \
-    ANDROID_ROOT=/system
+    ANDROID_ROOT=/system \
+    CROSS_CMAKE_SYSTEM_NAME=Android \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=armv7-a \
+    CROSS_CMAKE_CRT=android \
+    CROSS_CMAKE_OBJECT_FLAGS="--target=arm-linux-androideabi -DANDROID -ffunction-sections -fdata-sections -fPIC --target=armv7-linux-androideabi"
--- a/docker/Dockerfile.i686-pc-windows-gnu
+++ b/docker/Dockerfile.i686-pc-windows-gnu
@@ -34,13 +34,22 @@ RUN mkdir -p /usr/lib/binfmt-support/ && \
 COPY windows-entry.sh /
 ENTRYPOINT ["/windows-entry.sh"]
 
+COPY toolchain.cmake /opt/toolchain.cmake
+
 # for why we always link with pthread support, see:
 # https://github.com/cross-rs/cross/pull/1123#issuecomment-1312287148
 ENV CROSS_TOOLCHAIN_PREFIX=i686-w64-mingw32-
+ENV CROSS_TOOLCHAIN_SUFFIX=-posix
 ENV CROSS_SYSROOT=/usr/i686-w64-mingw32
-ENV CARGO_TARGET_I686_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
-    CARGO_TARGET_I686_PC_WINDOWS_GNU_RUNNER=wine \
+ENV CROSS_TARGET_RUNNER="wine"
+ENV CARGO_TARGET_I686_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc"$CROSS_TOOLCHAIN_SUFFIX" \
+    CARGO_TARGET_I686_PC_WINDOWS_GNU_RUNNER="$CROSS_TARGET_RUNNER" \
     AR_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"ar \
-    CC_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
-    CXX_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"g++-posix \
-    BINDGEN_EXTRA_CLANG_ARGS_i686_pc_windows_gnu="--sysroot=$CROSS_SYSROOT"
+    CC_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc"$CROSS_TOOLCHAIN_SUFFIX" \
+    CXX_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"g++"$CROSS_TOOLCHAIN_SUFFIX" \
+    CMAKE_TOOLCHAIN_FILE_i686_pc_windows_gnu=/opt/toolchain.cmake \
+    BINDGEN_EXTRA_CLANG_ARGS_i686_pc_windows_gnu="--sysroot=$CROSS_SYSROOT" \
+    CROSS_CMAKE_SYSTEM_NAME=Windows \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=x86 \
+    CROSS_CMAKE_CRT=gnu \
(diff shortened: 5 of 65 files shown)


Alexhuszagh:
Note that this isn't close to complete yet. This just adds the CMake unittests.
bors try

Alexhuszagh:
bors try --target *android*

Alexhuszagh:
bors try --target *android*

Alexhuszagh:
This currently uses a `rust-cmake-hello-world` which won't work for `no_std` targets, so we really should have both a library and a binary, where CMake module just does a `no_std` calculation or something.

Alexhuszagh:
I've fixed CMake support for Android and newlib here, as well as added C++ support for newlib.

bors try --target *android* --target *none*

ci/test.sh
```diff
@@ -71,8 +71,6 @@ main() {
         popd
 
         rm -rf "${td}"
-
```
Alexhuszagh, line 74: We now want to run the CMake tests as well.

ci/test.sh
```diff
@@ -166,7 +164,7 @@ main() {
     fi
 
     # Test C++ support
-    if (( ${CPP:-0} )); then
+    if (( ${STD:-0} )) && (( ${CPP:-0} )); then
```
Alexhuszagh, line 167: This assumes we have C++ and std support, so we have to make that explicit.

ci/test.sh
```diff
@@ -197,6 +195,22 @@ main() {
...
     fi
+
+    # tst cmake support
+    td="$(mkcargotemp -d)"
+
+    git clone --depth 1 https://github.com/cross-rs/rust-cmake-hello-world "${td}"
+
+    pushd "${td}"
+    retry cargo fetch
+    if (( ${STD:-0} )) && (( ${RUN:-0} )); then
```
Alexhuszagh, line 206: The `rust-cmake-hello-world` has both a `no_std` lib and a `std` binary, so we can check support for either/or.

docker/Dockerfile.thumbv6m-none-eabi
```diff
@@ -10,14 +10,22 @@ RUN /cmake.sh
 COPY xargo.sh /
 RUN /xargo.sh
 
-COPY qemu.sh /
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
     gcc-arm-none-eabi \
-    libnewlib-arm-none-eabi && \
-    /qemu.sh arm
+    libnewlib-arm-none-eabi \
+    libstdc++-arm-none-eabi-newlib
```
Alexhuszagh, line 16: These targets actually support C++, and `gcc-arm-none-eabi` actually contains the C++ compiler, just not `libstdc++v3`. If we install that, these targets support C++ just fine.

docker/Dockerfile.thumbv6m-none-eabi
```diff
@@ -10,14 +10,22 @@ RUN /cmake.sh
...
+
+COPY qemu.sh /
+RUN /qemu.sh arm
+
+COPY newlib.cmake /opt/newlib.cmake
+ENV CMAKE_TOOLCHAIN_FILE=/opt/newlib.cmake
+ENV CROSS_NEWLIB_OBJECT_FLAGS="-ffunction-sections -fdata-sections -mthumb -march=armv6s-m"
 
-ENV QEMU_CPU=cortex-m3 \
+ENV QEMU_CPU=cortex-m1 \
```
Alexhuszagh, line 25: `cortex-m3` is an ARMv7-M target, `cortex-m1` is the proper CPU here for ARMv6-M (see [ARM Cortex-M](https://en.wikipedia.org/wiki/ARM_Cortex-M#Cortex-M1)).

docker/Dockerfile.thumbv6m-none-eabi
```diff
@@ -10,14 +10,22 @@ RUN /cmake.sh
...
-    /qemu.sh arm
+    libnewlib-arm-none-eabi \
+    libstdc++-arm-none-eabi-newlib
+
+COPY qemu.sh /
+RUN /qemu.sh arm
+
+COPY newlib.cmake /opt/newlib.cmake
+ENV CMAKE_TOOLCHAIN_FILE=/opt/newlib.cmake
+ENV CROSS_NEWLIB_OBJECT_FLAGS="-ffunction-sections -fdata-sections -mthumb -march=armv6s-m"
```
Alexhuszagh, line 23: See `newlib.cmake` below for why these defaults are provided.

docker/android.cmake
```diff
@@ -0,0 +1,33 @@
+# toolchain file for android targets, see #1110
+
+set(CMAKE_SYSTEM_NAME Android)
+set(CMAKE_SYSTEM_PROCESSOR "$ENV{CROSS_ANDROID_ARCH}")
```
Alexhuszagh, line 4: This is not the same as the arch elsewhere: it must be `i686`, `aarch64`, `x86_64`, `armv7-a`, or `armv5te`. Just setting `arm` will not work.

docker/android.cmake
```diff
@@ -0,0 +1,33 @@
...
+set(CMAKE_ANDROID_STANDALONE_TOOLCHAIN /android-ndk)
+set(CMAKE_ANDROID_API "$ENV{CROSS_ANDROID_SDK}")
```
Alexhuszagh, line 6: We can't auto-detect the Android API, so we manually set it. CMake will still warn about this, but it's irrelevant since it's manually set.

docker/android.cmake
```diff
@@ -0,0 +1,33 @@
+# toolchain file for android targets, see #1110
+
+set(CMAKE_SYSTEM_NAME Android)
+set(CMAKE_SYSTEM_PROCESSOR "$ENV{CROSS_ANDROID_ARCH}")
+set(CMAKE_ANDROID_STANDALONE_TOOLCHAIN /android-ndk)
```
Alexhuszagh, line 5: Since we use a standalone Android NDK with the `/android-ndk/sysroot` with no version prefix, we need to use `CMAKE_ANDROID_STANDALONE_TOOLCHAIN ` and not `CMAKE_SYSROOT` nor `CMAKE_ANDROID_NDK`.

See the docs [here](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android).

docker/android.cmake
```diff
@@ -0,0 +1,33 @@
...
+set(CMAKE_ANDROID_API "$ENV{CROSS_ANDROID_SDK}")
+
+# not all of these are standard, however, they're common enough
+# that it's good practice to define them.
+set(prefix "$ENV{CROSS_ANDROID_TARGET}")
+set(CMAKE_C_COMPILER "${prefix}-gcc")
```
Alexhuszagh, line 11: Only `CMAKE_AR`, `CMAKE_RANLIB`, `CMAKE_C_COMPILER`, `CMAKE_ASM_COMPILER`, and `CMAKE_CXX_COMPILER` are standard, however, other tools will often use values provided by those variables even though they're not standard. Many toolchain files will provide them.

docker/android.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+# required because otherwise it tries to detect compiler features
+# from `/android-ndk/bin/clang`, which fails. these macros are
+# technically deprecated, but required for proper function
+include(CMakeForceCompiler)
+set(CMAKE_WARN_DEPRECATED FALSE)
+CMAKE_FORCE_C_COMPILER("${CMAKE_C_COMPILER}" GNU)
+CMAKE_FORCE_CXX_COMPILER("${CMAKE_CXX_COMPILER}" GNU)
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
```
Alexhuszagh, line 30: Standard for cross-compiling.

docker/newlib.cmake
```diff
@@ -0,0 +1,35 @@
+# toolchain file for newlib targets, see #1110
+
+# if we don't set generic, it tries to test the compiler which fails
```
Alexhuszagh, line 3: This comment needs to be changed: if we don't set it to generic, it will complain the system is unknown.

targets.toml
```diff
@@ -440,22 +440,26 @@ std = true
 [[target]]
 target = "thumbv6m-none-eabi"
 os = "ubuntu-latest"
-std = true
```
Alexhuszagh, line 443: All these targets do not support `std`, but they do now support C++.

Alexhuszagh:
bors try --target *android* --target *none*
This should be ready for review now.

ci/test.sh
```diff
@@ -197,6 +209,22 @@ main() {
 
         rm -rf "${td}"
     fi
+
+    # tst cmake support
```
Emilgardis, line 213: typo

Emilgardis (approved):

Alexhuszagh:
This still has minor issues, and I'm currently testing it with more complex projects:
https://github.com/Alexhuszagh/rust-re2

eli-schwartz:
> and Meson itself tries to entirely recreate the [Rust build system](https://mesonbuild.com/Rust.html) which is absolutely cursed

Alas, we live in a world where cargo doesn't really seem to want to provide a (working) build plan, see e.g. https://github.com/rust-lang/cargo/issues/7614. You should see how "absolutely cursed" bazel's solution to the same problem is. ;)

If and when Rust decides this is an interesting problem to solve for real in a non "so unstable we need to delete it" way, you can be sure literally everyone trying to integrate rust into multi-language build systems of a wide variety of descriptions, will be delighted to know. :p

docker/android.cmake
Alexhuszagh: It turns out setting these manually for the `CMAKE_ASM_COMPILER`, etc., fails badly with downstream configurations. See https://github.com/cross-rs/cross/issues/1110#issuecomment-1303781124.

Alexhuszagh:
I think I've fixed Android support with more complex targets:
bors try --target *android* --target *none*

Alexhuszagh:
This is currently failing due to missing locks on Android ARMv5te, which isn't supported by newer [NDKs](https://developer.android.com/ndk/guides/abis) anyway:

> Note: Historically the NDK supported ARMv5 (armeabi), and 32-bit and 64-bit MIPS, but support for these ABIs was removed in NDK r17.

It's probably just worth disabling the `re2` test for that platform specifically. `__libcpp_signed_lock_free` isn't present in `libc++_shared.so` for `arm-linux-androideabi`.

Alexhuszagh:
bors try --target *android* --target *none*

ci/test.sh
```diff
@@ -197,6 +209,29 @@ main() {
...
+
+    git clone \
+        --recursive \
+        --depth 1 \
+        https://github.com/cross-rs/rust-cmake-hello-world "${td}"
+
+    pushd "${td}"
+    retry cargo fetch
+    if [[ "${TARGET}" == "arm-linux-androideabi" ]]; then
+        # ARMv5te isn't supported anymore by Android, which produces missing
```
Alexhuszagh, line 224: `re2` might not be supported on some other platforms without native atomics, it's probably better to allow-list certain architectures.

Alexhuszagh:
bors try

ci/test.sh
Alexhuszagh: Looks like there's numerous linking issues with every musl target, likely due to a lack of locale support, but also including much more than just that. Missing symbols include:
- `__extendsftf2`
- `floor`
- `wmemcpy`
- `pthread_mutex_unlock`
- `setlocale`

Which is some soft-float intrinsics, wide character support, pthreads, and locale functions.

Alexhuszagh: Looks like the Windows GNU target issues are not our fault: re2 requires C++17 support for [shared_mutex](https://en.cppreference.com/w/cpp/thread/shared_mutex) while only providing flags for C++11 support. Let's see if we can go to an older re2 version...

https://github.com/google/re2/blob/7a65faf439295e941baa6640a717d89c1f13e9cd/CMakeLists.txt#L80-L84

Alexhuszagh: Another issue: the package we install on Ubuntu 20.04 doesn't support `std::thread`, since we have no thread support. We'll likely want to upgrade to Ubuntu 22.04 for the Windows GNU targets.

Alexhuszagh: x86_64 CentOS doesn't have `int32_t`... ok.
https://github.com/cross-rs/cross/actions/runs/3406755182/jobs/5665741079#step:12:2303

Alexhuszagh: For sufficiently-complex musl targets, it seems we need to either link to dynamic libstdc++ (see #902), or we can always link to libc (which will only provide missing symbols for the C runtime).

Adding `RUSTFLAGS="-C link-arg=-lc"` fixes the issue. @Emilgardis  thoughts on the best solution? Always link to libc, or revert #902, knowing that the latter will fail with a lot of targets? I think the latter is probably preferable, but it would require a custom musl linker that always adds `-lc`.

Alexhuszagh:
bors try
This still should be failing on musl and Windows GNU, however, the CentOS builds should now be fixed.

Alexhuszagh:
We should also probably set `CMAKE_CROSSCOMPILING_EMULATOR` as described here https://github.com/cross-rs/cross/issues/1110#issuecomment-1305590929.

Alexhuszagh:
I'm going to mark this as a draft until #1118, #1123, and #1125 are merged, since it will be easier to build upon them rather than the inverse.

docker/toolchain.cmake
```diff
@@ -0,0 +1,57 @@
...
+set(CMAKE_OBJCOPY "${prefix}objcopy")
+set(CMAKE_OBJDUMP "${prefix}objdump")
+set(CMAKE_RANLIB "${prefix}ranlib")
+set(CMAKE_STRIP "${prefix}strip")
+
+# these are cached so any build system that compiled outside of the rust
+# build system, such as a third-party cmake build and install of a shared
+# library, will still work. however, cmake-rs can override these values
+if(DEFINED ENV{CROSS_CMAKE_OBJECT_FLAGS})
+    set(CMAKE_C_FLAGS "$ENV{CROSS_CMAKE_OBJECT_FLAGS}" CACHE INTERNAL "C Compiler options")
```
Alexhuszagh, line 32: We cache these internally so `cmake-rs` can overwrite them.

docker/toolchain.cmake
```diff
@@ -0,0 +1,57 @@
...
+    set(CMAKE_CXX_FLAGS "$ENV{CROSS_CMAKE_OBJECT_FLAGS}" CACHE INTERNAL "C++ Compiler options")
+    set(CMAKE_ASM_FLAGS "$ENV{CROSS_CMAKE_OBJECT_FLAGS}" CACHE INTERNAL "ASM Compiler options")
+endif()
+
+# if cross-compiling, we need to disable where the root path
+# is found and also provide our own sysroot
+set(native ENV{CROSS_CMAKE_NATIVE})
```
Alexhuszagh, line 39: This can be removed, since the native Dockerfiles don't actually use a custom toolchain file.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -34,15 +34,21 @@ RUN /android-system.sh arm64
...
 ENV CROSS_SYSROOT=/android-ndk/sysroot
+ENV CROSS_ANDROID_SDK=$ANDROID_SDK
 COPY android-symlink.sh /
 RUN /android-symlink.sh aarch64 aarch64-linux-android
 
 COPY android-runner /
 
+COPY android.cmake /opt/android.cmake
+ENV CMAKE_TOOLCHAIN_FILE=/opt/android.cmake
+ENV CROSS_CMAKE_OBJECT_FLAGS="-DANDROID -ffunction-sections -fdata-sections -fPIC"
```
Alexhuszagh, line 45: All of the `CROSS_CMAKE_OBJECT_FLAGS` were generated manually by checking a project using CMake that failed and output the `CMAKE_C_FLAGS`, etc., and confirmed the C, C++, and ASM flags were all the same.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -34,15 +34,21 @@ RUN /android-system.sh arm64
...
+
 # Libz is distributed in the android ndk, but for some unknown reason it is not
 # found in the build process of some crates, so we explicit set the DEP_Z_ROOT
+ENV CROSS_TARGET_RUNNER="/android-runner aarch64"
```
Alexhuszagh, line 49: This allows us to use it both for the cargo runner and for `CMAKE_CROSSCOMPILING_EMULATOR`, which is required when using `try_run` in some CMake build systems.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -63,4 +69,7 @@ ENV CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     TMPDIR=/tmp/ \
     ANDROID_DATA=/ \
     ANDROID_DNS_MODE=local \
-    ANDROID_ROOT=/system
+    ANDROID_ROOT=/system \
+    CROSS_CMAKE_SYSTEM_NAME=Android \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=aarch64 \
+    CROSS_CMAKE_CRT=android
```
Alexhuszagh, line 75: `CROSS_CMAKE_CRT` is technically not needed for this: it's only needed for `newlib` and we could probably do `CROSS_CMAKE_NEWLIB` but I figured it might be nice just to have this. It might be better to check for `*-none-*` though since we only use it to check if we should build an executable or staticlib when testing if the compiler works.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -63,4 +69,7 @@ ENV CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     TMPDIR=/tmp/ \
     ANDROID_DATA=/ \
     ANDROID_DNS_MODE=local \
-    ANDROID_ROOT=/system
+    ANDROID_ROOT=/system \
+    CROSS_CMAKE_SYSTEM_NAME=Android \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=aarch64 \
```
Alexhuszagh, line 74: `CROSS_CMAKE_SYSTEM_PROCESSOR` is required for `CMAKE_SYSTEM_PROCESSOR` and `CROSS_CMAKE_SYSTEM_NAME` is just provided for consistency with every other toolchain.

docker/Dockerfile.aarch64-unknown-linux-gnu
```diff
@@ -30,13 +30,21 @@ RUN /linux-image.sh aarch64
 
 COPY linux-runner base-runner.sh /
 
+COPY toolchain.cmake /opt/toolchain.cmake
+ENV CMAKE_TOOLCHAIN_FILE=/opt/toolchain.cmake
```
Alexhuszagh, line 34: We can use a single CMake toolchain file for every target besides Android. The `CROSS_CMAKE_OBJECT_FLAGS` are provided so the same flags are used by default when CMake is used outside of `cmake-rs`, for whatever reason. `cmake-rs` will override these, so it's not a major deal.

docker/toolchain.cmake
```diff
@@ -0,0 +1,57 @@
...
+# required so CMAKE_CROSSCOMPILING_EMULATOR is set,
+# as well for embedded systems and other targets.
+#
+# all embedded systems without an OS should set the system name to generic
+# https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html
+
+set(CMAKE_SYSTEM_NAME "$ENV{CROSS_CMAKE_SYSTEM_NAME}")
+set(CMAKE_SYSTEM_PROCESSOR "$ENV{CROSS_CMAKE_SYSTEM_PROCESSOR}")
+if(DEFINED ENV{CROSS_TARGET_RUNNER})
+    set(CMAKE_CROSSCOMPILING_EMULATOR $ENV{CROSS_TARGET_RUNNER})
```
Alexhuszagh, line 11: Not all targets have runners, disable it if they don't.

docker/Dockerfile.x86_64-unknown-netbsd
```diff
@@ -13,10 +13,17 @@ RUN /xargo.sh
...
 ENV CROSS_SYSROOT=/usr/local/x86_64-unknown-netbsd
 ENV CARGO_TARGET_X86_64_UNKNOWN_NETBSD_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     AR_x86_64_unknown_netbsd="$CROSS_TOOLCHAIN_PREFIX"ar \
     CC_x86_64_unknown_netbsd="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CXX_x86_64_unknown_netbsd="$CROSS_TOOLCHAIN_PREFIX"g++ \
-    BINDGEN_EXTRA_CLANG_ARGS_x86_64_unknown_netbsd="--sysroot=$CROSS_SYSROOT"
+    BINDGEN_EXTRA_CLANG_ARGS_x86_64_unknown_netbsd="--sysroot=$CROSS_SYSROOT" \
+    CROSS_CMAKE_SYSTEM_NAME=NetBSD \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=x86_64 \
+    CROSS_CMAKE_CRT=netbsd
```
Alexhuszagh, line 29: All of the BSDs use their own `libc` variants, so I've given them all different names.

docker/Dockerfile.x86_64-unknown-linux-gnu.centos
```diff
@@ -26,6 +26,9 @@ COPY --from=0 /qemu /qemu
 
 COPY linux-runner base-runner.sh /
 
+# NOTE: this target is only supported on linux/amd64 hosts,
```
Alexhuszagh, line 29: I think we can actually remove this file since it's never used if `native.centos` is used instead.

docker/Dockerfile.x86_64-unknown-illumos
```diff
@@ -12,11 +12,18 @@ RUN /xargo.sh
...
 ENV PATH=$PATH:/usr/local/x86_64-unknown-illumos/bin/ \
     CARGO_TARGET_X86_64_UNKNOWN_ILLUMOS_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     AR_x86_64_unknown_illumos="$CROSS_TOOLCHAIN_PREFIX"ar \
     CC_x86_64_unknown_illumos="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CXX_x86_64_unknown_illumos="$CROSS_TOOLCHAIN_PREFIX"g++ \
-    BINDGEN_EXTRA_CLANG_ARGS_sparcv9_sun_solaris="--sysroot=$CROSS_SYSROOT"
+    BINDGEN_EXTRA_CLANG_ARGS_sparcv9_sun_solaris="--sysroot=$CROSS_SYSROOT" \
+    CROSS_CMAKE_SYSTEM_NAME=illumos \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=x86_64 \
+    CROSS_CMAKE_CRT=solaris
```
Alexhuszagh, line 29: Illumos and Solaris are very similar, so I gave it the same CRT name.

docker/Dockerfile.wasm32-unknown-emscripten
```diff
@@ -17,5 +17,7 @@ RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
 
 ENV CROSS_TOOLCHAIN_PREFIX=em
 ENV CROSS_SYSROOT=/emsdk/upstream/emscripten/cache/sysroot
-ENV CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node \
-    BINDGEN_EXTRA_CLANG_ARGS_wasm32_unknown_emscripten="--sysroot=$CROSS_SYSROOT"
+ENV CROSS_TARGET_RUNNER="node"
+ENV CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER="$CROSS_TARGET_RUNNER" \
+    BINDGEN_EXTRA_CLANG_ARGS_wasm32_unknown_emscripten="--sysroot=$CROSS_SYSROOT" \
+    CMAKE_TOOLCHAIN_FILE=/emsdk/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake
```
Alexhuszagh, line 23: wasm32 and asmjs provide their own CMake toolchain files.

docker/Dockerfile.wasm32-unknown-emscripten
```diff
@@ -17,5 +17,7 @@ RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
 
 ENV CROSS_TOOLCHAIN_PREFIX=em
 ENV CROSS_SYSROOT=/emsdk/upstream/emscripten/cache/sysroot
-ENV CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node \
-    BINDGEN_EXTRA_CLANG_ARGS_wasm32_unknown_emscripten="--sysroot=$CROSS_SYSROOT"
+ENV CROSS_TARGET_RUNNER="node"
```
Alexhuszagh, line 20: We could possibly provide the additional pthread flags to the runner. See:
https://github.com/emscripten-core/emscripten/blob/e9ba3598df8f219b8c9c505c4ed1ed4618001123/emcmake.py#L44

docker/Dockerfile.thumbv7m-none-eabi
```diff
@@ -10,15 +10,26 @@ RUN /cmake.sh
...
+
+COPY qemu.sh /
+RUN /qemu.sh arm
+
+COPY toolchain.cmake /opt/toolchain.cmake
+ENV CMAKE_TOOLCHAIN_FILE=/opt/toolchain.cmake
+ENV CROSS_CMAKE_OBJECT_FLAGS="-ffunction-sections -fdata-sections -mthumb -march=armv7-m"
 
 ENV CROSS_TOOLCHAIN_PREFIX=arm-none-eabi-
+ENV CROSS_SYSROOT="/usr/lib/arm-none-eabi"
```
Alexhuszagh, line 26: The THUMB newlib targets include their own binutils and include headers under this directory, so it's effectively a sysroot for most intents and purposes.

Alexhuszagh:
bors try

Alexhuszagh:
bors try-

Alexhuszagh:
bors try

Alexhuszagh:
bors try *bsd

Alexhuszagh:
bors try --target *bsd

Alexhuszagh:
bors try

Alexhuszagh:
We can actually make these changes interface less with the image: `CMAKE_TOOLCHAIN_FILE_thumbv6m_none_eabi` for example, can be provided, which `cmake-rs` will look at first.

Alexhuszagh:
bors try --target x86_64-unknown-dragonfly

Alexhuszagh:
bors try --target x86_64-unknown-dragonfly

docker/Dockerfile.i686-pc-windows-gnu
```diff
@@ -34,12 +34,20 @@ RUN mkdir -p /usr/lib/binfmt-support/ && \
...
 ENV CROSS_SYSROOT=/usr/i686-w64-mingw32
+ENV CROSS_TARGET_RUNNER="wine"
 ENV CARGO_TARGET_I686_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
-    CARGO_TARGET_I686_PC_WINDOWS_GNU_RUNNER=wine \
+    CARGO_TARGET_I686_PC_WINDOWS_GNU_RUNNER="$CROSS_TARGET_RUNNER" \
     CC_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
     CXX_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"g++-posix \
-    BINDGEN_EXTRA_CLANG_ARGS_i686_pc_windows_gnu="--sysroot=$CROSS_SYSROOT"
+    BINDGEN_EXTRA_CLANG_ARGS_i686_pc_windows_gnu="--sysroot=$CROSS_SYSROOT" \
+    CROSS_CMAKE_SYSTEM_NAME=Windows \
```
Alexhuszagh, line 51: For some reason, `cmake-rs` passes this as `windows` only for 32-bit Windows GNU but the correct value is `Windows`, and `Platform/windows` is unknown to CMake. So, this differs **slightly** from cmake-rs.

Alexhuszagh:
bors try --target *pc-windows-gnu

Alexhuszagh:
bors try --target *none*

docker/toolchain.cmake
```diff
@@ -0,0 +1,56 @@
...
+if(DEFINED ENV{CROSS_TARGET_RUNNER})
+    set(runner "$ENV{CROSS_TARGET_RUNNER}")
+    separate_arguments(runner)
+    set(CMAKE_CROSSCOMPILING_EMULATOR ${runner})
+endif()
+
+# not all of these are standard, however, they're common enough
+# that it's good practice to define them.
+set(prefix "$ENV{CROSS_TOOLCHAIN_PREFIX}")
+set(CMAKE_C_COMPILER "${prefix}gcc")
```
Alexhuszagh, line 19: This needs a suffix if we're using Windows for the `*posix` variants.

Alexhuszagh:
bors try --target *pc-windows-gnu

Alexhuszagh:
bors try
This should pass for all but the MIPS64 musl targets.

Alexhuszagh:
Everything should be fixed. I'm going to roll out the final musl patches to a separate PR, and then fix all the toolchain file names to ensure they're target specific, and this PR is ready.

bors try

Alexhuszagh:
bors try

docker/Dockerfile.armv7-linux-androideabi
```diff
@@ -56,11 +59,16 @@ ENV CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
...
     HOME=/tmp/ \
     TMPDIR=/tmp/ \
     ANDROID_DATA=/ \
     ANDROID_DNS_MODE=local \
-    ANDROID_ROOT=/system
+    ANDROID_ROOT=/system \
+    CROSS_CMAKE_SYSTEM_NAME=Android \
+    CROSS_CMAKE_SYSTEM_PROCESSOR=armv7-a \
+    CROSS_CMAKE_CRT=android \
+    CROSS_CMAKE_OBJECT_FLAGS="--target=arm-linux-androideabi -DANDROID -ffunction-sections -fdata-sections -fPIC --target=armv7-linux-androideabi"
```
Alexhuszagh, line 74: I believe this change is correct, with `--target` provided twice, since that's what `cmake-rs` does as of `0.1.49` and `cc-rs` as of `1.0.77` (this is using Rust 1.64.0).

Alexhuszagh:
bors try

Alexhuszagh:
bors try

Emilgardis (approved):
lgtm

Alexhuszagh:
bors r=Emilgardis

PR #1114, opened 2022-11-05
Remove `std = true` for newlib targets.

### LINKED ISSUE #1113: Newlib Targets Incorrectly Report `std` Support in `targets.toml`.

Alexhuszagh:
Closes #1113.

### DIFF
--- a/targets.toml
+++ b/targets.toml
@@ -440,22 +440,22 @@ std = true
 [[target]]
 target = "thumbv6m-none-eabi"
 os = "ubuntu-latest"
-std = true
+std = false
 
 [[target]]
 target = "thumbv7em-none-eabi"
 os = "ubuntu-latest"
-std = true
+std = false
 
 [[target]]
 target = "thumbv7em-none-eabihf"
 os = "ubuntu-latest"
-std = true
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
Actually I'll probably roll all of this into #1112, since we've got a lot more things we can probably patch here.

PR #1115, opened 2022-11-05
Initial version of UI tests.

Alexhuszagh:
This is a very initial draft, but the idea is to add UI tests (which we'll add more as this gets further along).

UI tests still to add:
- [ ] Remote
- [ ] Remote with incorrect endpoint
- [ ] Invalid image platform
- [ ] Invalid pre-build script
- [ ] Invalid pre-build command
- [ ] Invalid pre-build dockerfile
- [ ] ... Many more config options
- [ ] A few other config flags

Any suggestions for additional UI tests to add, please suggest them below. This also needs to add `ci/test-ui.sh` into the CI.


src/cli.rs
```diff
@@ -54,7 +54,7 @@ pub fn fmt_subcommands(stdout: &str, msg_info: &mut MessageInfo) -> Result<()> {
     }
     if !host.is_empty() {
         msg_info.print("Host Commands:")?;
-        for line in &cross {
+        for line in &host {
```
Alexhuszagh, line 57: This is a bug fix identified by the UI tests. Handy.

xtask/src/temp.rs
```diff
@@ -0,0 +1,35 @@
+use std::path::{Path, PathBuf};
```
Alexhuszagh, line 1: We might not need this. I felt it might be handy to integrate it with our CI scripts, but it's probably unneeded.

xtask/src/ui.rs
```diff
@@ -0,0 +1,480 @@
...
+{}
+
+stdout
+======
+{}
+
+stderr
+======
+{}
+"#,
```
Alexhuszagh, line 60: This is just so we get a very pretty diagnostic output of the actual error/output messages, the command, and the exit status if anything fails our expectations. This makes debugging way easier.

xtask/src/ui.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    fn outside_package(&self) -> cross::Result<()> {
+        self.failure()?;
+        self.fallsback()?;
+        self.no_metadata()?;
+        self.no_stdout()?;
+        self.check("did not error", |x| {
+            x.stderr.contains("error: could not find `Cargo.toml`")
+        })
+    }
```
Alexhuszagh, line 116: Since we only run all these tests once, **most** of these can be moved out of here.

xtask/src/ui.rs
```diff
@@ -0,0 +1,480 @@
...
+                path.pop();
+            }
+            path
+        })
+        .unwrap()
+}
+
+// Adapted from
+// https://github.com/assert-rs/assert_cmd/blob/363daac637b5f20ba3524ccf648de5f96f0dc45f/src/cargo.rs#L198-L208
+fn cargo_bin(bin: impl AsRef<str>, cwd: impl AsRef<Path>) -> cross::Result<Command> {
```
Alexhuszagh, line 356: This was the only item we needed from `assert_cmd`, which is very easy to implement itself. It's not worth dragging a large dependency, which prevents us from using `run_and_get_output` with the printed commands, and poorer debug messages, for ~30 lines of code.

xtask/src/ui.rs
```diff
@@ -0,0 +1,480 @@
...
+    cmd.arg(image);
+    if !msg_info.is_verbose() {
+        cmd.stderr(Stdio::null());
+    }
+    cmd.run(msg_info, !msg_info.is_verbose())
+        .map_err(Into::into)
+}
+
+fn build_cross(msg_info: &mut MessageInfo) -> cross::Result<()> {
+    msg_info.note("building cross: this might take a while")?;
```
Alexhuszagh, line 428: Running `xtask` doesn't necessarily build the cross binaries, so we need to run this prior. Build the full workspace, since `xtask` will benefit from it later.

Emilgardis:
I think instead of having methods to check the output we can have `.stderr` and `.stdout` file, i.e what `trycmd` does.

this way we can also present a diff and ensure every aspect is captured.

But this is definitely an interesting approach and may work very well 👍🏼

Alexhuszagh:
I've added the initial versions of `trycmd` (and discovered some very minor bugs in the process). There's a few issues, but it's mostly working well. Due to the limitations of the markdown format, we're using the toml/stderr/stdout for everything, so to avoid duplicating everything numerous times, there's a lot of symlinks.

Alexhuszagh:
I've completed reworked how CLI parsing works, and added an extensive set of integration tests. By default, these are ignored because they're very slow to run. I've also added in a regression test for #1141.

xtask/src/main.rs
```diff
@@ -8,6 +8,7 @@ pub mod crosstool;
 pub mod hooks;
 pub mod install_git_hooks;
 pub mod target_info;
+pub mod temp;
```
Alexhuszagh, line 11: I can get rid of `temp` now.

xtask/src/temp.rs
Alexhuszagh, line 1: This isn't used anymore, since we've switched to `trycmd`.

tests/cli.rs
```diff
@@ -0,0 +1,350 @@
+// simplifies addition of tests due to warning on last
+// of needless borrow on last test in block.
```
Emilgardis, line 2: this comment is uneccessary

Cargo.toml
```diff
@@ -63,6 +63,8 @@ lto = true
 regex = "1"
 once_cell = "1.15"
 ignore = "0.4"
+trycmd = "0.14.4"
+snapbox = "0.4"
```
Emilgardis, line 67: We could have this as its own workspace member alongside `cross` and `xtask`, not needed but just a thought

src/cli.rs
```diff
@@ -1,27 +1,415 @@
...
+impl<T> FlagOption<T> {
+    pub fn is_some(&self) -> bool {
+        matches!(self, Self::Some(_))
+    }
+
+    pub fn is_none(&self) -> bool {
+        matches!(self, Self::Some(_))
+    }
+
+    /// If the flag is present multiple times, always set to `Double`.
```
Emilgardis, line 33: this doc comment describes the behaviour for a certain case, but not what it does

src/tests.rs
```diff
@@ -138,7 +138,10 @@ release: {version}
 
 #[test]
 fn check_newlines() -> crate::Result<()> {
-    for file in walk_dir(get_cargo_workspace(), &[".git", "target"], |_| true) {
+    // tests contains the trycmd tests, which may have placeholders
```
Emilgardis, line 141:
```suggestion
    // these folders are not checked due to them containing binary files or files which should not have trailing newlines 
```

src/tests.rs
```diff
@@ -138,7 +138,10 @@ release: {version}
...
+    for file in walk_dir(get_cargo_workspace(), &[".git", "target", "tests"], |_| {
+        true
+    }) {
```
Emilgardis, line 144: instead of skipping the whole `tests` dir, instead skip if the file extension is `.stderr` or `.stdout`

Alexhuszagh:
This needs to handle escape sequences as well:
https://github.com/cross-rs/cross/actions/runs/3509739225/jobs/5879045993#step:6:238

Cargo.toml
Alexhuszagh: Is there any major advantage to that over faster compile times for the test suite?

ci/test-rest-args.sh
```diff
@@ -0,0 +1,39 @@
...
+# test that we correctly handle args after `--`
+
+set -x
+set -eo pipefail
+
+ci_dir=$(dirname "${BASH_SOURCE[0]}")
+ci_dir=$(realpath "${ci_dir}")
+. "${ci_dir}"/shared.sh
+
+main() {
```
Alexhuszagh, line 13: This needs a `clippy -- --dent warnings` just to ensure that `-Zbuild-std` is getting correctly added as well.

PR #1123, opened 2022-11-07
Use MinGW GCC with POSIX thread support when linking.

### LINKED ISSUE #1122: Add Support for C/C++ Dependencies with Threads for Windows GNU
### Checklist

### Describe your request

Currently, when compiling for Windows GNU targets (MinGW), any external C/C++ dependencies with threads (like `re2`) will compile, however, when linking them to Rust, there are numerous linker issues such as:

```bash
/usr/lib/gcc/x86_64-w64-mingw32/9.3-posix/include/c++/mutex:680: undefined reference to `std::__set_once_functor_lock_ptr(std::unique_lock<std::mutex>*)'
```

This is because we use the `*-gcc-posix` compilers for our dependencies, but only use the `*-gcc` variants for our Rust linker.

Alexhuszagh:
This changes from using WIN32 threads to POSIX threads when linking. MinGW (when using `libstdc++`) doesn't support C11 and C++11 threading support with WIN32 threads. This PR also makes WINE builds more resilient. Note that this will not affect any multi-threaded Rust code, which uses the Win32 API, calling `CreateThread` and `WaitForSingleObject` directly.

Closes #1122.

### DIFF
--- a/.changes/1123.json
+++ b/.changes/1123.json
@@ -1,0 +1,6 @@
+{
+    "description": "support external C/C++ dependencies using C11/C++11 threads for MinGW targets by switching linkers from `*-gcc` to `*gcc-posix`.",
+    "type": "added",
+    "issues": [1122],
+    "breaking": true
+}
--- a/.changes/591-1095.json
+++ b/.changes/591-1095.json
@@ -17,7 +17,7 @@
     },
     {
         "type": "changed",
-        "description": "updated powerpc64, riscv64, and sparc64 *-linux-gnu images to use a 6.x kernel instead of a 4.x kernel.",
+        "description": "updated powerpc64, riscv64, and sparc64 `*-linux-gnu` images to use a 6.x kernel instead of a 4.x kernel.",
         "breaking": true,
         "issues": [1094]
     }
--- a/docker/Dockerfile.i686-pc-windows-gnu
+++ b/docker/Dockerfile.i686-pc-windows-gnu
@@ -34,9 +34,11 @@ RUN mkdir -p /usr/lib/binfmt-support/ && \
 COPY windows-entry.sh /
 ENTRYPOINT ["/windows-entry.sh"]
 
+# for why we always link with pthread support, see:
+# https://github.com/cross-rs/cross/pull/1123#issuecomment-1312287148
 ENV CROSS_TOOLCHAIN_PREFIX=i686-w64-mingw32-
 ENV CROSS_SYSROOT=/usr/i686-w64-mingw32
-ENV CARGO_TARGET_I686_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
+ENV CARGO_TARGET_I686_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
     CARGO_TARGET_I686_PC_WINDOWS_GNU_RUNNER=wine \
     CC_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
     CXX_i686_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"g++-posix \
--- a/docker/Dockerfile.x86_64-pc-windows-gnu
+++ b/docker/Dockerfile.x86_64-pc-windows-gnu
@@ -32,9 +32,11 @@ RUN mkdir -p /usr/lib/binfmt-support/ && \
 COPY windows-entry.sh /
 ENTRYPOINT ["/windows-entry.sh"]
 
+# for why we always link with pthread support, see:
+# https://github.com/cross-rs/cross/pull/1123#issuecomment-1312287148
 ENV CROSS_TOOLCHAIN_PREFIX=x86_64-w64-mingw32-
 ENV CROSS_SYSROOT=/usr/x86_64-w64-mingw32
-ENV CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
+ENV CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
     CARGO_TARGET_X86_64_PC_WINDOWS_GNU_RUNNER=wine \
     CC_x86_64_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"gcc-posix \
     CXX_x86_64_pc_windows_gnu="$CROSS_TOOLCHAIN_PREFIX"g++-posix \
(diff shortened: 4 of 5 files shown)


Alexhuszagh:
It seems like Rust itself does this during CI, and WIN32 threads don't support `std::thread` with MinGW-w64 (is this still true?):

```yaml

          # 32/64-bit MinGW builds.
          #
          # We are using MinGW with POSIX threads since LLVM requires
          # C++'s std::thread which is disabled in libstdc++ with win32 threads.
          # FIXME: Libc++ doesn't have this limitation so we can avoid
          # winpthreads if we switch to it.
          #
          # Instead of relying on the MinGW version installed on CI we download
          # and install one ourselves so we won't be surprised by changes to CI's
          # build image.
          #
          # Finally, note that the downloads below are all in the `rust-lang-ci` S3
          # bucket, but they clearly didn't originate there! The downloads originally
          # came from the mingw-w64 SourceForge download site. Unfortunately
          # SourceForge is notoriously flaky, so we mirror it on our own infrastructure.
```

https://github.com/rust-lang/rust/blob/3f83906b30798bf61513fa340524cebf6676f9db/src/ci/github-actions/ci.yml#L588-L602

Alexhuszagh:
bors try --target *-pc-windows-gnu

Alexhuszagh:
You can verify this works by compiling and running with `re2` support [here](https://github.com/cross-rs/rust-cmake-hello-world):

```bash
# in the cross repository
cargo build-docker-image i686-pc-windows-gnu

# somewhere else
git clone https://github.com/cross-rs/rust-cmake-hello-world
cd rust-cmake-hello-world
CROSS_TARGET_I686_PC_WINDOWS_GNU_IMAGE=ghcr.io/cross-rs/i686-pc-windows-gnu:local cross run --target i686-pc-windows-gnu --features=re2 -vv
```

Alexhuszagh:
This extensively documents why we should use POSIX threads when linking rather than the default GCC without pthread support. MinGW only supports the C11 and C++11 threading model when POSIX threads are enabled (since GCC builds upon `gthread`, which is based on `pthread`). Note that linking with or without pthread support will not affect native Rust code, which uses `CreateThread` and `WaitForSingleObject` [directly](https://github.com/rust-lang/rust/blob/42325c525b9d3885847a3f803abe53c562d289da/library/std/src/sys/windows/thread.rs#L24-L60).

Internally, MinGW-w64 uses `winpthreads` for its pthreads implementation, which is based on top of the Win32 API although it does not use the exact same threading API (it does not use `CreateThread` directly). There are a few competing implementations, including include [mingw-std-threads](https://github.com/meganz/mingw-std-threads), [pthread-win32](https://github.com/GerHobbelt/pthread-win32), [mcfgthread](https://github.com/lhmouse/mcfgthread), and `winpthreads`, however, `winpthreads` is available by default and should be relatively competitive with any other pthreads implementation based on the Win32 API (although its [contribution](https://www.mingw-w64.org/contribute/) guidelines do mention `mcfgthread` and how its use of undocumented Windows NT syscalls for optimal performance).

Because pthreads require additional book-keeping on top of the Win32 API, they will inherently have additional [overhead](https://wiki.qt.io/MinGW-64-bit#GCC_Threading_model_.28posix_vs_win32.29) compared to calling `CreateThread` directly. However, this will mostly affect code using short-lived threads where spawning new threads is a major performance bottleneck.

Using MinGW with or without POSIX threads does change how `libgcc` works: with pthreads, it links to `winpthreads` while without it uses the Win32 API directly. However, without pthreads, MinGW does not support C++11 threads, which are widely used in external code due to their portability.

In short, the major trade-off her is for external C or C++ code that relies on how GCC [internally](https://stackoverflow.com/a/30390278/4131059) implements their threading features for `libgcc` and `libstdc++`. It does not affect user code, which can freely use both Win32 and POSIX threads in the same application. Since all portable client code in C or C++ requires C11 or C++11 threading features, which are only fully implemented with `winpthreads`, this has no potential pitfalls and should be implmented in all cases by default.

Some references on the internal implementations:

**winpthreads**
- `pthread_create`: https://github.com/mirror/mingw-w64/blob/7dcf65a7f26071c6aa26d5a5a8689089ae9a6740/mingw-w64-libraries/winpthreads/src/thread.c#L1579-L1684
- `pthread_join`: https://github.com/mirror/mingw-w64/blob/7dcf65a7f26071c6aa26d5a5a8689089ae9a6740/mingw-w64-libraries/winpthreads/src/thread.c#L1686-L1715

**mingw-std-threads**
- `std::thread::thread`: https://github.com/meganz/mingw-std-threads/blob/f6365f900fb9b1cd6014c8d1cf13ceacf8faf3de/mingw.thread.h#L183-L204
- `std::thread::join`: https://github.com/meganz/mingw-std-threads/blob/f6365f900fb9b1cd6014c8d1cf13ceacf8faf3de/mingw.thread.h#L212-L225

**pthread-win32**
- `pthread_create`: https://github.com/GerHobbelt/pthread-win32/blob/f70755704aa4a6a8f86423f1f07f00704fc99208/create.c#L51-L345
- `pthread_join`: https://github.com/GerHobbelt/pthread-win32/blob/f70755704aa4a6a8f86423f1f07f00704fc99208/pthread_join.c#L57-L164

**mcfgthread**
- `std::thread::thread`: https://github.com/lhmouse/mcfgthread/blob/132f69ede963cd2de081eb40c3bfcbdc865bb04d/src/thread.c#L36-L71
- `std::thread::join`: https://github.com/lhmouse/mcfgthread/blob/f31c6ed26828cbc5a1661823735b07ffd7bec7c4/src/libcxx.h#L548-L575

**std::thread (Rust)**
- `std::thread::spawn`: https://github.com/rust-lang/rust/blob/42325c525b9d3885847a3f803abe53c562d289da/library/std/src/sys/windows/thread.rs#L24-L60
- `std::thread::join`: https://github.com/rust-lang/rust/blob/42325c525b9d3885847a3f803abe53c562d289da/library/std/src/sys/windows/thread.rs#L72-L77

.changes/1123.json
```diff
@@ -0,0 +1,5 @@
+{
+    "description": "support external C/C++ dependencies using threads for MinGW targets.",
+    "type": "added",
+    "issues": [1122]
```
Emilgardis, line 4: I think this should have `breaking` and mention the cc used

Emilgardis (approved):
r=me with changelog change

.changes/591-1095.json
```diff
@@ -17,7 +17,7 @@
     },
     {
         "type": "changed",
-        "description": "updated powerpc64, riscv64, and sparc64 *-linux-gnu images to use a 6.x kernel instead of a 4.x kernel.",
+        "description": "updated powerpc64, riscv64, and sparc64 `*-linux-gnu` images to use a 6.x kernel instead of a 4.x kernel.",
```
Alexhuszagh, line 20: I quoted this with an inline code block just to avoid any potential formatting issues having an unescaped `*` could have if anything were to change (when validating the resulting changelog). Also, it's consistent with all our other changelogs entries like 727, 745, etc.

Alexhuszagh:
bors r=Emilgardis

PR #1125, opened 2022-11-07
Always link to libc when compiling musl targets.

Alexhuszagh:
This fixes a regression when changing musl targets to always statically link to libstdc++, and the static version of libstdc++ can reference but be missing requiring symbols from libc. Due to the linking order of Rust, these symbols sometimes are not resolved. This ensures that all binaries with external C++ dependencies cross-compiled for musl targets, such as `re2`, will be properly linked.

Closes #1124.

PR #1118, opened 2022-11-06
Added ARMv8-M newlib targets.

Alexhuszagh:
Closes #1116.

PR #1126, opened 2022-11-07
Export and use `CROSS_TOOLCHAIN_PREFIX`.

Alexhuszagh:


Alexhuszagh:
This will block #1112, #1118, #1123, and #1125 since it touches so many files, but simplifies a lot of the CMake files, as well as the linker files in #1125.
bors try

Alexhuszagh:
Windows and executable permissions strikes again :/
bors try --target *freebsd

Alexhuszagh:
bors try

docker/Dockerfile.mips64el-unknown-linux-muslabi64
```diff
@@ -27,15 +28,15 @@ RUN mkdir -p $CROSS_SYSROOT/usr/lib64
 RUN ln -s $CROSS_SYSROOT/usr/lib/libc.so $CROSS_SYSROOT/usr/lib64/libc.so
 RUN ln -s $CROSS_SYSROOT/usr/lib/libc.so.1 $CROSS_SYSROOT/usr/lib64/libc.so.1
 
-COPY mips64el-linux-musl-gcc.sh /usr/bin/
+COPY rust-1-65-musl-gcc.sh /usr/bin/"$CROSS_TOOLCHAIN_PREFIX"gcc.sh
```
Emilgardis, line 31: nice

docker/Dockerfile.native.centos
```diff
@@ -37,5 +37,6 @@ COPY linux-runner native-linux-runner base-runner.sh /
 
 ENV CROSS_TARGETARCH=$TARGETARCH
 ENV CROSS_TARGETVARIANT=$TARGETVARIANT
+ENV CROSS_TOOLCHAIN_PREFIX=
```
Emilgardis, line 40: do we need this to be defined as empty?

docker/Dockerfile.x86_64-unknown-linux-gnu.centos
```diff
@@ -26,5 +26,6 @@ COPY --from=0 /qemu /qemu
 
 COPY linux-runner base-runner.sh /
 
+ENV CROSS_TOOLCHAIN_PREFIX=
```
Emilgardis, line 29: ditto, do we need this?

Emilgardis (changes requested):

docker/Dockerfile.x86_64-unknown-linux-gnu.centos
Alexhuszagh: It ensures consistency in all cases, and makes sure that scripts with `set -u` won't fail. It might not be technically needed, but ensures we have the exact same behavior for all image types. The only somewhat difference is for `emscripten` images, but that's because we really can't have a viable prefix for them, so we just use `em`.

Emilgardis (approved):
bors r+

PR #1131, opened 2022-11-13
Fix progress clap parsing.

Alexhuszagh:


Alexhuszagh:
Fixes a bug where clap parsing of `--progress` values had a layer of indirection when downcasting, which meant the value parser wasn't working. It will be nice to have UI tests.

Emilgardis (approved):
😄 

bors r+

PR #1132, opened 2022-11-13
Add C++ exception support to FreeBSD targets.

Alexhuszagh:


Alexhuszagh:
So far I've only tested x86_64 so far: testing x86 right now.

Alexhuszagh:
bors try *freebsd

Alexhuszagh:
bors try-
bors try --target *freebsd

Alexhuszagh:
This passes the full test suite locally against the CMake re2 tests in #1112.

Alexhuszagh:
We can actually improve this IMO a lot, since we can just copy everything from `/lib` over (the entire contents are 10.6MB), which would guarantee improved future support, and then from there we can copy all the symlinks in `usr/lib`. Even though `libcrypto.so.11` exists in `/lib`, it won't conflict with the `openssl` packages which is installed in `/usr/local/lib`. And then do something like:

```bash
for lib in /lib/*.so.*; do
    base=$(basename "{lib}")
    link="${base}"
    # not strictly necessary since this will always work, but good fallback
    while [[ "${link}" == *.so.* ]]; then
        link="${link%.*}"
    done

    # just extra insurance that we won't overwrite it
    if [[ -n "${link}" ]] && [[ "${link}" != "${base}" ]]; then
        ln -s "${base}" "${destdir}/lib/${link}"
    fi
done
```

And then we have all our appropriate symlinks. We also need a few extra: specifically, `ln -s libthr.so.3 "${destdir}/lib/libpthread.so"`.

Alexhuszagh:
bors try --target *freebsd

This still passes the CMake tests.

Alexhuszagh:
Just a ping for a review on this.

Emilgardis (approved):

Alexhuszagh:
My finest creation yet
bors try --target *freebsd

And now our changelog check is failing :/
https://github.com/cross-rs/cross/actions/runs/3561136338/jobs/5981770295#step:4:1344

```
  Error: Unable to locate a common ancestor between main and 31e5947b48c69fc911e015ef27f0d527ed214e5c
  Error: Process completed with exit code 1.
```

Except `31e5947b48c69fc911e015ef27f0d527ed214e5c` isn't part of the commit tree.

Emilgardis:
https://github.com/cross-rs/cross/commit/31e5947b48c69fc911e015ef27f0d527ed214e5c

does exists though, I think the synchronize event misfired for some reason, triggered the ci again with a new event and it works now

docker/freebsd.sh
```diff
@@ -31,20 +31,73 @@ max_freebsd() {
...
+        # we need a timeout in case the server is down to avoid
+        # infinitely hanging. timeout error code is always 124
+        # these mirrors can be quite slow, so have a long timeout
+        timeout 20s curl --silent --list-only --location "${home}/${BSD_ARCH}/" >/dev/null
+        code=$?
+        if [[ "${code}" == 0 ]]; then
+            echo "${home}"
+            return 0
+        elif [[ "${code}" != 124 ]]; then
+            echo -e "\e[1;33mWarning:\e[0m Mirror ${home} does not seem to work." 1>&2
```
Emilgardis, line 84:
```suggestion
            echo -e "\e[1;33mwarning:\e[0m Mirror ${home} does not seem to work." 1>&2
```

docker/freebsd.sh
```diff
@@ -31,20 +31,73 @@ max_freebsd() {
...
+        fi
+    done
+    set -e
+
+    echo "Could not find a working FreeBSD mirror." 1>&2
```
Emilgardis, line 89:
```suggestion
    echo "error: Could not find a working FreeBSD mirror." 1>&2
```

docker/freebsd.sh
```diff
@@ -31,20 +31,73 @@ max_freebsd() {
...
+    local mirror="${1}"
+    local response=
+    local line=
+    local lines=
+    local releases=
+    local max_release=
+
+    response=$(curl --silent --list-only --location "${mirror}/${BSD_ARCH}/" | grep RELEASE)
+    if [[ "${response}" != *RELEASE* ]]; then
+        echo "Could not find a candidate release for FreeBSD ${BSD_MAJOR}." 1>&2
```
Emilgardis, line 43:
```suggestion
        echo "error: Could not find a candidate release for FreeBSD ${BSD_MAJOR}." 1>&2
```

Emilgardis (changes requested):

Alexhuszagh:
bors r=Emilgardis

Alexhuszagh:
It looks lijke we have a spurious error with the packages hosting site, but if this becomes an issue, I'll try to setup a mirror:

```
#20 4.262 + curl --retry 3 -sSfL https://pkg.freebsd.org/FreeBSD:12:amd64/quarterly/packagesite.txz -O
(6) Could not resolve host: pkg.freebsd.org
#20 25.32 curl: (6) Could not resolve host: pkg.freebsd.org
#20 37.34 curl: (6) Could not resolve host: pkg.freebsd.org
#20 51.38 curl: (6) Could not resolve host: pkg.freebsd.org
```

Issue #1124, opened 2022-11-07
Always Manually Link to Libc on Musl Targets


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar request
- [ ] This feature could be solved with a [custom docker image](https://github.com/cross-rs/cross#custom-docker-images) (optional)

### Describe your request

Currently, when building and linking to projects with more complex C/C++ dependencies, such as `re2`, the order of linking and the lack of a dynamic `libstdc++` prevents numerous symbols, such as `setlocale` and `__cxa_atexit` from being found. Ever since #1063, we prevent dynamic linking to `libstdc++` due to numerous issues it has in the generated binaries when linked to Rust code.

### Describe why this would be a good inclusion for `cross`

This improves C++ for most musl targets, which we should support out-of-the-box.

PR #1134, opened 2022-11-13
Use a linker script for libstdc++.

Alexhuszagh:
Use a linker script for `libstdc++.so` since we remove the dylib, to ensure a static build is always used to avoid segfaults at runtime. However, the archive can reference missing symbols present in `libc` or `libgcc`, and itself depends on symbols in `libm`. To ensure these libraries are properly linked when forcing a build against the static `libstdc++`, create a linker script for `libstdc++.so` that links to the static archive, and as needed, all dependencies.

Related to #1124.


Alexhuszagh:
bors try --target *musl*

Alexhuszagh:
So this is a much better solution than what I was doing before: to fix #902, we removed the shared libstdc++.so, however, this has issues since the static version reference symbols resolved in `libc`, `libgcc`, and itself depends on `libm`. To avoid any issues with detecting missing symbols in compiler-builtins, and to avoid using libgcc whenever it is not required, we only want to link to `libgcc` when needed, and linking another time to `libc` shouldn't be required. Finally, manually detecting the compiler-builtins path and linking to it uses that library for something it was never intended for: using it as `libgcc` so we don't have to link to `libgcc`.

The solution is actually quite simple: use a linker script (which I should have though of, a long time ago):

```ld
/* cross-rs linker script
 * this allows us to statically link libstdc++ to avoid segfaults
 * https://github.com/cross-rs/cross/issues/902
 */
GROUP ( libstdc++.a AS_NEEDED( -lgcc -lc -lm ) )
```

This way, we always link to our static C++ library, and then resolve the missing symbols if required.

docker/Dockerfile.aarch64-unknown-linux-musl
```diff
@@ -21,9 +21,7 @@ ENV CROSS_SYSROOT=/usr/local/aarch64-linux-musl
 COPY musl-symlink.sh /
 RUN /musl-symlink.sh $CROSS_SYSROOT aarch64
 
-COPY aarch64-linux-musl-gcc.sh /usr/bin/
-COPY rustc_info.sh /
-
+COPY COPY musl-gcc.sh /usr/bin/"$CROSS_TOOLCHAIN_PREFIX"gcc.sh
```
Alexhuszagh, line 24: Nice.

Alexhuszagh:
bors try --target *musl*

docker/musl-gcc.sh
```diff
@@ -1,20 +1,82 @@
...
 set -euo pipefail
 
 main() {
-    if [[ $# -eq 0 ]]; then
-        exec "${CROSS_TOOLCHAIN_PREFIX}"gcc "${@}"
+    local minor
+    local flags=()
+    local patched_minor="${CROSS_BUILTINS_PATCHED_MINOR_VERSION:-0}"
+
+    if [[ $# -eq 0 ]] || [[ "${patched_minor}" == 0 ]]; then
```
Alexhuszagh, line 30: We could probably just remove this wrapper script for just about every target now, actually.

Alexhuszagh:
bors try --target *musl*

Emilgardis (approved):
oops my bad!

yes this looks great

bors r+

PR #1135, opened 2022-11-14
CI: Fix most deprecation warnings from GitHub Actions

### LINKED ISSUE #1074: Migrate CI to use env files for save state
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

guihkx:
This fixes most deprecation warnings from GitHub Actions.

Changes:

- CI: Replace deprecated `::set-output` command
- CI: Replace deprecated `::save-state` command
- CI: Bump [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) from `v1` to `master`
- CI: Replace [`Ana06/get-changed-files`](https://github.com/Ana06/get-changed-files) by [`tj-actions/changed-files`](https://github.com/tj-actions/changed-files)

Fixes #1074

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -14,19 +14,19 @@ jobs:
 
       - name: Get Changed Files
         id: files
-        uses: Ana06/get-changed-files@v2.1.0
+        uses: tj-actions/changed-files@v34
         with:
-          # use JSON so we don't have to worry about filenames with spaces
-          format: 'json'
-          filter: '.changes/*.json'
+          separator: ';'
+          files: |
+            .changes/*.json
 
       - name: Validate Changelog Files
         id: changelog
         run: |
           set -x
           set -e
-          readarray -t added_modified < <(jq -r '.[]' <<<'${{ steps.files.outputs.added_modified }}')
-          readarray -t removed < <(jq -r '.[]' <<<'${{ steps.files.outputs.removed }}')
+          readarray -d ';' -t added_modified <<< '${{ steps.files.outputs.all_changed_files }}'
+          readarray -d ';' -t removed <<< '${{ steps.files.outputs.deleted_files }}'
           added_count=${#added_modified[@]}
           removed_count=${#removed[@]}
           if ${{ !contains(github.event.pull_request.labels.*.name, 'no changelog' ) }}; then
--- a/xtask/src/util.rs
+++ b/xtask/src/util.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
@@ -296,12 +297,13 @@ pub fn gha_error(content: &str) {
 }
 
 #[track_caller]
-pub fn gha_output(tag: &str, content: &str) {
+pub fn gha_output(tag: &str, content: &str) -> cross::Result<()> {
     if content.contains('\n') {
         // https://github.com/actions/toolkit/issues/403
-        panic!("output `{tag}` contains newlines, consider serializing with json and deserializing in gha with fromJSON()")
+        eyre::bail!("output `{tag}` contains newlines, consider serializing with json and deserializing in gha with fromJSON()");
     }
-    gha_output!("::set-output name={tag}::{}", content);
+    write_to_gha_env_file("GITHUB_OUTPUT", &format!("{tag}={content}"))?;
+    Ok(())
 }
 
 pub fn read_dockerfiles(msg_info: &mut MessageInfo) -> cross::Result<Vec<(PathBuf, String)>> {
@@ -408,6 +410,15 @@ pub fn write_to_string(path: &Path, contents: &str) -> cross::Result<()> {
         .truncate(true)
         .create(true)
         .open(path)?;
+    writeln!(file, "{}", contents)?;
+    Ok(())
+}
+
+// https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files
+pub fn write_to_gha_env_file(env_name: &str, contents: &str) -> cross::Result<()> {
+    let path = env::var(env_name)?;
+    let path = Path::new(&path);
+    let mut file = fs::OpenOptions::new().append(true).open(path)?;
     writeln!(file, "{}", contents)?;
     Ok(())
 }
--- a/.github/actions/cargo-install-upload-artifacts/action.yml
+++ b/.github/actions/cargo-install-upload-artifacts/action.yml
@@ -14,7 +14,7 @@ runs:
         metadata="$(cargo metadata --format-version 1 --no-deps)"
 
         package_name="cross"
-        echo "::set-output name=package-name::${package_name}"
+        echo "package-name=${package_name}" >> $GITHUB_OUTPUT
 
         out_dir="$(mktemp -d)"
         artifacts_dir="$(mktemp -d)"
@@ -24,8 +24,8 @@ runs:
           artifacts_dir="$(cygpath -w "${artifacts_dir}")"
         fi
 
-        echo "::set-output name=out-dir::${out_dir}"
-        echo "::set-output name=artifacts-dir::${artifacts_dir}"
+        echo "out-dir=${out_dir}" >> $GITHUB_OUTPUT
+        echo "artifacts-dir=${artifacts_dir}" >> $GITHUB_OUTPUT
       shell: bash
     - run: rm -rf .git
       shell: bash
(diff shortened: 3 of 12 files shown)


guihkx:
I'll add a changelog in my next rebase.

I'll also try to find a good replacement for the (apparently) unmaintained [Ana06/get-changed-files](https://github.com/Ana06/get-changed-files) action.

Alexhuszagh:
> I'll add a changelog in my next rebase.

The changes are internal so it's not required, since it's just a change to our Github actions.

> For the [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) action, I'll also have to use `@master` instead of `@v1`, just to get rid of other `::set-output` warnings. Is that okay with you guys?

I'm personally fine with that until a new release is out, especially since it's dtolnay.

> I'll also try to find a good replacement for the (apparently) unmaintained [Ana06/get-changed-files](https://github.com/Ana06/get-changed-files) action.

Thanks, if you can't find one, any contribution is more than enough. We might have to fork it, change the output variables, pin it, and hope it's minimal work to update again.

Alexhuszagh:
From the [action](https://github.com/cross-rs/cross/actions/runs/3461873354/jobs/5780026657#step:5:29)
> readarray -t added_modified < <(jq -r '.[]' <<<'[\".changes/1135.json\"]')
> parse error: Invalid numeric literal at line 1, column 3

Not sure why this is showing up now, I'm more than happy to add the `no changelog` label once this is merged because that shouldn't be happening.

.github/workflows/changelog.yml
```diff
@@ -14,19 +14,20 @@ jobs:
 
       - name: Get Changed Files
         id: files
-        uses: Ana06/get-changed-files@v2.1.0
+        uses: tj-actions/changed-files@v34
```
Alexhuszagh, line 17: Ah this might be the issue. It seems like the quoting in this isn't actually correct.

guihkx: Yeah, `tj-actions/changed-files` seems to escape  `"` by default when we use JSON output. And I didn't find an option to disable it. I could use `sed` to replace `\"` by `"`, but I IMO that's too cumbersome.

Anyway, I've experimented not using JSON output, and instead I'm using the `;` character as the file separator and it worked. I'll push the changes soon.

guihkx:
Okay, I think the changes look good now.

The 'changelog check' job is getting skipped now, but just to make sure it still works, I tested it on on my local repository and [it passed](https://github.com/guihkx/cross/actions/runs/3462754196/jobs/5782060251). ([test commit here](https://github.com/guihkx/cross/commit/0ac3c6f607a656fbbc26e8e47c8aed5e77190e83))

I also renamed the PR because I ended up changing more stuff other than just the `::set-output` warnings.

xtask/src/util.rs
```diff
@@ -301,7 +301,7 @@ pub fn gha_output(tag: &str, content: &str) {
         // https://github.com/actions/toolkit/issues/403
         panic!("output `{tag}` contains newlines, consider serializing with json and deserializing in gha with fromJSON()")
     }
-    gha_output!("::set-output name={tag}::{}", content);
+    gha_output!("tag={} >> $GITHUB_OUTPUT", content);
```
Emilgardis, line 304: this is not correct, `gha_output` is simply an alias to println/eprintln

Instead we need to write to the file pointed to from `GITHUB_OUTPUT`

Emilgardis (changes requested):

.github/workflows/changelog.yml
```diff
@@ -14,19 +14,19 @@ jobs:
...
+            .changes/*.json
 
       - name: Validate Changelog Files
         id: changelog
         run: |
           set -x
           set -e
-          readarray -t added_modified < <(jq -r '.[]' <<<'${{ steps.files.outputs.added_modified }}')
-          readarray -t removed < <(jq -r '.[]' <<<'${{ steps.files.outputs.removed }}')
+          readarray -d ';' -t added_modified <<< '${{ steps.files.outputs.all_changed_files }}'
```
Alexhuszagh, line 28: This change should be fine since we have a very limited subset of characters we accept in file names. Nice.

guihkx:
Okay, I tried to address @Emilgardis's comment to the best of my Rust knowledge.

Hope it isn't that bad... 😅

xtask/src/util.rs
```diff
@@ -296,12 +297,13 @@ pub fn gha_error(content: &str) {
 }
 
 #[track_caller]
-pub fn gha_output(tag: &str, content: &str) {
+pub fn gha_output(tag: &str, content: &str) -> cross::Result<()> {
     if content.contains('\n') {
         // https://github.com/actions/toolkit/issues/403
         panic!("output `{tag}` contains newlines, consider serializing with json and deserializing in gha with fromJSON()")
```
Alexhuszagh, line 303: If we're returning a result here, this probably can be an `eyre::bail!` case:

```rust
eyre::bail!("output `{tag}` contains newlines, consider serializing with json and deserializing in gha with fromJSON()");
```

Alexhuszagh:
Just that minor nit on the panic, since we return a result now. Otherwise, looks good to me. I'll leave the full review to Emil.

xtask/src/util.rs
```diff
@@ -411,3 +413,12 @@ pub fn write_to_string(path: &Path, contents: &str) -> cross::Result<()> {
     writeln!(file, "{}", contents)?;
     Ok(())
 }
+
+// https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files
+pub fn write_to_gha_env_file(env_name: &str, contents: &str) -> cross::Result<()> {
+    let path = env::var(env_name)?;
+    let path = Path::new(&path);
+    let mut file = fs::OpenOptions::new().write(true).open(path)?;
```
Emilgardis, line 421: this needs to append, not write over.

```suggestion
    let mut file = fs::OpenOptions::new().append(true).open(path)?;
```

Emilgardis (changes requested):
great! Just a minor nit and I think this will be ready.

guihkx:
I addressed the last two comments. Thanks a lot for the reviews, by the way. :)

Emilgardis:
bors try --target *unknown-linux*

guihkx:
Okay, we're down from 1000+ warnings to 478, which is still quite an expressive number.

I think I underestimated a bit the complexity of your CI... 😆 

However, it should be fairly easy to fix the remaining warnings as well.

Let me know if I should do it in this PR, or if I should open a new one.

Thanks!

Alexhuszagh (approved):
> Let me know if I should do it in this PR, or if I should open a new one.

Either is fine, this looks good to me as-is. I'll let Emil decide and merge, but everything looks sound. If you'd like to fix the rest in a subsequent PR, that would be appreciated as well.

Emilgardis:
I think we can fix the remaining warnings

need to update `docker/setup-buildx-action` to v2 and `docker/login-action` to v2 (unsure if `docker/login-action@v2` would solve it but I think it would)

Emilgardis:
@guihkx I added you as a member in bors so you can run `bors try`

guihkx:
>need to update `docker/setup-buildx-action` to v2 and `docker/login-action` to v2 (unsure if `docker/login-action@v2` would solve it but I think it would)

Yup, updating these two should get rid of most warnings.

There's also `mindsers/changelog-reader-action@v2.0.0`, which is displaying warnings about NodeJS 12, but this issue in specific has been fixed in [v2.2.1](https://github.com/mindsers/changelog-reader-action/releases/tag/v2.2.1). Although a couple of `set-output` warnings [still remain](https://github.com/mindsers/changelog-reader-action/issues/75). 

Nevertheless, I think it should be fairly harmless to just use `@v2` instead of a fixed version.

> @guihkx I added you as a member in bors so you can run `bors try`

Thank you. I'll take this opportunity to check if there are other actions that can be easily updated too.

guihkx:
bors try --target *unknown-linux*

guihkx:
Down to 7 warnings. Nice!

image

The first 2 warnings I'm not sure how to address right now. 🤔 

The next 4 warnings are related to usage of `set-output`, and as I mentioned earlier, come from [here](https://github.com/mindsers/changelog-reader-action/issues/75), and once upstream addresses them, they'll go away naturally since we've switched to `@v2`.

And the very last warning seems harmless to me. It's just GitHub informing that the `macos-latest` runner will soon use macOS 12 instead of 11.

Alexhuszagh:
Only the `set-output` ones are warnings worth dealing with: we use macOS latest intending to use 13 when available, and there's no real reason to check if `rustfmt` is already installed since attempting to install it does so anyway.

guihkx:
Nice! In that case, the PR is ready on my side.

Emilgardis (approved):
awesome! Thanks a lot!

bors r=Alexhuszagh,emilgardis

guihkx:
Ah, I see...

`readarray` trips when we pass an empty "here string" as argument. For example, this:

```bash
readarray -d ';' -t added_modified <<< ''
echo "array size: ${#added_modified[@]}"
```

Outputs:

```
array size: 1
```

My mistake. In this case, `readarray` isn't the right tool for the job. 

This is easily fixable, though. I'll send a PR soon.

PR #1139, opened 2022-11-16
CI: Fix changelog check

guihkx:
`readarray` is supposed to be used with actual files, so replace it with `read` and also configure bash's internal field separator (IFS) to use the `;` character as the delimiter.

This allows for the list of files to be properly split.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -25,8 +25,8 @@ jobs:
         run: |
           set -x
           set -e
-          readarray -d ';' -t added_modified <<< '${{ steps.files.outputs.all_changed_files }}'
-          readarray -d ';' -t removed <<< '${{ steps.files.outputs.deleted_files }}'
+          IFS=';' read -a added_modified <<< '${{ steps.files.outputs.all_changed_files }}'
+          IFS=';' read -a removed <<< '${{ steps.files.outputs.deleted_files }}'
           added_count=${#added_modified[@]}
           removed_count=${#removed[@]}
           if ${{ !contains(github.event.pull_request.labels.*.name, 'no changelog' ) }}; then


Alexhuszagh:
Looks good to me so far, as long as the input is how I expect it to be. I'll add the `no changelog` label shortly.

Can you do 3 pushes, just to confirm it works, since it's passing my local tests but I just want to ensure in the full workflow:
1. Add a `.changes/1139.json` (testing 1 file)
2. Add a `.changes/1138.json` and `.changes/1139.json` (testing 2 files)
3. Reverting to no changelogs once it all passes.

**1138.json**

```json
{
    "description": "description",
    "type": "fixed"
}
```

**1139.json**

```json
{
    "description": "description",
    "type": "added"
}
```

My local tests were the following, btw:

```bash
readarray_cmd() {
    readarray -d ';' -t var <<< "'${1}'"
    echo "${var[@]}"
    echo "${#var[@]}"
}

read_cmd() {
    IFS=';' read -a var <<< "${1}"
    echo "${var[@]}"
    echo "${#var[@]}"
}

$readarray_cmd ""
''

1
$read_cmd ""

0
$readarray_cmd ".changes/1139.json"
'.changes/1139.json'

1
$read_cmd ".changes/1139.json"
.changes/1139.json
1
$readarray_cmd ".changes/1138.json;.changes/1139.json"
'.changes/1138.json .changes/1139.json'

2
$read_cmd ".changes/1138.json;.changes/1139.json"
.changes/1138.json .changes/1139.json
2
```

guihkx:
Okay, that was with one file. I'll have to rebase and add two files simultaneously, correct? Because I think that if I push a commit that adds one more file right now, it will count as just one, right?

Alexhuszagh:
> Okay, that was with one file. I'll have to rebase and add two files simultaneously, correct? Because I think that if I push a commit that adds one more file right now, it will count as just one, right?

I think it's for the entire PR, so either a regular commit on top of it or a force push should work.

guihkx:
You were right, it counts for the entire PR. Can I revert the commits now?

Alexhuszagh (approved):
bors r+

PR #1149, opened 2022-11-22
Add a comment for how the Dockerfile path is resolved.

Alexhuszagh:


Alexhuszagh:
I'm not sure if this is actually required, but it's useful information on how the path to the Dockerfile is resolved. The context is already documented in `/src/docker/shared.rs`:

https://github.com/cross-rs/cross/blob/ab99baf7f4c39403a0b033cef95d798bd0b9a940/src/docker/shared.rs#L422-L427

src/docker/custom.rs
```diff
@@ -128,6 +128,10 @@ impl<'a> Dockerfile<'a> {
             }
         }
 
+        // note that this is always relative to the PWD: if we have `/Dockerfile`
+        // (relative to the workspace root), then running a build in `/src`
+        // requires the Dockerfile path to be specified as `../Dockerfile`,
+        // and running from `/..` requires `${PACKAGE_NAME}/Dockerfile`.
         docker_build.args(["--file".into(), path]);
```
Emilgardis, line 135: having issues parsing this, maybe

```suggestion
        // note that this is always relative to the PWD: if we have `$root/Dockerfile`
        // (relative to the workspace root), then running a build
        // `PWD=$root/src/ cross build`
        // would require the Dockerfile path to be specified as `../Dockerfile`.
        docker_build.args(["--file".into(), path]);
```

Alexhuszagh: If we use this with the `$root`, I think we can change it to `$workspace_root` and then remove `(relative to the workspace root)`.

Emilgardis (approved):
better than nothing

bors r+

PR #1151, opened 2022-11-22
Replace atty with is-terminal.

Alexhuszagh:
Although this doesn't affect cross (since we do not use custom allocators), this address [rustsec-1457](https://github.com/rustsec/advisory-db/issues/1457).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ dev = []
 members = ["xtask"]
 
 [dependencies]
-atty = "0.2"
+is-terminal = "0.4.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ dev = []
 members = ["xtask"]
 
 [dependencies]
-atty = "0.2"
+is-terminal = "0.4.0"
```
Alexhuszagh, line 29: We can probably change this to `0.4` since we have a lock the versions.

Emilgardis (approved):

Alexhuszagh:
bors r=Emilgardis

PR #1154, opened 2022-11-23
be more explicit about eol

Emilgardis:
related to #1152

### DIFF
--- a/docker/.gitattributes
+++ b/docker/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
-Dockerfile.* linguist-language=Dockerfile
+* text=auto
+
+Dockerfile.* linguist-language=Dockerfile eol=lf


docker/.gitattributes
```diff
@@ -0,0 +1,3 @@
+Dockerfile.* text=auto
```
Alexhuszagh, line 1: Should we also do this with bash scripts and source files? I think `*.py` (for the Android build system), `*.rs` (for our Rust source files), and `*.sh` for our bash scripts would be good for consistency.

Emilgardis: do you mean the `text=auto`? [this](https://www.git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto) is what it does

> When `text` is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. When the file has been committed with CRLF, no conversion is done.

Alexhuszagh (approved):

PR #1164, opened 2022-11-28
Add package site mirrors for FreeBSD.

### LINKED ISSUE #1162: Setup Fallback Mirrors for `pkg.freebsd.org`
In at least 3 separate workflow runs, we've had failures due to an inability to resolve `pkg.freebsd.org`. Normally, mirrors should be made available based on GeoDNS and [automatically handled](https://forums.freebsd.org/threads/how-to-change-mirror-in-pkg.79991/), however, it might be worth looking up a few and using logic similar in #1132 (specifically, commit aac375abdbcacbf473cad116f7e89c7a5d2396f2) to handle when the default gateway fails, or use other mirrors if possible. Less than ideal, but this might be necessary.

https://github.com/cross-rs/cross/actions/runs/3565770837/jobs/5991488382#step:10:28393
https://github.com/cross-rs/cross/actions/runs/3567054500/jobs/5994404235#step:10:28387
https://github.com/cross-rs/cross/actions/runs/3566358814/jobs/5992830213#step:10:28390

Alexhuszagh:
Closes #1162.

### DIFF
--- a/docker/freebsd-common.sh
+++ b/docker/freebsd-common.sh
@@ -6,57 +6,14 @@ set -euo pipefail
 # shellcheck disable=SC1091
 . freebsd-arch.sh
 
-export BSD_ARCH=
+export FREEBSD_ARCH=
 case "${ARCH}" in
     x86_64)
-        BSD_ARCH=amd64
+        FREEBSD_ARCH=amd64
         ;;
     i686)
-        BSD_ARCH=i386
+        FREEBSD_ARCH=i386
         ;;
 esac
 
-# we prefer those closer in geography to the US. they're triaged in
-# order of ease of use, reliability, and then geography. the mirror
-# list is at https://docs.freebsd.org/en/books/handbook/mirrors/.
-# these mirrors were known to work as of 2022-11-27. this does
-# not include any mirrors that are known to be rate-limited or
-# commercial.
-export BSD_HOME=(
-    # these do not return HTML, and only list the directories
-    "ftp.freebsd.org/pub/FreeBSD/releases"
-    # these return HTML output, and therefore are lower priority
-    "ftp11.freebsd.org/pub/FreeBSD/releases"
-    "ftp3.br.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.uk.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.nl.freebsd.org/pub/FreeBSD/releases"
-    "ftp6.fr.freebsd.org/pub/FreeBSD/releases"
-    "ftp1.de.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.de.freebsd.org/pub/FreeBSD/releases"
-    "ftp5.de.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.ru.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.gr.freebsd.org/pub/FreeBSD/releases"
-    "ftp4.za.freebsd.org/pub/FreeBSD/releases"
-    "ftp2.za.freebsd.org/pub/FreeBSD/releases"
-    "ftp4.tw.freebsd.org/pub/FreeBSD/releases"
-    "ftp3.jp.freebsd.org/pub/FreeBSD/releases"
-    "ftp6.jp.freebsd.org/pub/FreeBSD/releases"
-    # these only support HTTP, and not implicit
-    # FTP as well, and have HTML output
-    "http://ftp.uk.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.nl.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.fr.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.at.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.dk.freebsd.org/FreeBSD/releases"
-    "http://ftp.cz.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.se.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.lv.freebsd.org/freebsd/releases"
-    "http://ftp.pl.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.ua.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.gr.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.ru.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.nz.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.kr.freebsd.org/pub/FreeBSD/releases"
-    "http://ftp.jp.freebsd.org/pub/FreeBSD/releases"
-)
-export BSD_MAJOR=12
+export FREEBSD_MAJOR=12
--- a/docker/freebsd-extras.sh
+++ b/docker/freebsd-extras.sh
@@ -11,8 +11,16 @@ set -euo pipefail
 . freebsd-install.sh
 
 main() {
-    setup_packagesite
-    install_freebsd_package openssl sqlite3
+    apt-get update && apt-get install --assume-yes --no-install-recommends \
+        curl \
+        dnsutils \
+        jq \
+        xz-utils
+
+    local url=
+    url=$(fetch_best_freebsd_mirror)
+    FREEBSD_MIRROR="${url}" setup_freebsd_packagesite
+    FREEBSD_MIRROR="${url}" install_freebsd_package openssl sqlite3
 
     rm "${0}"
 }
--- a/docker/freebsd-install.sh
+++ b/docker/freebsd-install.sh
@@ -51,7 +51,7 @@ freebsd_mirror_works() {
     timeout 20s curl --retry 3 -sSfL "${path}" >/dev/null 2>&1
 }
 
-fetch_best_freebsd_mirror() {
+_fetch_best_freebsd_mirror() {
     # in case if the default mirror is down, we can use various known
     # fallbacks, or at worst, SRV fallbacks to find the ideal package
     # site. no individual mirror other than the default mirror is
@@ -100,6 +100,15 @@ fetch_best_freebsd_mirror() {
 
     echo -e "\e[31merror:\e[0m could not find a working FreeBSD package mirror." 1>&2
     exit 1
+}
+
+fetch_best_freebsd_mirror() {
+    set +e
+    _fetch_best_freebsd_mirror
+    code=$?
+    set -e
+
+    return "${code}"
 }
 
 setup_freebsd_packagesite() {
--- a/docker/freebsd-install.sh
+++ b/docker/freebsd-install.sh
@@ -6,35 +6,138 @@ set -euo pipefail
 # shellcheck disable=SC1091
 . freebsd-common.sh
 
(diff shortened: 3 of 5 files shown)


docker/freebsd-common.sh
```diff
@@ -6,57 +6,14 @@ set -euo pipefail
...
         ;;
 esac
 
-# we prefer those closer in geography to the US. they're triaged in
-# order of ease of use, reliability, and then geography. the mirror
-# list is at https://docs.freebsd.org/en/books/handbook/mirrors/.
-# these mirrors were known to work as of 2022-11-27. this does
-# not include any mirrors that are known to be rate-limited or
-# commercial.
-export BSD_HOME=(
```
Alexhuszagh, line 25: Since this is only used in `freebsd.sh`, and these mirrors aren't the same as the packagesite mirrors, I moved them to `freebsd,sh`.

Alexhuszagh:
bors try --target *freebsd

docker/freebsd-extras.sh
```diff
@@ -11,8 +11,16 @@ set -euo pipefail
 . freebsd-install.sh
 
 main() {
-    setup_packagesite
-    install_freebsd_package openssl sqlite3
+    apt-get update && apt-get install --assume-yes --no-install-recommends \
```
Alexhuszagh, line 14: These should be installed by default, and we need `dnsutils` for the `dig` command. Note that the `dig` output is only machine parseable without a real parser using the `+short` flag:

```bash
dig +short $url
```

docker/freebsd-extras.sh
```diff
@@ -11,8 +11,16 @@ set -euo pipefail
...
+        curl \
+        dnsutils \
+        jq \
+        xz-utils
+
+    local url=
+    url=$(fetch_best_freebsd_mirror)
+    FREEBSD_MIRROR="${url}" setup_freebsd_packagesite
```
Alexhuszagh, line 22: Make the mirrors optional, so the commands can be used by people building their own custom Docker images, without needing to hard-code the mirror or doing redundant work.

docker/freebsd-install.sh
```diff
@@ -6,35 +6,138 @@ set -euo pipefail
 # shellcheck disable=SC1091
 . freebsd-common.sh
 
-export PACKAGESITE=/opt/freebsd-packagesite/packagesite.yaml
-export PKG_SOURCE="https://pkg.freebsd.org/FreeBSD:${BSD_MAJOR}:${BSD_ARCH}/quarterly"
-export TARGET="${ARCH}-unknown-freebsd${BSD_MAJOR}"
+# list of SRV records to query if the default mirror fails
```
Alexhuszagh, line 9: Everything has been prefixed with `freebsd` since these could be included in user scripts.

docker/freebsd-install.sh
```diff
@@ -6,35 +6,138 @@ set -euo pipefail
...
+    "_http._tcp.pkg.all.freebsd.org"
+    # this only returns geodns mirrors
+    "_http._tcp.pkg.freebsd.org"
+)
+FREEBSD_PACKAGEDIR="/opt/freebsd-packagesite"
+FREEBSD_PACKAGESITE="${FREEBSD_PACKAGEDIR}/packagesite.yaml"
+FREEBSD_TARGET="${ARCH}-unknown-freebsd${FREEBSD_MAJOR}"
+FREEBSD_DEFAULT_MIRROR="pkg.freebsd.org"
+# NOTE: these mirrors were known to work as of 2022-11-28.
+# no availability guarantees are made for any of them.
```
Alexhuszagh, line 21: Pain.

docker/freebsd-install.sh
```diff
@@ -6,35 +6,138 @@ set -euo pipefail
...
+    # site. no individual mirror other than the default mirror is
+    # guaranteed to exist, so we use a tiered approach. only
+    # the default mirror supports https.
+    if freebsd_mirror_works "${FREEBSD_DEFAULT_MIRROR}" "https"; then
+        echo "https://${FREEBSD_DEFAULT_MIRROR}"
+        return 0
+    fi
+
+    # if we've gotten here, it could be a DNS issue, so using a DNS
+    # resolver to fetch SRV fallbacks may not work. let's first try
```
Alexhuszagh, line 66: Pain.

docker/freebsd-install.sh
```diff
@@ -6,35 +6,138 @@ set -euo pipefail
...
+            echo "http://${mirror}"
+            return 0
+        fi
+    done
+
+    local http_tcp_source=
+    local response=
+    local lines=
+    # shellcheck disable=SC2016
+    local regex='/[._A-Za-z0-9-]+ has SRV record \d+\s+\d+\s+\d+\s+(.*)\./; print $1'
```
Alexhuszagh, line 80: This needs to be modified when we use `dig`: `nslookup` and `host` (BIND9) are [deprecated](https://salsa.debian.org/dns-team/bind9/-/commit/336ed6adb20e1c9fd6f082a428361fd736171a52).

docker/freebsd-install.sh
```diff
@@ -6,35 +6,138 @@ set -euo pipefail
...
+    # shellcheck disable=SC2016
+    local regex='/\d+\s+\d+\s+\d+\s+(.*)\./; print $1'
+    for http_tcp_source in "${FREEBSD_HTTP_TCP_SOURCES[@]}"; do
+        # the output will have the following format, but we only want the
+        # target and ignore everything else:
+        #   $priority $port $weight $target.
+        #
+        # some output may not match, so we skip those lines, for example:
+        #   96.47.72.71
+        response=$(dig +short srv "${http_tcp_source}")
```
Alexhuszagh, line 88: I recommend testing this output manually just to ensure that the actual mirror list works locally.

For example, I have:

```bash
FREEBSD_ARCH=amd64
FREEBSD_MAJOR=12
FREEBSD_HTTP_TCP_SOURCES=(
    "_http._tcp.pkg.all.freebsd.org"
    "_http._tcp.pkg.freebsd.org"
)

freebsd_package_source() {
    local url="${1}"
    echo "${url}/FreeBSD:${FREEBSD_MAJOR}:${FREEBSD_ARCH}/quarterly"
}

freebsd_mirror_works() {
    local mirror="${1}"
    local scheme="${2}"
    local pkg_source=

    # meta.conf is a small file for quick confirmation the mirror works
    pkg_source=$(freebsd_package_source "${scheme}://${mirror}")
    local path="${pkg_source}/meta.conf"

    timeout 20s curl --retry 3 -sSfL "${path}" >/dev/null 2>&1
}

regex='/\d+\s+\d+\s+\d+\s+(.*)\./; print $1'
for http_tcp_source in "${FREEBSD_HTTP_TCP_SOURCES[@]}"; do
    response=$(dig +short srv "${http_tcp_source}")
    readarray -t lines <<< "${response}"
    for line in "${lines[@]}"; do
        mirror=$(echo "${line}" | perl -nle "${regex}")
        if [[ -n "${mirror}" ]]; then
            echo "${mirror}"
        fi
    done
done
```

And then try:

```bash
mirror="pkg0.bra.freebsd.org"
if freebsd_mirror_works "${mirror}" "http"; then
    echo "http://${mirror}"
fi
```

This shows that the regex works as does the check for the mirrors working.

Alexhuszagh:
We might also want to add a retry flag here:
https://github.com/cross-rs/cross/actions/runs/3570698725/jobs/6001977624#step:10:270

```bash
curl --silent --list-only --location ftp.freebsd.org/pub/FreeBSD/releases/i386/
```

docker/freebsd.sh
```diff
@@ -38,9 +82,9 @@ latest_freebsd() {
     local releases=
     local max_release=
 
-    response=$(curl --silent --list-only --location "${mirror}/${BSD_ARCH}/" | grep RELEASE)
+    response=$(curl --retry 3 -sSflL "${mirror}/${FREEBSD_ARCH}/" | grep RELEASE)
```
Alexhuszagh, line 85: Added the retry options to avoid the failures if the mirror is found but the subsequent attempt to query the server fails, like in our [workflows](https://github.com/cross-rs/cross/actions/runs/3570698725/jobs/6001977624#step:10:270).

Alexhuszagh:
Some of these need `set +e` and `set -e` wrappers for them, specifically for the mirror checking.

docker/freebsd.sh
```diff
@@ -67,21 +111,21 @@ latest_freebsd() {
...
     local code=
 
     set +e
-    for home in "${BSD_HOME[@]}"; do
+    for mirror in "${MIRRORS[@]}"; do
         # we need a timeout in case the server is down to avoid
         # infinitely hanging. timeout error code is always 124
         # these mirrors can be quite slow, so have a long timeout
-        timeout 20s curl --silent --list-only --location "${home}/${BSD_ARCH}/" >/dev/null
+        timeout 20s curl --retry 3 -sSflL "${mirror}/${FREEBSD_ARCH}/" >/dev/null
```
Alexhuszagh, line 122: So as of very recently, this no longer works unless we add `http` to the front:

```bash
$ mirror="ftp.freebsd.org/pub/FreeBSD/releases"
$ url="${mirror}/amd64"
$ curl --silent --list-only --location "${url}"
CI-IMAGES
ISO-IMAGES
README.TXT
TIMESTAMP
VM-IMAGES
amd64
arm
arm64
i386
powerpc
riscv
sparc64
```

This is because the "file" no longer exists, which `--list-only` is masking. We should *always** use the HTTP gateway for this (and probably everything else).

docker/freebsd-install.sh
```diff
@@ -6,35 +6,147 @@ set -euo pipefail
...
+            fi
+        done
+    done
+
+    echo -e "\e[31merror:\e[0m could not find a working FreeBSD package mirror." 1>&2
+    exit 1
+}
+
+fetch_best_freebsd_mirror() {
+    set +e
```
Alexhuszagh, line 106: I've added wrappers for the functions since the flow control inside has multiple exit points, just to ensure code downstream exits on error.

docker/freebsd.sh
```diff
@@ -66,43 +107,45 @@ latest_freebsd() {
...
         fi
     done
-    set -e
 
     echo -e "\e[31merror:\e[0m could not find a working FreeBSD mirror." 1>&2
     exit 1
 }
 
+freebsd_mirror() {
+    set +e
```
Alexhuszagh, line 133: Same as above.

docker/freebsd.sh
```diff
@@ -8,6 +8,47 @@ set -euo pipefail
...
 
+# we prefer those closer in geography to the US. they're triaged in
+# order of ease of use, reliability, and then geography. the mirror
+# list is at https://docs.freebsd.org/en/books/handbook/mirrors/.
+# these mirrors were known to work as of 2022-11-27. this does
+# not include any mirrors that are known to be rate-limited or
+# commercial. everything returns HTML output.
+MIRRORS=(
+    # this is a guaranteed mirror, unlike those below.
+    "http://ftp.freebsd.org/pub/FreeBSD/releases"
```
Alexhuszagh, line 19: The main mirror got changed `curl --list-only ftp.freebsd.org/pub/FreeBSD/releases/amd64` no longer lists the releases, so we have to use our HTML regex for everything.

Alexhuszagh:
This should be ready.
bors try --target *freebsd

Emilgardis (approved):
lgtm

the changelog entry should also link to this pr

r=me when file has been renamed `1163-1164.json`

Alexhuszagh:
bors r=Emilgardis

PR #1166, opened 2022-12-04
freebsd: include memstat in build image

taiki-e:
libc 0.2.138 added memstat_* functions, but the build with cross fails because the build image does not contain the corresponding library. See https://github.com/rust-lang/libc/issues/3025 for more context. See also https://github.com/rust-lang/rust/pull/105222 which is the same fix to rust-lang/rust's build image.

### DIFF
--- a/.changes/1166.json
+++ b/.changes/1166.json
@@ -1,0 +1,4 @@
+{
+    "description": "freebsd: include memstat in build image to fix build with libc 0.2.138 and up.",
+    "type": "fixed"
+}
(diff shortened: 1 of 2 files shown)


Emilgardis (approved):
awesome!

Do you think this needs a changelog entry? I'm fine without one

taiki-e:
Added a changelog entry for this PR.

Emilgardis (approved):
bors r+

thanks!

PR #1171, opened 2022-12-08
Bugfix: custom image tag can't end with neither a `-` nor `.`

growse:
The original bug fix solved the problem of a image name ending in a single `-` character (which docker doesn't permit). However, docker also doesn't permit names ending in a `.`, so we need to handle that case too.

Also, removing just the last char is insufficient, as names may end up ending with multiple consecutive `-` or `.` characters, so we need to loop until they're all gone.

### DIFF
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -247,11 +247,12 @@ fn docker_tag_name(file_name: &str) -> String {
         }
     }
 
+    // in case our result ends in an invalid last char `-` or `.`
+    // we remove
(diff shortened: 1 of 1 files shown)


Alexhuszagh (changes requested):
Very minor change for idiomatic Rust expressions, otherwise, looks good to me.

Alexhuszagh:
lgtm, just squash the commit history and I'll approve and merge it.

Emilgardis (approved):
Thanks!

bors r=Alexhuszagh,emilgardis

Emilgardis:
bors retry

PR #1178, opened 2022-12-27
fix owo-colors dep

Emilgardis:
doesn't bump dependency


Emilgardis:
going to wait for action in https://github.com/jam1garner/owo-colors/issues/82

PR #1193, opened 2023-01-21
suggest reinstalling the toolchain

### LINKED ISSUE #1192: Docs request: add info about x86_64-unknown-linux-gnu toolchain to "Getting Started"
I ran into the issue at this discussion: https://github.com/cross-rs/cross/discussions/1161 and it was pretty tough to find the solution. Would it be possible to document this issue in the getting started guide?

The message is just pretty confusing, since my target isn't even `x86_64-unknown-linux-gnu`. I assume it needs that installed to build in Docker, so isn't this a problem for all non-linux users?

Emilgardis:
also fixes suggestion when using wrong channel

resolves #1192

### DIFF
--- a/src/rustup.rs
+++ b/src/rustup.rs
@@ -1,8 +1,6 @@
 use std::path::PathBuf;
 use std::process::Command;
 
-use color_eyre::owo_colors::OwoColorize;
-use color_eyre::SectionExt;
 use rustc_version::{Channel, Version};
 
 use crate::errors::*;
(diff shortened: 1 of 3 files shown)


src/rustup.rs
```diff
@@ -93,14 +91,22 @@ pub fn available_targets(
...
+            .to_section_report();
+        if String::from_utf8_lossy(&output.stderr).contains("is a custom toolchain") {
+            err = err.wrap_err("'{toolchain}' is a custom toolchain.")
+            .suggestion(r#"To use this toolchain with cross, you'll need to set the environment variable `CROSS_CUSTOM_TOOLCHAIN=1`
+cross will not attempt to configure the toolchain further so that it can run your binary."#);
+        } else if String::from_utf8_lossy(&output.stderr).contains("does not support components") {
+            err = err.suggestion(format!(
+                "try reinstalling the '{toolchain}' toolchain
+$ rustup toolchain uninstall stable-x86_64-unknown-linux-gnu
+$ rustup toolchain install stable-x86_64-unknown-linux-gnu --force-non-host"
```
Emilgardis, line 106:
```suggestion
$ rustup toolchain uninstall {toolchain}
$ rustup toolchain install {toolchain} --force-non-host"
```

Emilgardis (commented):
bors r+

PR #1216, opened 2023-03-08
refactor into functions

### LINKED ISSUE #1215: cross install seems to be broken
### Checklist

### Describe your issue

`cross build` works like expected. But using `cross install` results in:

```
[cross] warning: unable to get metadata for package
    Updating crates.io index
  Installing trunk v0.16.0
   Compiling libc v0.2.139
[…]
  --- stderr


  error occurred: Failed to find tool. Is `aarch64-linux-gnu-gcc` installed?


error: failed to compile `trunk v0.16.0`, intermediate artifacts can be found at `/tmp/cargo-installQKUg1O`
```

It looks like as it `install` just isn't supported, so it falls back to running "plain cargo", which of course fails.

However, running `cross --help` shows `install` too.

other (specify in description)

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [X] podman

### cross version

0.2.5

### Example

```shell
cross install trunk --target aarch64-unknown-linux-gnu
```

_No response_

Emilgardis:
resolves #1215


src/bin/commands/install.rs
```diff
@@ -0,0 +1,146 @@
...
+            uses_build_std,
+        );
+
+        let cwd = std::env::current_dir()?;
+
+        let paths = docker::DockerPaths::create(
+            &engine,
+            cross::CargoMetadata {
+                workspace_root: cwd.clone(),
+                target_directory: cross::file::absolute_path(self.root)?,
```
Emilgardis, line 112: this is misleading

`cargo install` creates a temporary directory where it builds, we don't use this path as the normal target-dir, rather as the target directory for the `cargo install` output

src/docker/shared.rs
```diff
@@ -1034,8 +1036,11 @@ impl DockerCommandExt for Command {
...
             ])
-            .args(["-e", "CARGO_TARGET_DIR=/target"])
             .args(["-e", &cross_runner]);
+
+        if !options.skip_target_dir {
+            self.args(["-e", "CARGO_TARGET_DIR=/target"]);
+        }
         if options.cargo_variant.uses_zig() {
             // otherwise, zig has a permission error trying to create the cache
             self.args(["-e", "XDG_CACHE_HOME=/target/.zig-cache"]);
```
Emilgardis, line 1046: is this comment about zig a problem?

src/bin/commands/install.rs
```diff
@@ -0,0 +1,146 @@
...
+        }) = cross::setup(&host_version_meta, None, &args, target_list, msg_info)? else {
+            eyre::bail!("couldn't setup context for cross (see warning)")
+        };
+
+        let mut is_nightly = toolchain.channel.contains("nightly");
+        let mut rustc_version = None;
+        if let Some((version, channel, _)) = toolchain.rustc_version()? {
+            is_nightly = channel == rustc_version::Channel::Nightly;
+            rustc_version = Some(version);
+        }
```
Emilgardis, line 82: this doesn't emit warning on mismatch like normal cross does

Emilgardis (commented):
This does work, but I'm not happy with the code-reuse.

I want to make `Install::run` not be a rewrite of `cross::run`

src/bin/commands/install.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub color: Option<String>,
+    /// Container engine (such as docker or podman).
+    #[clap(long)]
+    pub engine: Option<String>,
+    /// Path to crate
+    #[clap(long)]
+    pub path: Option<String>,
+    /// Path to Cross.toml
+    #[clap(long)]
+    pub config: Option<std::path::PathBuf>,
```
Emilgardis, line 28: this is not implemented

PR #1231, opened 2023-03-25
ensure sysroot is correct

Emilgardis:


src/rustup.rs
```diff
@@ -83,18 +84,82 @@ pub fn active_toolchain(msg_info: &mut MessageInfo) -> Result<String> {
...
+                if mode.is_overriden() {
+                    mode = ToolchainMode::DefaultOverride;
+                } else {
+                    mode = ToolchainMode::Default;
+                }
+                l = l.replace(" (default)", "");
+            }
+
+            (l, mode)
+        })
```
Emilgardis, line 151: I'm not super happy about this.

src/rustup.rs
```diff
@@ -83,18 +84,82 @@ pub fn active_toolchain(msg_info: &mut MessageInfo) -> Result<String> {
...
+    /// [`DefaultOverride`]: ToolchainMode::DefaultOverride
+    #[must_use]
+    pub fn is_defaulted(&self) -> bool {
+        matches!(self, Self::Default)
+    }
+}
+
+pub fn installed_toolchains(
+    msg_info: &mut MessageInfo,
+) -> Result<Vec<(String, ToolchainMode, std::path::PathBuf)>> {
```
Emilgardis, line 126: should probably return a vec of `struct`, like `available_targets`

src/rustc.rs
```diff
@@ -364,15 +395,6 @@ pub fn target_list(msg_info: &mut MessageInfo) -> Result<TargetList> {
...
 
-pub fn sysroot(msg_info: &mut MessageInfo) -> Result<PathBuf> {
-    let stdout = rustc_command()
-        .args(["--print", "sysroot"])
-        .run_and_get_stdout(msg_info)?
-        .trim()
-        .to_owned();
-    Ok(PathBuf::from(stdout))
-}
-
```
Emilgardis, line 375: this is not needed anymore, `rustup toolchain list -v` gives us this, for all toolchains

src/rustc.rs
```diff
@@ -196,19 +196,30 @@ impl QualifiedToolchain {
...
+        {
+            (toolchain_name, sysroot)
+        } else if let Some((toolchain_name, _, sysroot)) = installed_toolchains
+            .iter()
+            .find(|(_, mode, _)| mode.is_defaulted())
+        {
+            (toolchain_name, sysroot)
+        } else {
+            eyre::bail!("no default toolchain found");
+        };
```
Emilgardis, line 217: I feel like there is a better way to write this, not sure how though

PR #1240, opened 2023-04-03
output configured targets

Emilgardis:


Emilgardis:
@mcandre does this work for you? I can add more data to the array

src/bin/commands/list_targets.rs
```diff
@@ -0,0 +1,131 @@
...
+        #[derive(serde::Serialize)]
+        struct Target {
+            triplet: String,
+            platforms: Vec<String>,
+        }
+        let mut targets: Vec<_> = cross::docker::PROVIDED_IMAGES
+            .iter()
+            .filter_map(|i| {
+                Some(Target {
+                    triplet: Some(i.name).filter(|i| *i != "zig")?.to_owned(),
```
Emilgardis, line 101: this is a nice usage of try trait :D

src/bin/commands/list_targets.rs
```diff
@@ -0,0 +1,131 @@
...
+                        .image
+                        .as_ref()
+                        .map(|i| {
+                            i.toolchain
+                                .iter()
+                                .map(ToString::to_string)
+                                .collect::<Vec<_>>()
+                        })
+                        .filter(|v| !v.is_empty())
+                        .unwrap_or_else(|| vec![docker::ImagePlatform::DEFAULT.to_string()]),
```
Emilgardis, line 123: this is meant to replicate what happens when we do https://github.com/Emilgardis/cross/blob/list-supported-targets/src/docker/image.rs#L30

src/bin/commands/list_targets.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use clap::Args;
+use cross::docker;
+use cross::shell::MessageInfo;
+
+#[derive(Args, Debug)]
+pub struct ListTargets {
+    /// Format version
+    #[clap(long)]
+    format_version: Option<FormatVersion>,
```
Emilgardis, line 11: do we really need this kind of stability?

mcandre:
> @mcandre does this work for you? I can add more data to the array

For my purposes, I would like to see the support tier number included in the output. Perhaps a field like `"support_tier": 3` as a sister field beside the `"triplet"` fields.

If possible, I would prefer a flatter array design, that doesn't restate some of the triple and platform information. I just want to be able to associate triple with support tier number.

If we go with JSON, then I would like to see *all* cross data models exported as a stable Rust library. The current policy treats only the CLI as stable. Until the data models and schemas (e.g., the entire Cross.toml schema) enters a stable library, I would actually prefer to see non-JSON in cross CLI tool output.

That is, `docker ps -a`-style text tables would be fine. These are easier to parse through via simple regex captures, than a long series of manual JSON value casting and checking snippets.

Emilgardis:
> For my purposes, I would like to see the support tier number included in the output. Perhaps a field like "support_tier": 3 as a sister field beside the "triplet" fields.

The tier is not relevant to cross, all targets emitted are the targets that actually has been setup for the project. i.e not every tier 1 and tier 2 target are emitted, only targets in any tier, including custom targets when specified in a projects config, are emitted **iff** they are supported by cross or have been setup to work with cross in the project which `cross-util` has been executed from.

i.e

```toml
# Cross.toml
[target.my-custom-target]
image = "my-image:latest" 
```

would output with `cross-util list-targets --format-version=1` that the target `my-custom-target` is a target.

> If we go with JSON, then I would like to see all cross data models exported as a stable Rust library. The current policy treats only the CLI as stable. Until the data models and schemas (e.g., the entire Cross.toml schema) enter a stable library, I would actually prefer to see non-JSON in cross CLI tool output.

That's the entire reason behind `--format-version 1`, we are guaranteeing that if you use `--format-version 1` you'll get the same object/schema.

PR #1248, opened 2023-04-30
Customize `--cache-to`, `--cache-from` options for `build-docker-image`

strowk:
Additional Dockerfiles for some of targets (f.e MacOS) require building docker images from [cross-toolchains](https://github.com/cross-rs/cross-toolchains) repositoriy, which is supported by `build-docker-image` command.

When running build-docker-image on CI environments, such as Github Actions, the process could be made greatly faster if cache from previous build is stored. The official support of buildx system includes f.e. possibility to use Github Actions Cache directly - https://docs.docker.com/build/cache/backends/gha/ , or for other CI systems it is also possible to use local cache in folder - https://docs.docker.com/build/cache/backends/local/ . However, such configurations require passing `--cache-to` and `--cache-from` options to docker build command, which is currently not customizeable by `build-docker-image` command supplied by `cross`.

This PR intends to fix this issue by introducing two new options for  `build-docker-image`, namely  `--cache-to` and `--cache-from`. Option names match those of docker CLI and only used unless other configurations do not contradict these new options, f.e. when push is specified to true, cache-to would still be using inline cache, as it used to before these changes. Same goes for `--cache-from`, which will not be used in case if no_cache is set to true or if engine does not support cache from type.

### DIFF
--- a/.changes/1248.json
+++ b/.changes/1248.json
@@ -1,4 +1,4 @@
 {
     "description": "customize `--cache-from` and `--cache-to` options for `build-docker-image`",
-    "type": "added"
+    "type": "internal"
 }
--- a/.changes/1248.json
+++ b/.changes/1248.json
@@ -1,0 +1,4 @@
+{
+    "description": "customize `--cache-from` and `--cache-to` options for `build-docker-image`",
+    "type": "added"
+}
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -63,7 +63,7 @@ pub struct BuildDockerImage {
     /// Option `--cache-to` for docker, only would work if push is not set to true
     ///
     /// Additionally you can use {base_name} to replace with base name of the image
-    /// If not specified, would not be passed to docker
+    /// If not specified, would not be passed to docker unless `--push` is used
     #[clap(long)]
     pub cache_to: Option<String>,
     /// Option `--cache-from` for docker, would only work if engine supports cache from type and no_cache is not set to true
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -277,10 +277,8 @@ pub fn build_docker_image(
 
         if push {
             docker_build.args(["--cache-to", "type=inline"]);
-        } else {
-            if let Some(ref cache_to) = cache_to {
-                docker_build.args(["--cache-to", cache_to]);
-            }
+        } else if let Some(ref cache_to) = cache_to {
+            docker_build.args(["--cache-to", cache_to]);
         }
 
         for tag in &tags {
--- a/xtask/src/build_docker_image.rs
+++ b/xtask/src/build_docker_image.rs
@@ -60,6 +60,12 @@ pub struct BuildDockerImage {
     /// Do not load from cache when building the image.
     #[clap(long)]
     pub no_cache: bool,
+    /// Cache to option for docker, only would work if push is not set to true
+    #[clap(long)]
+    pub cache_to: Option<String>,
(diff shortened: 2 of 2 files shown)


xtask/src/build_docker_image.rs
```diff
@@ -248,10 +256,14 @@ pub fn build_docker_image(
...
-            docker_build.args([
-                "--cache-from",
-                &format!("type=registry,ref={base_name}:main"),
-            ]);
+            if let Some(ref cache_from) = cache_from {
+                docker_build.args(["--cache-from", cache_from]);
+            } else {
+                docker_build.args([
+                    "--cache-from",
+                    &format!("type=registry,ref={base_name}:main"),
```
Emilgardis, line 264: would be nice if we could have some string replacements here so that it's possible to express the same thing the `format!` does here

that way we could also make the arguments have defaults, so something like

```rs
    /// Cache from option for docker, would only work if engine supports cache from type and no_cache is not set to true
    #[clap(long, default_value = "type=registry,ref={base_name}:main")]
    pub cache_from: String,
```

Emilgardis (approved):
looks great, can merge as is!

One small suggestion to make it more expressive, what do you think?

xtask/src/build_docker_image.rs
strowk: mhmm, problem is, format! won't accept anything that is not string literal - it is a macros and it wants to see the actual string in compile time. Assuming I understood correctly what you are asking, I have a feeling that `format!` will not allow it

strowk: I guess I can do something like `&cache_from.replace("{base_name}", &base_name)`

strowk: Though, I am not sure if in this case cache-to should also support such "templated" input or not

strowk: Ok, I have unified how --cache-from is passed - now instead of format!, which would not allow dynamic template, I am simply using `replace` and cache-from comes with default that maches usual behavior. From another side, cache-to remained optional, however when specified, same replacement would take place (not sure if necessary or not TBH, but just to make behavior consistent for two options). 

This is commit: 
https://github.com/cross-rs/cross/pull/1248/commits/0a157e58188ceec08d4f4f6c5ff3627f3fb1e292

.changes/1248.json
```diff
@@ -0,0 +1,4 @@
+{
+    "description": "customize `--cache-from` and `--cache-to` options for `build-docker-image`",
+    "type": "added"
```
Emilgardis, line 3:
```suggestion
    "type": "internal"
```

xtask/src/build_docker_image.rs
```diff
@@ -60,6 +60,17 @@ pub struct BuildDockerImage {
     /// Do not load from cache when building the image.
     #[clap(long)]
     pub no_cache: bool,
+    /// Option `--cache-to` for docker, only would work if push is not set to true
+    ///
+    /// Additionally you can use {base_name} to replace with base name of the image
+    /// If not specified, would not be passed to docker
```
Emilgardis, line 66:
```suggestion
    /// If not specified, would not be passed to docker unless `--push` is used
```

Emilgardis (approved):
Thanks!

Missed an issue with the changelog entry and a minor nit on documentation, otherwise this is ready. Thank you!

bors d+

strowk:
Applied suggestions

bors r+

Emilgardis:
bors r+ thanks again!

the failures are due to something else

Emilgardis:
spurious network error, lets try again

bors retry

strowk:
Thanks a lot for quick response and help with the bot (wouldn't know what to touch and where)!

Now I can switch from using my own repo to the upstream, great :)

Issue #680, opened 2022-04-05
glibc compatibility / make ` x86_64-unknown-linux-gnu` ubuntu again and provide image for glibc compatibility


Emilgardis:
I think it's unexpected all linux images are ubuntu, except `x86_64-unknown-linux-gnu`, I think this can be remedied by building two images: `x86_64-unknown-linux-gnu` and ` x86_64-unknown-linux-gnu.2.17`

Thankfully, we have not published the centos change yet, so this can be remedied

Related to the glibc issue, I think we can remedy this by suggesting users use `zig cc` ala [cargo-zigbuild](https://github.com/messense/cargo-zigbuild). We can even use a similar approach setting `*_CC` to `zig cc --target x86_64-linux-gnu.2.17` and make it easier for users of the `x86_64-unknown-linux-gnu` image to choose a glibc

ping #591

a1ecbr0wn:
This is a problem caused by the way we have an out of date version of Centos as the basis for the container.

svenstaro:
I don't really understand the problem or context or goal here given the issues on this that I've read. Is the goal to make everything Ubuntu based or not? What's the glibc issue that's being talked about here?

Emilgardis:
> What's the glibc issue that's being talked about here?

the issue is that some people are still using rhel/centos 7, and we've been kind enough to make it easier for those users to compile with cross without linking in glibc > 2.17, and to do this were using centos 7.

see #455 and #501 where the change was done

Emilgardis:
The reason I personally want all images to be ubuntu is due to #678

otavio:
At O.S. Systems we did a [custom image](https://github.com/OSSystems/rust-cross-images/blob/master/Dockerfile.x86_64-linux-compat) in the lines of Rust toolchain used one which allows even older Glibc (2.12). We could make use of Debian for this and avoid CentOS use.

Alexhuszagh:
> At O.S. Systems we did a [custom image](https://github.com/OSSystems/rust-cross-images/blob/master/Dockerfile.x86_64-linux-compat) in the lines of Rust toolchain used one which allows even older Glibc (2.12). We could make use of Debian for this and avoid CentOS use.

I don't think we need to go below 2.17 (RHEL7), but that could work nicely.

Emilgardis:
This has now been solved.

To opt in to using CentOS, specify in your cross configuration (e.g `Cross.toml`)

```toml
[target.x86_64-unknown-linux-gnu]
[...]

PR #1251, opened 2023-05-02
Update README.md for centos image target notes [5] and [7]

fatihaziz:
Based on https://github.com/cross-rs/cross/issues/680, it is stated to add `ghcr.io/cross-rs/...` before the image tag. However, this is not stated in the README.md file, resulting in the error "docker login" being needed. 
With this update, the problem is fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ terminate.
 
 [6] libc = emscripten and GCC = clang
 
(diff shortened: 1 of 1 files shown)


Emilgardis (approved):
thanks!

bors r+

PR #1265, opened 2023-05-23
use signal-hook instead of ctrlc

### LINKED ISSUE #1264: Suddenly unusable on CircleCI "machine" VMs
### Checklist

### Describe your issue

I use CircleCI to run cross-compile jobs using `cross` for [Geph](https://github.com/geph-official/geph4-client). To avoid the complications of docker-in-docker, I've always used the "machine" class, which spins up a full virtual machine.

Recently, my build scripts have suddenly started failing with this error:

```
#!/bin/bash -eo pipefail
cross build --release --locked  --target armv7-unknown-linux-musleabihf
Error: 
   0: Ctrl-C error: Ctrl-C signal handler already registered

Exited with code exit status 1
CircleCI received exit code 1
```

This happened even though the `cross` utility did not change versions, and my build script did not change. I suspect it's some issue with the docker container itself; it looks like some conflict with whatever code handles "ctrl+c" gracefully shutting down the docker container, conflicting with a signal handler that circleci attaches, possibly to handle the "stop compile" button.

_No response_

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD

- [X] x86_64 / AMD64
- [ ] arm32

- [X] docker
- [ ] podman

### cross version

cross 0.2.5

### Example

_No response_

_No response_

Emilgardis:
Fixes #1264

### DIFF
--- a/.changes/1265.json
+++ b/.changes/1265.json
@@ -1,0 +1,4 @@
+{
+    "description": "replace `ctrlc` signal handler with `signal-hook`",
+    "type": "changed"
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,7 +208,6 @@ dependencies = [
  "clap",
  "color-eyre",
  "const-sha1",
- "ctrlc",
  "directories",
  "dunce",
  "eyre",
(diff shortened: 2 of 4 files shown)


Detegr:
This change is not functionally equivalent to using Ctrlc with `termination` feature. You're now handling `SIGINT` only, where previously you were handling `SIGTERM` and `SIGHUP` also.

Emilgardis:
@Detegr noted, I wonder if that would be an issue.

Detegr:
It depends whether you want to have for example `kill cross` run the handler or not. I'm not sure if `cross` can be used on Windows, but on Windows killing a process from a task manager or closing the command line window will trigger `CTRL_CLOSE_EVENT` and that won't be handled either.

For just exiting the process with Ctrl-C will work with this solution in unixes and Windows.

Emilgardis:
@Detegr I see that you've reverted the behaviour in ctrlc to how it used to be and introduced try_set_handler instead. Does that mean this pr could be reverted with that update?

Detegr:
Yes, I brought the previous overwriting behavior back as it broke more things than I anticipated (should've been a major number change, I guess).

Issue #1012, opened 2022-09-06
Allow Read-Only or Unlabelled Mounts


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

Currently, if using podman (or likely any rootless container engine), attempting to mount a directory without having write permissions leads to an error (or any directory in `/usr`).

If done in `/usr/` or `/etc`, we get `Error: relabeling content in /usr is not allowed`. Anywhere else without write permissions, we get: `Error: lsetxattr /nix/store: operation not permitted`.

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [X] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [ ] docker
- [X] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.4 (078eab5 2022-07-19)

### Example

On a system with SELinux enabled (Fedora 36), add the following to `Cross.toml`, for a directory without read permissions

```toml
[build.env]
volumes = ["NIX=/nix/store"]
```

And then attempt to run it with the following:

```bash
# as a privileged user
$ sudo mkdir -p /nix/store
# as an unprivileged user
$ CROSS_CONTAINER_ENGINE=podman cross build --target aarch64-unknown-linux-gnu
Error: lsetxattr /nix/store: operation not permitted
```

Likewise, if we try to relabel in `/usr`, we get:

```toml
[build.env]
volumes = [ "MINGW=/usr/x86_64-w64-mingw32"]
```

```bash
# as a privileged user
$ sudo mkdir -p /usr/x86_64-w64-mingw32
# as an unprivileged user
$ CROSS_CONTAINER_ENGINE=podman cross build --target aarch64-unknown-linux-gnu
Error: relabeling content in /usr is not allowed
```

### Additional information / notes

A solution would be to only allow read-only permissions in `/usr` or `/etc` if using podman (using the `ro` SELinux label rather than `z`), and allowing read-only labels for mounted volumes in the TOML configuration. Manually using the command and using `ro` allows the volume to be mounted, even in enforcing mode. Likewise, using no labels also allows the volume to mounted (although attempting to modify data within these volumes will likely be denied).

The best approach would likely be:
[...]

PR #1315, opened 2023-08-21
Specify arbitrary container arguments

har7an:
Implements the ability to pass arbitrary arguments to the underlying container runtime (`podman`/`docker`). This functionality is currently exposed:

- in the config file, as `build.env.extra_args`
- via environment-variables as `EXTRA_ARGS`

and takes a vector of strings. These strings are then injected into the container commandline.
This allows solving problems such as #1012 without having to wait for a new release of `cross` to handle every users needs.

The PR currently lacks tests and documentation. I wanted to put this up here to discuss the idea first before investing more time into it.


Emilgardis (changes requested):
How is this different from [`CROSS_CONTAINER_OPTS`](https://github.com/cross-rs/cross/wiki/Configuration#environment-variables) defined here https://github.com/cross-rs/cross/blob/37c681a923407b67044ea09c40205012c0ea86d5/src/docker/shared.rs#L1057C45-L1057C45?

obviously this adds a way to specify it in the config file, but other than that I don't see any difference.

If it is the same, I'd prefer removing `CROSS_EXTRA_ARGS` and making `CROSS_CONTAINER_OPTS` available in config file.

har7an:
@Emilgardis 

> How is this different from [`CROSS_CONTAINER_OPTS`](https://github.com/cross-rs/cross/wiki/Configuration#environment-variables) defined here https://github.com/cross-rs/cross/blob/37c681a923407b67044ea09c40205012c0ea86d5/src/docker/shared.rs#L1057C45-L1057C45?

I guess it isn't at all, but the cross README (which is the usual entrypoint for people new to a project) has a section *Configuration*, which lists three options, two of which link [here][1]. Neither of the "highlighted" config options talks about configuration through environment, nor does the `cross_toml.md`. The file browser visible to the left when reading `cross_toml.md` through the link listed also shows only a single file in the `docs` folder and no mention of environment variables.

Finding the [docs about environment variables][2] seems to be mostly a matter of coincidence (unless of course I missed something obvious in which case I'm sorry) and now that I found it, it seems to me that this page also supersedes the `cross_toml.md` linked in the README as there seems to be more information about that file as well. [This][3] is also a nice read.

So it seems this is mostly a matter of documentation. I'll happily update the docs if that's alright for you. About exposing the env variable settings through the config file: Should I keep the code snippet you listed and add the file-based config alongside it, or would you prefer if I updated the code in my PR to read `CROSS_CONTAINER_OPTS` (instead of `CROSS_EXTRA_ARGS`) and check for the existence of `DOCKER_OPTS` like the current code does?


[1]: https://github.com/cross-rs/cross/blob/main/docs/cross_toml.md
[2]: https://github.com/cross-rs/cross/wiki/Configuration#environment-variables
[3]: https://github.com/cross-rs/wiki_assets/blob/main/Configuration/crossrc.bash_aliases

Emilgardis:
Alright!

Thank you for identifying the cause of the problem.

We link to the wiki in the README under [#usage](https://github.com/cross-rs/cross#usage), but this can definitely be improved.

I think we could make the situation you've identified better by

* Linking to the wiki in docs/cross_toml.md, `For additional help and examples, see [link to wiki]`
* Update docs/ to be more like https://github.com/cross-rs/cross/wiki/Configuration, like including the env vars available and better examples, we could probably have CI check that they are in-sync before a release, but we can fix that in another PR.
* We'd like the README to not be a dumping ground for all config options, but for something like `CROSS_CONTAINER_OPTS` it would probably be worth mentioning in the readme.

as for the question about config, I'm not entirely sure what I'd prefer myself. There's some alternatives available as I see it

1. Only have an env var, e.g keep it as-is
2. Add a new `build.container-opts` that is another way to set `CROSS_CONTAINER_OPTS`
3. Add `build.container-opts` and `target.TARGET.container-opts` that is another way to set `CROSS_CONTAINER_OPTS`

`CROSS_DOCKER_OPTS` and `CROSS_CONTAINER_OPTS` should just be aliases of eachother while `DOCKER_OPTS` is being deprecated.

har7an:
Hi @Emilgardis,

sorry for taking so long but I recently "rediscovered" my previous progress on the matter on my PC and decided to invest some more time into it. I opened a new PR #1438 that deals with the documentation "issue".

I'm fine with leaving the code as is and having only the env var to set additional arguments for the container runtime. This perfectly suits my needs. I'll leave the PR open for now pending a reaction in #1438. If you feel its' no longer needed you can close it as well, of course.

PR #1332, opened 2023-09-14
fix: ubuntu does not provide mips packages

sarfata:
Ubuntu does not support mips, so we cannot add libraries with pre-build hooks like [suggested in the documentation](https://github.com/cross-rs/cross#pre-build-hook).

This changes the mipsel-unknown-linux-musl image to build from Debian stable.

You can now do:

```
pre-build = ["dpkg --add-architecture mipsel && apt-get update && apt-get install -y zlib1g-dev:mipsel && ln -s /usr/include/zlib.h /usr/local/mipsel-linux-muslsf/include"]
```

I realize this is not a benign change but it ended up being pretty small. I am curious if this is something you would consider merging - or if you would suggest a better approach.

### DIFF
--- a/docker/Dockerfile.mipsel-unknown-linux-musl
+++ b/docker/Dockerfile.mipsel-unknown-linux-musl
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM debian:bookworm
 ARG DEBIAN_FRONTEND=noninteractive
 
 COPY common.sh lib.sh /
--- a/docker/lib.sh
+++ b/docker/lib.sh
@@ -10,7 +10,7 @@ set_centos_ulimit() {
 }
 
 install_packages() {
-    if grep -i ubuntu /etc/os-release; then
+    if grep -i debian /etc/os-release || grep -i ubuntu /etc/os-release; then
         apt-get update
 
         for pkg in "${@}"; do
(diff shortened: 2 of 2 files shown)


docker/lib.sh
```diff
@@ -49,7 +49,7 @@ if_centos() {
 }
 
 if_ubuntu() {
```
Emilgardis, line 51: this should probably be renamed to `if_apt`

Emilgardis (changes requested):
This is a good idea, but maybe we should keep the ubuntu based image just incase? 

One issue I see is our override of [sources here](https://github.com/cross-rs/cross/blob/539a2b07da6718b7e989cab01f47d2796e5936b9/docker/linux-image.sh#L89), should probably remove that for this target.

The ubuntu image can be kept as `docker/Dockerfile.mipsel-unknown-linux-musl.ubuntu`, it would then need to be added to [`targets.toml`](https://github.com/cross-rs/cross/blob/main/targets.toml) like we've done for [`centos`](https://github.com/cross-rs/cross/blob/539a2b07da6718b7e989cab01f47d2796e5936b9/targets.toml#L38-L46)

docker/Dockerfile.mipsel-unknown-linux-musl
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM debian:bookworm
```
joseluisq, line 1:
```suggestion
FROM debian:12.1-slim
```

sarfata:
Sorry - Keep meaning to get back to this but have not found the time. Will close the PR for now and will re-open when I have applied the requested changes.

PR #1340, opened 2023-10-05
Fix non-provided targets with only dockerfile specified

Emilgardis:


src/lib.rs
```diff
@@ -743,6 +743,11 @@ pub fn setup(
     let zig_version = config.zig_version(&target)?;
     let image = match docker::get_image(&config, &target, uses_zig) {
         Ok(i) => i,
+        Err(docker::GetImageError::NoCompatibleImages(..))
+            if config.dockerfile(&target)?.is_some() =>
+        {
+            "localhost/cross-rs/if-you-are-seeing-this-cross-did-something-wrong:temp".into()
```
Emilgardis, line 749: this is not very nice.

Technically this could be just an empty string, it could also be a `Option<PossibleImage>`, but that would then propagate very far into the rest of the code

Emilgardis: maybe `scratch`?

PR #1330, opened 2023-09-12
Update zig-cc and zigbuild versions.

sagudev:

PR #1349, opened 2023-10-17
use cargo-zigbuild 0.17.3 instead of 0.17.2

thomas725:
to solve this problem: https://github.com/rust-cross/cargo-zigbuild/issues/151

### DIFF
--- a/.changes/1330-1349.json
+++ b/.changes/1330-1349.json
@@ -1,0 +1,4 @@
+{
+    "description": "update cargo-zigbuild from 0.17.2 to 0.17.3",
+    "type": "changed"
+}
--- a/docker/zig.sh
+++ b/docker/zig.sh
@@ -119,7 +119,7 @@ install_zig_source() {
 
 install_zigbuild() {
     local platform="${1}"
-    local version="0.17.2"
+    local version="0.17.3"
     local dst="/usr/local"
     local triple=
 


Emilgardis:
```
  > [9/9] RUN /zig.sh linux/amd64:
5.302 + local version=0.17.3
5.302 + local dst=/opt/zig
5.302 + local filename=zig-linux-x86_64-0.17.3.tar.xz
5.302 + local td
5.303 ++ mktemp -d
5.304 + td=/tmp/tmp.hg3QrDA9EJ
5.304 + pushd /tmp/tmp.hg3QrDA9EJ
5.304 /tmp/tmp.hg3QrDA9EJ /
5.304 + curl --retry 3 -sSfL https://ziglang.org/download/0.17.3/zig-linux-x86_64-0.17.3.tar.xz -O
he requested URL returned error: 403 
```

sorry! I was a bit hasty in my response in matrix, it's line 121 that needs to be updated. We probably also should update zig itself

thomas725:
hey there! I just checked https://ziglang.org/download/ and it looks like 0.11.0 is still the latest stable version, the only newer one is 0.12.0.dev, and this "dev" suffix and the block header "master" seem's to suggest this is an unstable version.

I'll research how to modify my pull request to undo the wrong change and do the correct one you suggested.

thomas725:
okey, and now about the cangelog, if I understand the README correctly we'd need to first create an issue for this in the "cross" project, we probably can't just reference the one that exists for the "cargo-zigbuild" project, right?

thomas725:
Okey, looking at the previous changelog entries I see the issue property is optional, so I created one without.

Emilgardis:
perfect!

bors try --target zig

Emilgardis:
also, I totally forgot about https://github.com/cross-rs/cross/pull/1330

so I think the problem is solved already on main since https://github.com/rust-cross/cargo-zigbuild/commit/79e53f4c36f2d8466ed3b13968b5d61a7ec16f93 exists in 0.17.2 :3

thomas725:
Ah, you're right!
https://github.com/rust-cross/cargo-zigbuild/issues/151
has been solved by
https://github.com/rust-cross/cargo-zigbuild/commit/79e53f4c36f2d8466ed3b13968b5d61a7ec16f93
from 2023-07-29_0604, the first release containing that commit was
https://github.com/rust-cross/cargo-zigbuild/releases/tag/v0.17.0
from 2023-07-29_0656.

And since before this pull request we're already using 0.17.2 upgrading is not required to solve this issue.

My open question then is: Why did I ran into this issue and how can I use a version of cross-rs new enough to not have this issue?

Emilgardis:
You'd want to use a newer image, a versioned cross installed like `cargo install cross` or `cargo install cross --version a.b.c` will use images with the tag `:a.b.c`. To use another image you'd either set the image manually to be [`image = "ghcr.io/cross-rs/zig:main"`](https://github.com/cross-rs/cross/pkgs/container/zig), install cross from our main branch `cargo install cross --git https://github.com/cross-rs/cross` or install cross with a locally checked out clone.

We should probably update zigbuild to 0.17.3 anyway, can you change the changelog-file to be named `1330-1349.json` and then squash the changes, see https://rustc-dev-guide.rust-lang.org/git.html#squash-your-commits if you're not familiar how to do it

thomas725:
squash done

Emilgardis:
there is no glibc 1.2 so that's the reason you're getting an error :)

thomas725:
> there is no glibc 1.2 so that's the reason you're getting an error :)
> 
> edit: you can find a list of versions [here](https://sourceware.org/glibc/wiki/Glibc%20Timeline)

ah, thanks for the link to the existing glibc versions. my PowerPC OpenWRT device says it has:
```
$ opkg info libc
Package: libc
Version: 1.2.3-4
```
But I guess that's not glibc?

Also, changing my `Cross.toml` section to read `zig = true` instead gives the same error.

Emilgardis:
yes @thomas725 , as I explained vaguely, `1.2` is musl libc, the libc package on openwrt is a metapackage that is either musl or glibc

I'm not sure about the zig error, maybe it's the linux version it's having issues with :/

thomas725:
I don't know where the linux kernel version 5.10 is coming from, my physical machine is running on 6.1.55

Issue #1249, opened 2023-05-02
Migrate away from bors-ng


Emilgardis:
Bors-NG is now deprecated, see https://bors.tech/newsletter/2023/05/01/tmib-76/

We have two options as specified in the post.

1. Use a self-hosted instance of Bors-NG
2. Use built-in [github merge queues](https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/)

Emilgardis:
For option 2, merge queues, we should find a way to do what `bors try --args` does. I imagine we can use [`issue_comment`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) to solve this.

Emilgardis:
bors is dead now, this means we have to switch over

for now, I'll disable the check and enable merge-queue + fix the ci script. We'll have to look into `bors try`, but I think normal queue will work fine with `pull-request`.

PR #1375, opened 2023-11-18
rip bors - switch to merge queues

Emilgardis:
this pr implements part of #1249

We now use merge queues, which work in a similar way, except there is no batching of prs.


Emilgardis:
ehm, interesting, that shouldn't have merged...

Issue #1098, opened 2022-10-29
error: failed to create directory `/target/release`


abcd-ca:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

My project stopped cross compiling (even when I comment everything out except for a hello world). 

I'm sorry I don't know how to reproduce but my question is: Where is the location of whatever is being held onto that is preventing this from working. 

My command is, `cross build --verbose --release --target=armv7-unknown-linux-musleabihf`

The error I get is, "error: failed to create directory `/target/release`"

This project used to compile. I think I might have corrupted something by cancelling (ctrl-c on mac) during a cross build.

Here's the interesting bit: If I rename the root project folder (the one that contains my Cargo.toml) then re-run the command, it succeeds!

I have tried,

* `cargo clean`
* `cross clean`
* uninstalled cross, deleted `~/.cargo/git` and `~/.registry` folders, reinstalled cross 
* deleting all the docker containers and images that are related to this project and cross (I see no volumes to delete in docker)

### What target(s) are you cross-compiling for?

armv7-unknown-linux-musleabihf

### Which operating system is the host (e.g computer cross is on) running?

- [X] macOS
- [ ] Windows
- [ ] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [ ] x86_64 / AMD64
- [ ] arm32
- [X] arm64 (including Mac M1)

### What container engine is cross using?

- [X] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

1.64.0

### Example

1. `cd my-project`
2. `cross build --verbose --release --target=armv7-unknown-linux-musleabihf`. Fails, "error: failed to create directory `/target/release`"
3. Duplicate my project in macOS, name it my-project-copy
4. `cd ../my-project-copy`
5. `cross build --verbose --release --target=armv7-unknown-linux-musleabihf`. Succeeds! "Finished release [optimized] target(s) in 12.12s"

### Additional information / notes

I can reproduce the same issue with `aarch64-unknown-linux-gnu` target. It's all about the root project folder name.

Emilgardis:
Can you post the verbose output of the failing build?

what version of cross are you actually using? You've shown us what version cargo is, but not cross

abcd-ca:
@Emilgardis Thank you for taking a look. 

`cross --version` output:

``` 
cross 0.2.4 (a8082de 2022-10-26)
[...]

PR #1380, opened 2023-12-03
feat(rootless): help detection with execv check "is current builder rootless?"

fenollp:
I recently encountered #1098 and propose an iteration on the fix in #890

This patch adds a last minute check that looks up the current builder endpoint. This may be seen as a costly operation however, as mentioned in #889

Here's the output of `docker builder inspect` on my rootless install:
```
Name:          rootless
Driver:        docker
Last Activity: 2023-12-03 02:04:14 +0000 UTC

Nodes:
Name:      rootless
Endpoint:  rootless  # <= THIS HERE
Status:    running
Buildkit:  v0.11.7+d3e6c1360f6e
Platforms: linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
Labels:
 org.mobyproject.buildkit.worker.moby.host-gateway-ip: 172.17.0.1
```

### DIFF
--- a/src/docker/engine.rs
+++ b/src/docker/engine.rs
@@ -155,19 +155,58 @@ fn is_rootless(ce: &Path, msg_info: &mut MessageInfo, kind: EngineType) -> bool 
             b => Some(bool_from_envvar(b)),
         })
         .or_else(|| (!kind.is_docker()).then_some(true))
-        .unwrap_or_else(|| {
-            let mut cmd = Command::new(ce);
-            cmd.args(["info", "-f", "{{.SecurityOptions}}"])
-                .run_and_get_output(msg_info)
-                .ok()
-                .and_then(|cmd| cmd.stdout().ok())
-                .map(|out| {
-                    out.to_lowercase()
-                        .replace([' ', '[', ']'], ",")
-                        .contains(",name=rootless,")
-                })
-                .unwrap_or_default()
+        .unwrap_or_else(|| is_docker_rootless(ce, msg_info))
+}
+
+fn is_docker_rootless(ce: &Path, msg_info: &mut MessageInfo) -> bool {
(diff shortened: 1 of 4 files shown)


fenollp:
> [name=seccomp,profile=builtin name=rootless name=cgroupns]

Alright I'll see about extracting this then, then :)

fenollp:
@Emilgardis here ya go

src/docker/engine.rs
```diff
@@ -143,6 +146,69 @@ impl Engine {
...
+fn is_docker_rootless(ce: &Path, msg_info: &mut MessageInfo) -> bool {
+    let mut cmd = Command::new(ce);
+    cmd.args(["info", "-f", "{{.SecurityOptions}}"])
+        .run_and_get_output(msg_info)
+        .ok()
+        .and_then(|cmd| cmd.stdout().ok())
+        .map(|out| {
+            out.to_lowercase()
+                .replace(['[', ' ', ']'], ",")
+                .contains(",name=rootless,")
```
Emilgardis, line 169: could this just be

```suggestion
            out.to_lowercase()
                .contains("name=rootless")
```

Emilgardis (changes requested):
Nice!

src/docker/engine.rs
fenollp: It could, I just wanted to be safe.

The patch you suggest allows for all keys ending in `name` and all values starting with `rootless` :)

fenollp: Hi @Emilgardis All good with you?

Emilgardis: Sorry! Yes, this looks good

Emilgardis (approved):
LGTM!

Issue #1389, opened 2023-12-21
Investigate bindgen woes - "file not found"


Emilgardis:
Some (if not most?) `*-dev` packages on ubuntu dist only include headers in `/usr/include`, this means that the `BINDGEN_EXTRA_CLANG_ARGS_<target>=--sysroot=$CROSS_SYSROOT` that we set in our images means `bindgen` can't properly find these headers if the `*-sys`-crate doesn't use `pkg-config` probing.

`bindgen` errors with: `Unable to generate bindings: ClangDiagnostic("wrapper.h:1:10: fatal error: 'mylib.h' file not found\n")`

To solve this problem, there's two options, nr.1 requires a change in the `sys`-crate

## 1. Modify the `sys`-crate

Crates using bindgen for API generation can do the following

```rs
std::env::set_var("PKG_CONFIG_ALLOW_SYSTEM_CFLAGS", "1");
let library = pkg_config::probe_library("mylib").expect("Unable to probe `mylib`");
let bindings = bindgen::Builder::default()
    .clang_args(library.include_paths.iter().map(|path| format!("-I{}", path.to_string_lossy())))
    .header("wrapper.h")
    .parse_callbacks(Box::new(bindgen::CargoCallbacks))
    .generate()
    .expect("Unable to generate bindings");
```
## 2. modify `BINDGEN_EXTRA_CLANG_ARGS_<target>` to include `/usr/include`

(make sure to get the correct sysroot path and use the correct target names, `<target>` is lowercase snake_case target triple, `<TARGET>` is uppercase kebab-case target triple.)

```toml
# Cross.toml
[target.<TARGET>.env]
passthrough = ["BINDGEN_EXTRA_CLANG_ARGS_<target>=--sysroot=<path_to_target_sysroot> -idirafter/usr/include"]
```

## Going forward

I've filed https://github.com/rust-lang/rust-bindgen/issues/2701 to potentially make option 2 easier, as we then could just set `BINDGEN_EXTRA_CLANG_ARGS=-idirafter/usr/include` for all targets and all is well.

Can we solve this problem in our images? 

Related: https://github.com/rust-lang/rust-bindgen/issues/1247 for pkg-config with bindgen.

Emilgardis:
> Can we solve this problem in our images?

[...]

PR #1391, opened 2023-12-22
Add `-idirafter/usr/include` to bindgen's clang invocation on GNU/Linux

Emilgardis:
A workaround/solution to #1389


Emilgardis:
I'm merging this :)

PR #1393, opened 2023-12-23
fix linux kernel

Emilgardis:


Emilgardis:
this should fix most build errors, still some remaining but getting closer

PR #1406, opened 2024-01-07
fix try comment

Emilgardis:


Emilgardis:
going to wait for https://github.com/Emilgardis/cross/actions/runs/7438286897?pr=1+%28comment%29 to make sure that the comment actually works

PR #1409, opened 2024-01-08
apt-get: make sure DEBIAN_FRONTEND is propagated to env

ydirson:
DEBIAN_FRONTEND was set to "interactive" in Dockerfile's using an ARG directive.  For some (undocumented?) reason it seems to let this variable each the environment of RUN commands, when building with Docker.  OTOH Podman does not do that (which AFAICT is consistent with Docker's own doc), and with `podman build` we experience the problem described in #589.

ARG's documented use anyway is to let the caller of the build to set a variable's value, and we don't need that.  OTOH we also want this env value to persist into the container, so installation of build-deps for a given task do not suffer the same problem, so that undocumented "feature" of ARG in Docker is indeed insufficient, it should really have been ENV from the start.


Emilgardis:
For completion, we do need the envvar set during the builds of our images. tz which is a dependency of something we use (can't remember what) needs user interaction.

Making the variable persist makes sense to me, it shouldn't do any harm and only help.

Emilgardis (approved):
Thanks!

PR #1411, opened 2024-01-09
freebsd: let helper scripts work not just from /

ydirson:
Scripts should not source their libs as if they were going to be run with `PWD=/`


Emilgardis (approved):
thanks!

Issue #1055, opened 2022-10-05
docker in pre-build is failing: `unknown flag: --label`


qiujiangkun:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

docker --version
Docker version 20.10.18, build b40c2f6b5d

```
cross build
unknown flag: --label
See 'docker --help'.

Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

Options:
      --config string      Location of client config files (default
                           "/home/jack/.docker")
  -c, --context string     Name of the context to use to connect to the daemon
                           (overrides DOCKER_HOST env var and default context set
                           with "docker context use")
  -D, --debug              Enable debug mode
  -H, --host list          Daemon socket(s) to connect to
  -l, --log-level string   Set the logging level
                           ("debug"|"info"|"warn"|"error"|"fatal") (default "info")
      --tls                Use TLS; implied by --tlsverify
      --tlscacert string   Trust certs signed only by this CA (default
                           "/home/jack/.docker/ca.pem")
      --tlscert string     Path to TLS certificate file (default
                           "/home/jack/.docker/cert.pem")
      --tlskey string      Path to TLS key file (default "/home/jack/.docker/key.pem")
      --tlsverify          Use TLS and verify the remote
  -v, --version            Print version information and quit

Management Commands:
  builder     Manage builds
  compose*    Docker Compose (Docker Inc., 2.11.1)
  config      Manage Docker configs
  container   Manage containers
  context     Manage contexts
  image       Manage images
  manifest    Manage Docker image manifests and manifest lists
  network     Manage networks
  node        Manage Swarm nodes
  plugin      Manage plugins
  secret      Manage Docker secrets
  service     Manage services
  stack       Manage Docker stacks
  swarm       Manage Swarm
  system      Manage Docker
  trust       Manage trust on Docker images
  volume      Manage volumes

Commands:
  attach      Attach local standard input, output, and error streams to a running container
  build       Build an image from a Dockerfile
  commit      Create a new image from a container's changes
  cp          Copy files/folders between a container and the local filesystem
  create      Create a new container
  diff        Inspect changes to files or directories on a container's filesystem
  events      Get real time events from the server
  exec        Run a command in a running container
  export      Export a container's filesystem as a tar archive
  history     Show the history of an image
  images      List images
  import      Import the contents from a tarball to create a filesystem image
[...]

PR #1419, opened 2024-01-19
Remove the --progress flag

chanderlud:
Newer versions of Docker no longer support the --progress flag on the docker build command. This causes Cross to fail when using versions of Docker such as 23 and 24. 

For example:
```
docker --version
Docker version 24.0.5, build 24.0.5-0ubuntu1~22.04.1
cross build --target aarch64-unknown-linux-gnu
info: downloading component 'rust-std' for 'aarch64-unknown-linux-gnu'
info: installing component 'rust-std' for 'aarch64-unknown-linux-gnu'
unknown flag: --progress
See 'docker --help'.
```

I have removed the progress code from cross in this pull request, this error no longer occurs.

### DIFF
--- a/docker/musl.sh
+++ b/docker/musl.sh
@@ -51,6 +51,7 @@ main() {
         LINUX_HEADERS_SITE="${linux_headers_site}" \
         LINUX_VER="${linux_ver}" \
         OUTPUT=/usr/local/ \
+        "GCC_CONFIG += --enable-default-pie" \
         "${@}"
 
     purge_packages


Emilgardis:
This pr solves a problem which doesn't exist, the situation here is unfortunate though.

We don't call `docker build`, we call `docker buildx build` , which absolutely [does support `--progress`](https://docs.docker.com/engine/reference/commandline/buildx_build/#progress). The issue here is how docker handles error reporting. See https://github.com/cross-rs/cross/issues/1055

You **should** be getting an error message in the cross output suggesting how to fix this, which is to check that [buildx/buildkit](https://docs.docker.com/build/architecture/#install-buildx) is available or to set `CROSS_CONTAINER_ENGINE_NO_BUILDKIT=1`

chanderlud:
Yup, installing buildx got rid of the progress error. Do you think it would be reasonable for Cross to check if buildx is installed before trying to use it? That way a more helpful error message could be delivered to the user.

Emilgardis:
If you can find a good way to detect buildx, I'd be happy to include a check like that!

chanderlud:
It looks like the `docker buildx version` command is a good candidate for testing installation. I wrote this function to test for buildx being installed, but I am not sure what the best place to call it from would be.
```rust
fn is_buildx() -> bool {
    let mut cmd = Command::new("docker");
    cmd.args(["buildx", "version"])
        .output()
        // if buildx is present, it will print the version and exit 0
        // if buildx is not present, it will print a warning and exit 1
        .map_or(false, |out| out.status.success())
}```

Issue #1037, opened 2022-09-29
Add UI test


Emilgardis:
We should consider using UI tests to the CI/tests ala [`snapbox`](https://docs.rs/snapbox/latest/snapbox/) or [`trycmd`](https://docs.rs/trycmd/latest/trycmd/)

this would help catch unexpected changes to output that don't necessarily cause the build to error.

Alexhuszagh:
I think [assert_cmd](https://crates.io/crates/assert_cmd) might be the proper tool for the job here:
https://crates.io/crates/assert_cmd

Alexhuszagh:
I think some variant of this should be finished prior to 0.3.0, just to avoid any regressions during our next release.

PR #1420, opened 2024-01-24
WIP: set QEMU_LD_PREFIX in linux-runner in order to be able to use env -i

glehmann:
I'm working on making assert_cmd working with cross.

There is one case a bit more difficult:
when using `clear_env()` on a process, or `env -i`, the binary is not able to run anymore because the
`QEMU_LD_PREFIX` envvar is not set anymore.

A (relatively) easy fix might be to ensure in `linux-runner` — and in any other runner — that the
necessary environment is configured before running the executable. Something as proposed
in this PR for `aarch64-unknown-linux-gnu`.

Would that be an acceptable solution?

https://github.com/assert-rs/assert_cmd/pull/193


docker/linux-runner
```diff
@@ -157,4 +157,6 @@ if [[ -t 1 ]] && [[ -t 2 ]]; then
   tty_flag='-t'
 fi
 
+export QEMU_LD_PREFIX=${QEMU_LD_PREFIX:-@DEFAULT_QEMU_LD_PREFIX@}
```
Emilgardis, line 160: This should be documented with how and why this is done.

```suggestion
# Ensure that the correct prefix is set.
# `@DEFAULT_QEMU_LD_PREFIX@` is replaced during image build.
export QEMU_LD_PREFIX=${QEMU_LD_PREFIX:-@DEFAULT_QEMU_LD_PREFIX@}
```

Emilgardis (changes requested):
Nice workaround! I'm fine with this approach, I would however like to see it done in all images and not just aarch64!

I also have a slight concern about what happens if the env is wiped and `@DEFAULT_QEMU_LD_PREFIX@` hasn't been replaced by us. We should, if this lands, ensure that https://github.com/cross-rs/cross-toolchains also has been updated

glehmann:
> Nice workaround! I'm fine with this approach, I would however like to see it done in all images and not just aarch64!

Sure! I was just checking that you were ok with that before patching all the images — there is quite a lot :)

I'm thinking to start by adding a test that runs `env -i ${CROSS_TARGET_RUNNER} /a/command`. That way we should be sure that we have the minimum environment configured in each image.

> I also have a slight concern about what happens if the env is wiped and `@DEFAULT_QEMU_LD_PREFIX@` hasn't been replaced by us.

`@DEFAULT_QEMU_LD_PREFIX@` is just the default value in `${QEMU_LD_PREFIX:-@DEFAULT_QEMU_LD_PREFIX@}`, and `QEMU_LD_PREFIX` is still set in the docker image. So if `@DEFAULT_QEMU_LD_PREFIX@` is not replaced by its value, cross will continue to work as it does today: just fine in most case, and not able to run the executable if the env is cleaned.

> We should, if this lands, ensure that https://github.com/cross-rs/cross-toolchains also has been updated

ok, just a few more images ;-)

Emilgardis:
> `@DEFAULT_QEMU_LD_PREFIX@` is just the default value in `${QEMU_LD_PREFIX:-@DEFAULT_QEMU_LD_PREFIX@}`, and `QEMU_LD_PREFIX` is still set in the docker image. So if `@DEFAULT_QEMU_LD_PREFIX@` is not replaced by its value, cross will continue to work as it does today: just fine in most case, and not able to run the executable if the env is cleaned.

I understand this, my concern is simply about what qemu does when the env has been cleared, the default has not been replaced by the docker build and the only path in `QEMU_LD_PREFIX` is a relative-looking non-existant path (`@DEFAULT_QEMU_LD_PREFIX@`). I don't think anything bad will happen but just wanted to raise it :D.


`rg -l 'CROSS_TARGET_RUNNER=' -g 'Dockerfile.*linux*'` should be all the files that need to be fixed (haven't tested this, no access to my computer rn)

glehmann:
> I understand this, my concern is simply about what qemu does when the env has been cleared, the default has not been replaced by the docker build and the only path in `QEMU_LD_PREFIX` is a relative-looking non-existant path (`@DEFAULT_QEMU_LD_PREFIX@`). I don't think anything bad will happen but just wanted to raise it :D.

so I've tested it:
* the run fails in the exact same way when the environment variables are cleared as when `QEMU_LD_PREFIX` is not set. Qemu is probably checking if some files exist in `@DEFAULT_QEMU_LD_PREFIX@`, but there's nothing visible for the end user.
* everything runs at it should when not clearing the environment variables

docker/android-runner
```diff
@@ -31,6 +31,10 @@ case "${arch}" in
 esac
 libdir="/android-ndk/sysroot/usr/lib/${android_abi}"
 
+# Ensure that the correct prefix is set even if the user has cleared the env.
+# `@DEFAULT_QEMU_LD_PREFIX@` is replaced during image build.
+export QEMU_LD_PREFIX=${QEMU_LD_PREFIX:-@DEFAULT_QEMU_LD_PREFIX@}
+
```
Emilgardis, line 37: I don't see much value in adding this on android, we don't depend on this var

Emilgardis (changes requested):

Emilgardis:
The errors are expected I think

Emilgardis:
I just realised something I shouldve thought about 😊 the tests dont ever hit this :D

Emilgardis:
I think we should a new testcase that either uses assert_cmd or simulates the env clearing.

glehmann:
I think the tests won't help until the PR made it to the `main` branch: they are not using the image from this PR.

https://github.com/cross-rs/cross/actions/runs/7679215720/job/20929912781?pr=1420##step:11:29

interesting to see that some tests are passing though — `QEMU_LD_PREFIX` doesn't seem to be necessary for all the targets

Emilgardis:
They are definitely using the new images, see [`--tag ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main`](https://github.com/cross-rs/cross/actions/runs/7679215720/job/20929912781?pr=1420##step:10:130) and [sed](https://github.com/cross-rs/cross/actions/runs/7679215720/job/20929912781?pr=1420##step:10:313)

glehmann:
oh, ok my bad: I've been confused by the name `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main`. I've downloaded it locally and it doesn't contain the changes of the PR. 

I'm looking at it

glehmann:
I tried to rebuild the image from scratch and rerun the test locally — it still runs without problem.

I've set `CROSS_DEBUG` in `test.sh` to get a better chance to understand what is going on in the CI

glehmann:
my attempt at triggering a new test run deesn't seem very effective

@Emilgardis could you trigger a new run for this PR?

Emilgardis:
If you can't figure it out I'll take a closer look into it myself! Seems very strange that it passes for you locally :/

Sidenote: I've been wanting to make the tests a bit more reliable to run and be more structured, meaning migrating the bash script to rust. I have a local branch doing just that but need to finish it

glehmann:
on the CI:
* a first run is passing with the `qemu-user` runner https://github.com/cross-rs/cross/actions/runs/7687457572/job/20947526693?pr=1420##step:12:4295
* a second run fails with the `qemu-system` runner https://github.com/cross-rs/cross/actions/runs/7687457572/job/20947526693?pr=1420##step:12:4624

it fails on the line 3 of `launch.rs`: it can't retreive the `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_RUNNER` env var.

I don't know what is `qemu-system` is it possible that it clears the environment variables?

glehmann:
BTW I am launching the test locally with `TARGET=aarch64-unknown-linux-gnu CROSS_TARGET_AARCH64_UNKNOWN_LINUX_GNU_IMAGE=cross-aarch64-unknown-linux-gnu ./ci/test.sh`, without any `RUNNERS` env var, so I am never running with `qemu-system`.

Emilgardis:
ofc! That's a tricky problem, I'd say lets just skip this test if the runner is `qemu-system`

glehmann:
> ofc! That's a tricky problem, I'd say lets just skip this test if the runner is `qemu-system`

done! Could you trigger the CI again?

ci/test.sh
```diff
@@ -243,6 +243,35 @@ main() {
...
+        .unwrap();
+    std::process::exit(status.code().unwrap());
+}
+EOF
+    cross_build --target "${TARGET}"
+    # don't use cross_run here to not test with qemu-system
+    echo > "${CARGO_TMP_DIR}"/Cross.toml
+    "${CROSS[@]}" run --target "${TARGET}" --bin launch  ${CROSS_FLAGS}
+    popd
+    rm -rf "${td}"
```
Emilgardis, line 274:
```suggestion
    # test running binaries with cleared environment
    if (( ${RUN:-0} )); then
	    td="$(mkcargotemp -d)"
	    pushd "${td}"
	    cargo init --bin --name foo .
	    mkdir src/bin
	    upper_target=$(echo "${TARGET}" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
	    cat <<EOF > src/bin/launch.rs
fn main() {
    let runner: Vec<_> = std::env::var("CARGO_TARGET_${upper_target}_RUNNER")
        .unwrap()
        .split(' ')
        .map(str::to_string)
        .collect();
    let status = std::process::Command::new(&runner[0])
        .args(&runner[1..])
        .arg("/target/${TARGET}/debug/foo")
        .env_clear()
        .status()
        .unwrap();
    std::process::exit(status.code().unwrap());
}
EOF
	    cross_build --target "${TARGET}"
	    # don't use cross_run here to not test with qemu-system
	    echo > "${CARGO_TMP_DIR}"/Cross.toml
	    "${CROSS[@]}" run --target "${TARGET}" --bin launch  ${CROSS_FLAGS}
	    popd
	    rm -rf "${td}"
	fi
```

Emilgardis (changes requested):

Emilgardis:
going to do one last try with all targets to see if we missed anything obvious, I expect this try to fail for mips and riscv due to other problems not related to this changeset.

glehmann:
apparently we'll need a few more environment variables: https://github.com/cross-rs/cross/actions/runs/7693713886/job/20963153896#step:12:3102

the `CARGO_TARGET_*_RUNNER` can't be found on windows: https://github.com/cross-rs/cross/actions/runs/7693713886/job/20963144833#step:12:847 . Same for wasm: https://github.com/cross-rs/cross/actions/runs/7693713886/job/20963151310#step:12:839 .

I'm not sure how much sense it makes to run this kind of test in wasm: are we even allowed to start a new process?

For windows I think I have to write the test to not depend on `CARGO_TARGET_*_RUNNER` env var, because the `launch` binary can probably launch another program without a runner. We probably have to add a `.exe` in the command path…

zig test is failing, but it doesn't seem related to the changes here was it failing before?

Emilgardis:
The missing variables are `CROSS_TARGETARCH`, `CROSS_TARGETVARIANT`, they are only used in `Dockerfile.native` and `Dockerfile.native.centos`

For windows the problem is similar to the issue with `qemu-system`. We want to not respect the runner, just run it directly.

For wasm I have absolutely no idea :/

Emilgardis:
Hmm actually, I might've misspoke there, on windows we do want to run `wine`, we don't emulate cargo...

We need to preserve `WINEPATH` and `WINEPREFIX`, however the error message doesn't make sense to me...

glehmann:
> The missing variables are `CROSS_TARGETARCH`, `CROSS_TARGETVARIANT`, they are only used in `Dockerfile.native` and `Dockerfile.native.centos`

this problem should be fixed :-)

> For windows the problem is similar to the issue with `qemu-system`. We want to not respect the runner, just run it directly.

I've changed the test to only use the runner if it's defined in the env var — in fact that's exactly what I did in `assert_cmd` — and I rolled back my changes to not test `qemu-system` since they are now passing :-)

glehmann:
> Now thinking about this even more. The issue here is that build.rs is ran on the host right? The problem isn't "`cross run --target <target>` can't execute a binary compiled for it's own `<target>` via `std::process::Command`" right?

the problem is to run everything for the target passed to cross. That may or may not require a runner depending on the virtualisation system used. I think the test does the right thing

I think we are not that far to have everything working :-)

Emilgardis:
test should use `CROSS_FLAGS` to correctly use build-std

glehmann:
> test should use `CROSS_FLAGS` to correctly use build-std

mmm, I think they are, aren't they?

I don't see what I have changed that makes the bsd systems fail :-/

glehmann:
for windows, `launch.exe`, which is launched with `wine` is able to launch `foo.exe` directly, without calling `wine`.

The test fails if I make `launch.exe` use `wine` because it can't find the `wine` executable. So I guess I have to drop the `CARGO_TARGET_<target>_RUNNER`, so we don't try to use wine if we are already in wine

Emilgardis:
you removed the if RUN check, that should be there because some of these we can't run on at all

glehmann:
ok, I misunderstood what RUN was for — this is happening too often for a single PR!

glehmann:
should be ok for windows too. I've used `env -u` to remove the runner config from the env. Just let me know if you prefer a plain script :-)

glehmann:
and I need to look into what is going on with wasm.

The project is more complex than I thought (maybe I've been fooled by how easy it is to use), and this shows in the test suite. I'm curious to see what approach you chose to refactor the test suite in rust. Is it available in a branch?

Emilgardis:
> The project is more complex than I thought

Yes :D 

> I'm curious to see what approach you chose to refactor the test suite in rust. Is it available in a branch?

Its nothing concrete right now, basically replace bash with rust, then rewrite it to be prettier, we do have https://github.com/cross-rs/cross/issues/1037 for the bigger issue of actual UI tests

Emilgardis:
failures look fine, except [the wasm one](https://github.com/cross-rs/cross/actions/runs/7703230077/job/20993309896?pr=1420#), should probably just exempt it from running this test.

ci/test.sh
```diff
@@ -243,6 +243,38 @@ main() {
...
+    } else {
+        vec![]
+    };
+    let executable = format!("/target/${TARGET}/debug/foo{}", std::env::consts::EXE_SUFFIX);
+    command.push(executable.to_string());
+    let status = dbg!(std::process::Command::new(&command[0])
+        .args(&command[1..])
+        .env_clear()) // drop all environment variables
+    .status()
+    .unwrap();
```
Emilgardis, line 269: should be indented

Emilgardis (changes requested):
LGTM, just one minor fix and this should be good!

ci/test.sh
glehmann: `cargo fmt` seems to disagree here :-)

that's because of the `dbg!()` around the command. If it's not there `cargo fmt` indeed indent these lines

Emilgardis: alright, lets let it be :)

Emilgardis (approved):
Thank you!

glehmann:
thanks for the kind guidance during the PR!

I'll look at https://github.com/cross-rs/cross-toolchains soon

PR #1426, opened 2024-01-31
improve changelog xtask

Emilgardis:


Emilgardis:
oops, I had already merged this, so the commits make no sense really, oh well!

PR #1431, opened 2024-02-01
allow try to specify tests, and add `no-ci-tests` label to skip tests

Emilgardis:


Emilgardis:
try wont work properly until this is merged since this edits the action

PR #1433, opened 2024-02-01
improve tests

Emilgardis:
fixes issue introduced in #1431


Emilgardis:
somehow this left out the actual fix... pushed db19a77e directly to fix this...

PR #1438, opened 2024-02-06
Docs: Restructure docs folder

har7an:
Sort of a follow-up to #1315 which revealed a "mismatch" between what the `README` and `docs/` folder immediately provide as info compared to what's actually available.

This restructures the `docs/` folder and adds information from various sources to it:

- The README,
- The Wiki
- Examples found in the Wiki

At the moment this duplicates a decent amount of information already available on the Wiki. Strictly speaking this was already the case before this MR, where the `README` and `cross_toml` each contained information from the Wiki. My motivation for moving these things into the `docs/` folder is that:

1. I'm much more used to finding information in a `docs` folder than to check for the Wiki. Most projects just don't use the Wiki
2. I usually expect the documentation to be part of what I check out, and the Wiki is *not* a part of this repo, but its own repo
3. Docs can be updated *as part of the same PR* that updates the respective code, and arbitrary checkouts of the repo will always contain the docs *valid for that version*

On a personal note, I much prefer editing text in an editor I'm comfortable with, and Browsers don't fall into that category.

I'm curious to hear your thoughts on this proposed change.


docs/config_file.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+`cross` will build and use the image that was built instead of the default
+image. It's recommended to base your custom image on the default Docker image
+that `cross` uses: `ghcr.io/cross-rs/{{TARGET}}:{{VERSION}}` (where
+`{{VERSION}}` is `cross`'s version). This way you won't have to figure out how
+to install a cross-C toolchain in your custom image.
+
+> *NOTE*: `$CROSS_DEB_ARCH` is automatically provided by cross, [see
+> here][custom_images_automatic_arch].
```
Emilgardis, line 103: github ui is giving me red marks in the document from here and onwards. strange

image

I can't see anything obviously wrong with it, doubt this is a problem with the file and just how they parse md for syntax

docs/config_file.md
```diff
@@ -0,0 +1,304 @@
...
+The `cross` configuration in the `Cross.toml` file can contain the following
+elements:
+
+
+# `build`
+
+The `build` key allows you to set global variables, e.g.:
+
+> *NOTE*: `$CROSS_DEB_ARCH` is automatically provided by cross, [see
+> here][custom_images_automatic_arch].
```
Emilgardis, line 37:
```suggestion
> *NOTE*: `$CROSS_DEB_ARCH` is automatically provided by cross, 
> [see here][custom_images_automatic_arch].
```

README.md
```diff
@@ -323,91 +194,99 @@ QEMU gets upset when you spawn multiple threads. This means that, if one of your
...
+| `armv5te-unknown-linux-gnueabi`        | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
+| `armv5te-unknown-linux-musleabi`       | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
+| `armv7-linux-androideabi` [1]          | 9.0.8  | 9.0.8  | ✓   | 6.1.0 |   ✓    |
+| `armv7-unknown-linux-gnueabi`          | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
+| `armv7-unknown-linux-gnueabihf`        | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
+| `armv7-unknown-linux-musleabi`         | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
+| `armv7-unknown-linux-musleabihf`       | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
+| `i586-unknown-linux-gnu`               | 2.31   | 9.4.0  | ✓   | N/A   |   ✓    |
+| `i586-unknown-linux-musl`              | 1.2.3  | 9.2.0  | ✓   | N/A   |   ✓    |
+| `i686-unknown-freebsd`                 | 1.5    | 6.4.0  | ✓   | N/A   |        |
```
Emilgardis, line 217: the diff here looks wrong, specifically in regards to the whitespace in the C++ and test column

docs/environment_variables.md
```diff
@@ -0,0 +1,116 @@
...
+  runs as root and all other container engines are rootless.
+- `CROSS_CONTAINER_USER_NAMESPACE`: Custom the [container user
+  namespace][container-user-namespace]. If set to `none`, user namespaces will
+  be disabled. If not provided or set to `auto`, it will use the default
+  namespace.
+- `CROSS_CUSTOM_TOOLCHAIN_COMPAT`: A descriptive name for a custom toolchain so
+  `cross` can convert it to a fully-qualified toolchain name.
+- `CROSS_CONTAINER_ENGINE_NO_BUILDKIT`: The container engine does not have
+  `buildx` command (or BuildKit support) when building custom images.
+
```
Emilgardis, line 52: this should mention `CROSS_NO_WARNINGS`, forgot to document it in https://github.com/cross-rs/cross/pull/661

```suggestion
- `CROSS_CONTAINER_ENGINE_NO_BUILDKIT`: The container engine does not have
  `buildx` command (or BuildKit support) when building custom images.
- `CROSS_NO_WARNINGS`: Set to `1` to panic on warnings from `cross`,
  before building the executables.
  Use `0` to disable this behaviour.
  The no warnings behaviour is implicitly enabled in CI pipelines
```

Emilgardis (changes requested):
Awesome job! Some nits but looks great, thank you :)

This makes me also want to migrate to a mdbook, which we probably should've done from the start...

I'm wondering now though, we should probably edit the wiki with this

README.md
```diff
@@ -87,16 +81,22 @@ $ cross test --target mips64-unknown-linux-gnuabi64
...
+#### Option 1: Configuring `cross` directly in your `Cargo.toml`
 
-You can directly set [configuration values](docs/cross_toml.md) in your `Cargo.toml` file, under the `[package.metadata.cross]` table, i.e. key prefix.
-An example config snippet would look like this:
+You can directly set [configuration values][config_file] in your `Cargo.toml`
+file, under the `[package.metadata.cross]` table, i.e. key prefix. An example
+config snippet would look like this:
 
 ```toml,cargo
 [package.metadata.cross.target.aarch64-unknown-linux-gnu]
```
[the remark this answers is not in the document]
har7an: That page isn't loading for me. I assume that's expected as it doesn't exist yet?

docs/custom_images.md
```diff
@@ -0,0 +1,100 @@
...
+```toml
+[target.aarch64-unknown-linux-gnu]
+pre-build = [
+  "dpkg --add-architecture $CROSS_DEB_ARCH", 
+  "apt-get update && apt-get --assume-yes install libssl-dev:$CROSS_DEB_ARCH"
+]
+```
+
+Here, `CROSS_DEB_ARCH` will automatically evaluate to `arm64`, without you
+having to explicitly provide it.
```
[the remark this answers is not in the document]
har7an: Done, now I'm referencing the example further down in text which is already installing `libssl-dev`.

docs/config_file.md
har7an: I have these marks, too, but the rendered HTML is fine. I think it's safe to ignore this.

README.md
har7an: I can't quite follow you there. To me the diff looks normal, there aren't any whitespace changes in the C++ column. I checked and the text doesn't contain literal tab chars either. Can you clarify what's off here?

har7an:
In e636ee5 I've made the `cross_toml` module in the Rust sources public to eliminate some warnings during building the docs. Now the `cross_toml` module shows up in the crate documentation, pulling in the content of `docs/config_file.md` as its' module docs.

While I think it's nice to have that part of the docs mirrored there, unfortunately all the "external" links to other MD docs are broken there. How should we proceed? Imo we have several options:

1. Keep the current state with the broken links
2. Revert the changes, make `cross_toml` a private module again and remove the `#![doc]` directive completely
3. Use something like [rustdoc_assets][1] to copy the `docs/` folder as static assets into the generated docs, making the links in (1) work

Approach 3 seems to rely on build files, tho. I see you're already using xtask, but I have *no* idea to what extent we can influence how the docs on `docs.rs` are built, i.e. whether it always runs the same static command, or whether we can tell it to use xtask. Since afaik the docs are built without network access, though, my assumption is that using `xtask` (which has to compile itself) is out of scope... :(

[1]: https://docs.rs/rustdoc-assets/latest/rustdoc_assets/index.html

README.md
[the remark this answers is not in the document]
Emilgardis: Correct

README.md
Emilgardis: There is a space missing in c++ and test  when the cell does not contain a checkmark

Emilgardis: Screenshot_20240213-075818

Emilgardis: Actually, you're right, this is just a font issue where the checkmark is not mono

Emilgardis:
> > This makes me also want to migrate to a mdbook, which we probably should've done from the start...
> 
> Weeeell, I can work out a proposal if you like. In case I do step up to the task: Should we do that in a separate PR?

That can be done separately. I opened https://github.com/cross-rs/cross/issues/1446 to track it

> > I'm wondering now though, we should probably edit the wiki with this
> 
> Is it really worth the effort in the long run? Especially when moving towards `mdbook` for documentation, wouldn't it be easier for everyone if there were only one source of documentation for the entire project?

What I meant with editing the wiki is simply just to make it say on each page "see docs/<page> for documentation regarding ..."



> While I think it's nice to have that part of the docs mirrored there, unfortunately all the "external" links to other MD docs are broken there. How should we proceed? Imo we have several options:

I think we can just remove the doc from the rustdoc, it's not needed and we can instead just point to the files in the crate package. The only reason its done was because the cross_toml.md file used to be the module documentation for that file https://github.com/cross-rs/cross/commit/bc54c2a37fa0976c7047a20961a3179460d5c619#diff-a93d7b71ff09816ff274b195eaafc678ec01ab637abfe3ce49c0a14dfec32913

har7an:
Sorry for the slow responses!



> What I meant with editing the wiki is simply just to make it say on each page "see `docs/<page>` for documentation regarding ..."

I see. Will you do that once this is merged, or should I?


> I think we can just remove the doc from the rustdoc, it's not needed and we can instead just point to the files in the crate package.

Alright, got it. Now I have:

1. Made `cross_toml` a private module again and removed the `docs` directive from the top, so it has no module docs anymore.
2. Added a reference to the `docs/` directory in the repo to the "root" crate docs (next to README and wiki).
3. Fixed a remaining link pointing to `Cross.toml`, which now points to the `config_file.md` file instead.

What do you think?

src/lib.rs
```diff
@@ -899,7 +899,7 @@ macro_rules! commit_info {
     };
 }
 
-/// Obtains the [`CrossToml`] from one of the possible locations
+/// Obtains the [`CrossToml`][1] from one of the possible locations
 ///
 /// These locations are checked in the following order:
 /// 1. If the `CROSS_CONFIG` variable is set, it tries to read the config from its value
```
Emilgardis, line 905: Hmm, this shouldn't be a broken link, it's a [intra-doc link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html), please revert this

Emilgardis (changes requested):
Changes look great except the "broken" link being fixed.

Aside from that, I think this is ready for being merged, do you want to squash/fixup the commits you've marked as `squash!` or `fixup!`?

har7an:
@Emilgardis That first rebase was a little too eager, sorry... Changes are in now!

src/cross_toml.rs
```diff
@@ -1,4 +1,9 @@
-#![doc = include_str!("../docs/cross_toml.md")]
+//! The `Cross.toml` configuration file.
+//!
+//! For a detailed user documentation of the file and the contents please refer to the [docs in the
+//! repo][1].
+//!
+//! [1]: https://github.com/har7an/cross/blob/docs/restructure-docs-folder/docs/config_file.md
```
Emilgardis, line 6: wrong link!

Emilgardis (changes requested):

Emilgardis (approved):
Thank you for this! I'll edit the wiki to point to these documents

PR #1443, opened 2024-02-09
Docs: Change latest to main in dockerfile example

maxpowel:
The tag latest uses a very old image (with very old gcc and glib). The actual "latest" is called main. This little thing cost me several hours and this change will prevent other people to suffer the same.


Emilgardis:
I'm not sure if this should be merged. The reason is that you should use the tag associated with the cross version you have.

latest = v0.2.5
main/edge = last CI build

I'd highly recommend you to use `FROM $CROSS_BASE_IMAGE` instead, documented under this example.

I'd be open to change from latest to `0.2.5`, but not to `main`. I can be persuaded though :D

maxpowel:
You are right the solution is there, I guess that we can discard this PR then.

But from the point of view of someone new to cross, I would say that for me this part was a bit confusing.
`FROM $CROSS_BASE_IMAGE` is actually what I needed and I in my opinion this is the best and recommended option.
Maybe putting first the `good` solution would help for the new people who are not very sure about what are they doing (like me hehe).

Emilgardis:
I'm not sure why you changed the base branch for the pr. What I meant with `I'd be open to change from latest to 0.2.5, but not to main` was using `FROM ghcr.io/cross-rs/aarch64-unknown-linux-gnu:0.2.5`

PR #1449, opened 2024-02-23
fix error path when running pre-build-script 

key2p:
rt


Emilgardis (changes requested):
Oops! This looks good!

Can you remove the merge commit please

PR #1481, opened 2024-04-30
Allow environment variables to contain numbers

pizaCat:
When environment variable pass-through is used, a warning is displayed if the name of the variable contains a number and the value is taken from the host. For example:
```
[cross] warning: got environment variable of "MY_VAR_AES256" which is not a valid environment variable name. the proper syntax is `passthrough = ["ENVVAR=value"]`
```

The variable is still passed onto the cross container with the value from the host, but the warning is misleading.

### DIFF
--- a/.changes/1481.json
+++ b/.changes/1481.json
@@ -1,0 +1,4 @@
+{
+    "description": "allow pass-through environment variables to contain numbers",
+    "type": "fixed"
+}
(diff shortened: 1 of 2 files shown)


Emilgardis (approved):

Emilgardis:
Awesome, good catch. Feel free to add a changelog entry via `.changes`, altough maybe it's not needed

pizaCat:
I added the .changes entry, hopefully the filename is ok. It looked like the PR # was used for filename so I did that.

PR #1483, opened 2024-05-05
Fix paths when using `CROSS_CONTAINER_IN_CONTAINER`

thirteenowls:
Currently, `MountFinder` is used for almost all paths, rather than only for the host path in bind/volume mounts.

Because of this, `docker build` invocations use the wrong working directory (`/var/lib/docker/.../workdir-path` instead of `/workdir-path`), causing failures for builds that use `pre-build` with a relative file path, and `docker run` invocations use incorrect `-v` arguments (e.g. `/var/lib/docker/.../toolchain-dir:/var/lib/docker/.../toolchain-dir` instead of `/var/lib/docker/.../toolchain-dir:/toolchain-dir`).

This PR fixes those issues by only using `find_path` and `find_mount_path` when necessary.

### DIFF
--- a/.changes/1483.json
+++ b/.changes/1483.json
@@ -1,0 +1,4 @@
+{
+    "description": "Fix paths when using `CROSS_CONTAINER_IN_CONTAINER`",
+    "type": "fixed"
+}
--- a/src/docker/local.rs
+++ b/src/docker/local.rs
@@ -95,20 +95,25 @@ pub(crate) fn run(
         // Prevent `bin` from being mounted inside the Docker container.
         .args(["-v", &format!("{}/bin", toolchain_dirs.cargo_mount_path())]);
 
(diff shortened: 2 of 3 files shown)


thirteenowls:
I forgot about the changelog entry, sorry!

thirteenowls:
The Windows CI failure seems to be caused by a [recent commit](https://github.com/dtolnay/rust-toolchain/commit/18c33afdda3f4cf7a99f3b733af3df530f90c048) to `dtolnay/rust-toolchain`, which sets `CARGO_HOME`: https://github.com/cross-rs/cross/actions/runs/8956947724/job/24605204754?pr=1483#step:4:21

Emilgardis (approved):
Thanks!

Emilgardis:
failure is unrelated

PR #1485, opened 2024-05-07
Use `/proc/self/mountinfo` as a fallback for `docker inspect` if using `HOSTNAME` fails

### LINKED ISSUE #1321: Act: Build fails with `sh: 1: cargo: not found`
Hello, I've just finished the new version of my project. But when I did the PR an error occurred in the workflow. So I decided to revert the merge and tested the workflow locally with Act. Unfortunately, I discovered that, locally, I had one more error.
For Information MacOS isn't running with Act.

The cross version is the `v0.2.5`

The Error
```
| [command]cross build --release --target x86_64-pc-windows-gnu
| info: downloading component 'rust-src'
| info: installing component 'rust-src'
| sh: 1: cargo: not found
```
The Job:
```yml
deploy_building:
  name: Deploy Building
  needs: [ test ]
  strategy:
    matrix:
      os: [ ubuntu-latest ] # Default os for build
      target: [ aarch64-unknown-linux-gnu,
                i686-pc-windows-gnu, x86_64-pc-windows-gnu,
                i686-unknown-linux-gnu, x86_64-unknown-linux-gnu,
                x86_64-apple-darwin ]
      exclude: # Do not build for macOS on Linux
        - target: x86_64-apple-darwin
          os: ubuntu-latest
      include:
        - target: x86_64-apple-darwin
          os: macos-latest
  # Runs on latest ubuntu by default except for windows targets
  runs-on: ${{ matrix.os }}
  steps:
    - name: Checkout
      uses: actions/checkout@v1
    - name: Install rust
      uses: actions-rs/toolchain@v1
      with:
        toolchain: stable
        profile: minimal
        override: true
        target: ${{ matrix.target }}
    - name: Build release
      uses: actions-rs/cargo@v1
      with:
        use-cross: true
        command: build
        args: --release --target ${{ matrix.target }}
```

[Here ](https://github.com/cross-rs/cross/files/12454570/act_log_verbose_one_target.txt)  is the verbose log for only one target (windows x86_64)

Thanks to all the people who will try to solve my problem

thirteenowls:
Fixes #1321.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -1351,19 +1351,76 @@ pub fn get_image(
     Ok(image)
 }
 
+fn docker_inspect_self_mountinfo(engine: &Engine, msg_info: &mut MessageInfo) -> Result<String> {
+    if cfg!(not(target_os = "linux")) {
+        eyre::bail!("/proc/self/mountinfo is unavailable when target_os != linux");
+    }
+
+    // The ID for the current Docker container might be in mountinfo,
+    // somewhere in a mount root. Full IDs are 64-char hexadecimal
+    // strings, so the first matching path segment in a mount root
+    // containing /docker/ is likely to be what we're looking for. See:
+    // https://www.kernel.org/doc/Documentation/filesystems/proc.txt
+    // https://community.toradex.com/t/15240/4
+    let mountinfo = file::read("/proc/self/mountinfo")?;
+    let container_id = mountinfo
+        .lines()
+        .filter_map(|s| s.split(' ').nth(3))
+        .filter(|s| s.contains("/docker/"))
+        .flat_map(|s| s.split('/'))
+        .find(|s| s.len() == 64 && s.as_bytes().iter().all(u8::is_ascii_hexdigit))
+        .ok_or_else(|| eyre::eyre!("couldn't find container id in mountinfo"))?;
+
+    engine
+        .subcommand("inspect")
+        .arg(container_id)
+        .run_and_get_stdout(msg_info)
+}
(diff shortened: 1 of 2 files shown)


Emilgardis:
this is awesome! Isn't the description a bit misleading though? The fallback is actually docker now, not proc?

thirteenowls:
Oh, you're right, it is ambiguous. Does "Use `/proc/self/mountinfo` as a fallback if `docker inspect $HOSTNAME` fails" seem better to you?
(just in case, proc *is* the fallback, it first tries to use the hostname as usual and only reads mountinfo if that doesn't work)

Emilgardis:
I think that works!

src/docker/shared.rs
```diff
@@ -1351,19 +1351,75 @@ pub fn get_image(
...
+        if out.status.success() {
+            Ok(out.stdout()?)
+        } else {
+            let val = serde_json::from_slice::<serde_json::Value>(&out.stdout);
+            if let Ok(val) = val {
+                if let Some(array) = val.as_array() {
+                    // `docker inspect` completed but returned an empty array, most
+                    // likely indicating that the hostname isn't a valid container ID.
+                    if array.is_empty() {
+                        return docker_inspect_self_mountinfo(engine, msg_info);
```
Emilgardis, line 1401: can you add a msg_info.debug here similar to L1413

Emilgardis (changes requested):

thirteenowls:
Done!

Emilgardis (approved):

PR #1488, opened 2024-05-09
Allow `build-std` to take an array of crate names

### LINKED ISSUE #896: Update `build-std` to support a boolean or array.

thirteenowls:
Closes #896.

This is technically a breaking change because arbitrary strings passed to the `CROSS_BUILD_STD` environment variable (e.g. `yes`) are now no longer parsed as truthy values, but rather as a single crate name (or multiple, if containing commas) to pass to `-Zbuild-std` (resulting in e.g. `-Zbuild-std=yes`, which is invalid).
Behaviour remains the same when `CROSS_BUILD_STD` is set to `true`, `false`, or a number that fits in an `i32`.

### DIFF
--- a/.changes/1488.json
+++ b/.changes/1488.json
@@ -1,0 +1,6 @@
+{
+    "description": "Allow `build-std` to take an array of crate names",
+    "issues": [896],
+    "type": "changed",
+	"breaking": true
+}
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -161,10 +161,10 @@ targets:
 
 ```toml
 [target.aarch64-unknown-linux-gnu]
-build-std = false          # always build the std library. has precedence over xargo
-xargo = false              # disable the use of xargo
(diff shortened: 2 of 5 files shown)


Emilgardis (approved):
lgtm! thank you!

PR #1491, opened 2024-05-12
Allow specifying only a tag or subtarget for images in config

### LINKED ISSUE #1169: Specify only a tag for image to use
### Checklist

### Describe your request

> Is it possible in the `Cross.toml` to supply just the `tag` of the image and not the whole image? Might be a bit less error-prone as it will just take the github image associated with the target, but with a specific tag.

_Originally posted by @blu3beri in https://github.com/cross-rs/cross/issues/1167#issuecomment-1338969796_
      

This would make it easier to switch between versions.

for example one could do

```toml
# always use the image for the target with the tag `main`
image = ":main"
```

```toml
# always use the image for the target with the sha `DEADBEEF`
image = "@sha256:DEADBEEF"
```

optionally, we could also use this to switch to what we call a "subtarget", but not sure what the syntax would look for that.

thirteenowls:
Closes #1169.

Also supports specifying only a hash (`@sha256:...`) or subtarget (`-centos`).

### DIFF
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -234,6 +234,14 @@ image = ":edge"
 image = "@sha256:77db671d8356a64ae72a3e1415e63f547f26d374fbe3c4762c1cd36c7eac7b99"
 ```
 
+You can also specify a subtarget with no tag nor image name:
+
+```toml
+[target.x86_64-unknown-linux-gnu]
+# Translates to `ghcr.io/cross-rs/x86_64-unknown-linux-gnu:0.3.0-centos`
+image = "-centos"
+```
+
 The `image` key can also take the toolchains/platforms supported by the image:
 
 ```toml
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -221,6 +221,19 @@ the default one. Normal Docker behavior applies, so:
 - If only `image:latest` is specified, then Docker won't look in Docker Hub.
 - If the tag is omitted, then Docker will use the `latest` tag.
 
+If you specify a tag but no image name, `cross` will use the default image with
+the tag you provided:
+
+```toml
+[target.aarch64-unknown-linux-gnu]
+# Translates to `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge`
+image = ":edge"
+
+[target.x86_64-unknown-linux-musl]
+# Translates to `ghcr.io/cross-rs/x86_64-unknown-linux-musl@sha256:77db671d8356a64ae72a3e1415e63f547f26d374fbe3c4762c1cd36c7eac7b99`
+image = "@sha256:77db671d8356a64ae72a3e1415e63f547f26d374fbe3c4762c1cd36c7eac7b99"
+```
+
 The `image` key can also take the toolchains/platforms supported by the image:
 
 ```toml
--- a/src/docker/image.rs
+++ b/src/docker/image.rs
@@ -21,18 +21,23 @@ impl std::fmt::Display for Image {
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct PossibleImage {
-    pub name: String,
+    #[serde(rename = "name")]
+    pub reference: ImageReference,
     // The toolchain triple the image is built for
     pub toolchain: Vec<ImagePlatform>,
 }
 
 impl PossibleImage {
-    pub fn to_definite_with(&self, engine: &Engine, msg_info: &mut MessageInfo) -> Image {
+    pub fn to_definite_with(&self, engine: &Engine, msg_info: &mut MessageInfo) -> Result<Image> {
+        let ImageReference::Name(name) = self.reference.clone() else {
+            eyre::bail!("cannot make definite Image from unqualified PossibleImage");
+        };
+
         if self.toolchain.is_empty() {
-            Image {
-                name: self.name.clone(),
+            Ok(Image {
+                name,
                 platform: ImagePlatform::DEFAULT,
-            }
+            })
         } else {
             let platform = if self.toolchain.len() == 1 {
                 self.toolchain.first().expect("should contain at least one")
@@ -71,10 +76,10 @@ impl PossibleImage {
                     platform
                 }
             };
-            Image {
+            Ok(Image {
                 platform: platform.clone(),
-                name: self.name.clone(),
-            }
+                name,
+            })
         }
     }
 }
@@ -82,7 +87,7 @@ impl PossibleImage {
 impl<T: AsRef<str>> From<T> for PossibleImage {
     fn from(s: T) -> Self {
         PossibleImage {
-            name: s.as_ref().to_owned(),
+            reference: s.as_ref().to_owned().into(),
             toolchain: vec![],
         }
     }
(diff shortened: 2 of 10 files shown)


src/docker/shared.rs
```diff
@@ -1600,6 +1609,48 @@ mod tests {
...
+        };
+
+        let default_ver = format!(":{DEFAULT_IMAGE_VERSION}");
+        let mut map = HashMap::new();
+        test(map.clone(), &default_ver, &default_ver)?;
+
+        map.insert("CROSS_TARGET_X86_64_UNKNOWN_LINUX_GNU_IMAGE", ":edge");
+        test(map.clone(), ":edge", ":edge")?;
+
+        // `image` always takes precedence over `zig.image`, even when `uses_zig` is `true`
```
thirteenowls, line 1641: This is the current behaviour, which I left as-is, but it seems somewhat counter-intuitive to me?

src/docker/image.rs
```diff
@@ -27,6 +27,12 @@ pub struct PossibleImage {
 }
 
 impl PossibleImage {
+    pub fn target_is_specified(&self) -> bool {
+        // Hostnames cannot start with either of these characters,
+        // nor can image names.
+        !(self.name.starts_with(':') || self.name.starts_with('@'))
+    }
+
```
Emilgardis, line 35: I don't think I like this, I think it would be better to actually deserialize differently, e.g represent the data as


```rust
pub struct PossibleImage {
    pub reference: ImageReference,
    pub toolchain: Vec<ImagePlatform>,
}
pub enum ImageReference {
   // partially qualified reference. with or without tag/digest  
   Name(String),
   /// unqualified reference, only a tag or digest
   Identifier(String),
}
```

However, I don't have a very strong opinion about it and it would be possible to fix afterhand.

I do think that the name of this function should be changed. `image_name_is_specified`

src/docker/mod.rs
```diff
@@ -31,6 +31,10 @@ impl ProvidedImage {
     pub fn image_name(&self, repository: &str, tag: &str) -> String {
         image_name(self.name, self.sub, repository, tag)
     }
+
+    pub fn default_image_name(&self) -> String {
+        self.image_name(CROSS_IMAGE, DEFAULT_IMAGE_VERSION)
+    }
```
Emilgardis, line 37: is this function really needed? I feel like it could be done inplace.

src/docker/shared.rs
Emilgardis: I think you're correct. If zig is enabled then `zig.image` should be used

Emilgardis (changes requested):
Some small notes! Good thinking about deduping get_image/get_image_name, it's been a concern of mine!

src/docker/shared.rs
thirteenowls: Good to know. Since it's a breaking change, I'll file a separate PR to fix it once this is merged.

src/docker/image.rs
thirteenowls: I agree, I'll address this.

src/docker/mod.rs
Emilgardis: my only concern with it is that in context it's a bit hard to understand what default_image_name actually is.

thirteenowls: I apologize, I'm not sure I understand what you mean by that. 😅
I think that without it, you'd still have to look at the definition of `image_name` and the two consts, so in that respect I don't think it adds much mental overhead. I'm not sure that's what you meant, though.

Emilgardis: you're right, lets leave the function!

Emilgardis:
one thing I just thought about, would it be possible to also support a syntax like `-centos`, which would pick `{default_image}:{default_tag}-centos`?

docs/config_file.md
```diff
@@ -221,6 +221,27 @@ the default one. Normal Docker behavior applies, so:
...
+[target.x86_64-unknown-linux-musl]
+# Translates to `ghcr.io/cross-rs/x86_64-unknown-linux-musl@sha256:77db671d8356a64ae72a3e1415e63f547f26d374fbe3c4762c1cd36c7eac7b99`
+image = "@sha256:77db671d8356a64ae72a3e1415e63f547f26d374fbe3c4762c1cd36c7eac7b99"
+```
+
+You can also specify a subtarget with no tag nor image name:
+
+```toml
+[target.x86_64-unknown-linux-gnu]
+# Translates to `ghcr.io/cross-rs/x86_64-unknown-linux-gnu:0.3.0-centos`
```
thirteenowls, line 241: I'm still not sure which tag should be shown as the default there. I didn't want to give users the idea that `main` is always the default, so I picked what should be the first tagged release with this feature.

thirteenowls:
I've pushed the requested change, as well as subtarget support. :)
Please let me know what you think, I'm not entirely confident about some bits.

docs/config_file.md
Emilgardis: that's good, could add a release replacement [here](https://github.com/cross-rs/cross/blob/53a45dbf5819ebc86e0c71152757fcdad7e81d51/Cargo.toml#L77) to automatically update

src/docker/image.rs
```diff
@@ -21,18 +21,23 @@ impl std::fmt::Display for Image {
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct PossibleImage {
-    pub name: String,
+    #[serde(rename = "name")]
+    pub reference: ImageReference,
```
Emilgardis, line 30: hmm, not a huge fan of this, maybe it would make sense to do 

```rs
    #[serde(alias = "name")]
    pub reference: ImageReference,
```

and we steer towards using `reference` when using `-sub` or `:tag`, could possible even make it an error to use `name` with this new feature.

Emilgardis (changes requested):
one small change needed I think, otherwise this looks great

docs/config_file.md
thirteenowls: Oh neat, I didn't know about this! Will do.

src/docker/image.rs
thirteenowls: Hm, I feel like that might make it confusing? I'm concerned about there being two ways of doing the same thing.
I'm partial to just keeping `name` as I find it clearer, but if we want to enforce using a different config key for this feature, IMO we should go all-out and add `tag`, `digest`, and `subtarget` (or something to that effect).

Emilgardis: that makes sense to me, it should then also be enforced that only one of those are applicable.

I think `tag`, `digest` and `subtarget` is good, it also makes it seem less "magical"

thirteenowls: Sounds good, I'll do that then. I assume we're still allowing the "magic" syntax in short form (e.g. `image = ":tag"`)?

Emilgardis: No, I don't think it should be allowed

Emilgardis: Or actually, scratch that, I'm not sure what I think

thirteenowls: I thought this over once more, here's how I feel about the options:

1. Only having the special syntax (as `name`) is the most convenient option, but it does look "magical", which I don't like either. However, I think that sufficiently documenting the feature can mostly counteract this. It also is relatively intuitive, as the syntax isn't really new: it's a normal image name, with the part you want "autocompleted" removed.

2. Having both `name` and `reference` feels unnecessary if they're just aliases. Requiring the use of `reference` for the special syntax makes more sense, but I think that decoupling it from `name` makes it seem *more* magical, and therefore confusing.

3. Adding `tag`, `digest`, and `subtarget` has the large benefit of avoiding magic and being very explicit, especially if the special syntax is removed from the `image` shortform. It seemed right to me, but after thinking it over, I worry that having four (those three + `name`) mutually exclusive config keys (out of five in `image`) would be confusing/a bad user experience. They could also be misinterpreted as the exact opposite, i.e. `name = "foo"` + `tag = "bar"` being equivalent to `image = "foo:bar"` (although sufficient documentation should take care of this as well). Anecdotally, this option would require adding mutually exclusive `*_[ZIG_]IMAGE_{TAG,DIGEST,SUBTARGET}` environment variables.

4. A fourth option I considered would be to only keep `name`, but introduce new syntax that uses the name of what's being specified (e.g. `image = "subtarget=centos"`). It solves the mutual exclusivity issue, but is prone to typos and seems a bit confusing, so I don't think I like it.

Emilgardis: Thanks for the writeup! let's keep it as is for now and if it becomes a problem or needs clarification it can be expanded on without breakage

Emilgardis:
Can we merge this?

thirteenowls:
I believe so! Thank you for taking the time to review my PRs. :)

PR #1494, opened 2024-05-14
Fix `zig.image` precedence

thirteenowls:
Currently, `image` always takes precedence over `zig.image`.
This change makes it so `image` doesn't get used when zig is enabled, even if `zig.image` is unset (and vice versa).

This is a breaking change.


Emilgardis (approved):
Thank you!

PR #1513, opened 2024-06-13
chore(deps): Update dependencies

reneleonhardt:
### Features
* Update dependencies with `cargo upgrade`
* Update GitHub Actions manually (others can be merged thanks to [Dependabot #1464](https://github.com/cross-rs/cross/pull/1464))
* Configure Dependabot to update Rust, Python and Git too

### Recommendations
* Merge Dependabot updates or [configure automerge](https://stackoverflow.com/questions/64116781/how-do-i-automerge-dependabot-updates-config-version-2)
* Release more often to fix security vulnerabilities for your users (unfixed in v0.2.5)

| Library | Vulnerability | Severity |
|-|-|-|
| remove_dir_all | GHSA-mc8h-8q98-g5hr | LOW |


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+- Update dependencies
+
```
Emilgardis, line 11: Please do not edit this changelog directly, instead see [.changes](https://github.com/cross-rs/cross/tree/main/.changes#readme)

.github/dependabot.yml
```diff
@@ -48,3 +48,35 @@ updates:
...
+        patterns:
+          - "*"
+  - package-ecosystem: "cargo"
+    directory: "/xtask"
+    schedule:
+      interval: "weekly"
+    groups:
+      rust_dependencies:
+        patterns:
+          - "*"
```
Emilgardis, line 66: Would it be possible to only do a pr/update when a (cargo) semver break is done? e.g the first non-zero digit is changed. Reason is that I'm not sure we need to always be up to date, I can be persuaded about this though, I've just never seen the reason to keep track of the changes so closely.

.github/dependabot.yml
```diff
@@ -48,3 +48,35 @@ updates:
...
+  - package-ecosystem: "pip"
+    directory: "/docker/android"
+    schedule:
+      interval: "weekly"
+    groups:
+      python_dependencies:
+        patterns:
+          - "*"
```
Emilgardis, line 82: Ditto for this, I'd rather do this less often or only when a break happens.

.github/dependabot.yml
Emilgardis: Maybe changing it to be done monthly?

Emilgardis (changes requested):
Thank you for the initiative! I like it, I just have some concerns about the frequency.

Also a note abot CHANGELOG.md

PR #1524, opened 2024-07-08
Fixed a typo

adamperkowski:
In Installation section of README
'using' -> 'use'

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ cargo install cross --git https://github.com/cross-rs/cross
 ```
 
 It's also possible to directly download the pre-compiled [release
-binaries](https://github.com/cross-rs/cross/releases) or using
+binaries](https://github.com/cross-rs/cross/releases) or use
 [cargo-binstall](https://github.com/cargo-bins/cargo-binstall).
 
 ## Usage


Emilgardis:
is this actually a typo?

to me `using` makes sense in this context

adamperkowski:
Well, `using` in a sentence should be followed by a verb to actually tell what are you using. For example:
> She is using a pen to write something.

The correct structure would be to use `use` after `or` to maintain parallel structure in the sentence. ([just looked that up](https://www.oxfordlearnersdictionaries.com/definition/english/use_1))

`... to directly download the pre-compiled release binaries or to use ...` would be correct too.

Emilgardis:
the intended meaning of the `or` is 

"You can also install cross via the precompiled binaries from our github releases or use cargo-binstall"

maybe that's a better way to phrase it?

Emilgardis:
if you disagree with that i'm fine accepting this pr :)

adamperkowski:
I don't disagree. I understand the meaning. In fact,
> You can also install cross via the precompiled binaries from our github releases or use cargo-binstall

you did the same thing as me here, just changed the first part of the sentence (`download` -> `install cross via`) and you used `use`. 
So I don't disagree, I'm fine with you accepting this pr or just changing it by yourself.

Emilgardis:
Alright, lets do it this way then! Thank you!

Emilgardis (approved):

PR #1569, opened 2024-10-07
Fix in docker_inspect_self_mountinfo() fo Gitlab runner

zweikaesehoch:
Crossbuilding in our buildimage on a Gitlab runner failed:
image

Container ID could not be resolved from /proc/self/mountinfo, because "/docker/" string was expected, but "/docker-data/" provided:
image

I extended string matching with "/docker-data/" and our pipeline is working again.

### DIFF
--- a/src/docker/shared.rs
+++ b/src/docker/shared.rs
@@ -1364,7 +1364,7 @@ fn docker_inspect_self_mountinfo(engine: &Engine, msg_info: &mut MessageInfo) ->
     let container_id = mountinfo
         .lines()
         .filter_map(|s| s.split(' ').nth(3))
-        .filter(|s| s.contains("/docker/"))
+        .filter(|s| s.contains("/docker/") || s.contains("/docker-data/")) // Check for both "/docker/" and "/docker-data/"
         .flat_map(|s| s.split('/'))
         .find(|s| s.len() == 64 && s.as_bytes().iter().all(u8::is_ascii_hexdigit))
         .ok_or_else(|| eyre::eyre!("couldn't find container id in mountinfo"))?;


Emilgardis (approved):
Thanks!

zweikaesehoch:
We further investigated the error which led to our custom data-root setting in /etc/docker/daemon.json:
image

The fallback implemented due to https://github.com/cross-rs/cross/issues/1321 misses the container id because of our changed data-root to "docker-data" instead of default "docker".

The real issue here is, that docker_inspect_self_mountinfo() only works, if data-root setting contains "/docker/".

The current PR bears no value for other users now.
One would need to parse daemon.json data-root setting to make the fallback more robust. 

Pseudocode:
Look into /etc/docker/daemon.json
if data-root is set
  extract last path element as DATA_ROOT
  if DATA_ROOT not equals "/docker"
    add DATA_ROOT to OR-filter strings in line 1367

andrejohansson:
This is currently blocking me and I have no authority over our agents. Is there something I can do as a workaround?

Emilgardis:
> This is currently blocking me and I have no authority over our agents. Is there something I can do as a workaround?

if you add `-v` do you get any debug messages about `docker inspect found no containers matching HOSTNAME, retrying using mountinfo`? if no, try setting the `HOSTNAME` envvar to the name of a container

I think I have a workaround, I'll create a new PR and then you and @zweikaesehoch could maybe try it

arakholia:
I have also run into this issue attempting to run cross on GitHub Codespaces. It seems local devcontainers use `/var/lib/docker/containers/` while on the Codespaces it uses just `/containers/`. The following pattern match (similar to this PR) resolved the issue and seems to handle both uses cases:

```rust
.filter(|s| s.contains("/docker/") || s.contains("/containers/"))```
```

although there may be a better way.

Emilgardis:
Please give #1588 a try 👍🏼

andrejohansson:
My workaround was to just use the cross-rs images directly as the base image in CI. No docker-in-docker.

PR #1572, opened 2024-10-10
Remove a dbg!

SomeoneToIgnore:
After following the installation step at https://github.com/cross-rs/cross?tab=readme-ov-file#installation 
I see the following output when running the commands:
```
❯ ccrossbuild --target aarch64-unknown-linux-gnu

[src/lib.rs:944:30] workspace_metadata.get("cross") = None
``` 

which looks like a leftover.


Emilgardis (approved):
Curious! Thanks!

PR #1574, opened 2024-10-15
Require dockerfile to be specified relative to cargo workspace root

ziutech:
The docs say that [the dockerfile should be specified relative to Cargo.toml](https://github.com/cross-rs/cross/blob/main/docs/config_file.md#targettargetdockerfile), but actually it is specified [relative to current working directory](https://github.com/cross-rs/cross/blob/main/src/docker/custom.rs#L150-L153). This is confusing. 

This PR makes it so that dockerfiles are always relative to workspace root.

### DIFF
--- a/src/docker/custom.rs
+++ b/src/docker/custom.rs
@@ -120,7 +120,9 @@ impl<'a> Dockerfile<'a> {
         }
 
         let path = match self {
-            Dockerfile::File { path, .. } => PathBuf::from(path),
+            Dockerfile::File { path, .. } => {
+                paths.metadata.workspace_root.join(PathBuf::from(path))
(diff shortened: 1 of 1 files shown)


Emilgardis:
Thanks for spotting this! This was definitely a mistake. Could you add a changelog entry as such

add a file `1574.json` in `.changes/`
```json
{
    "type": "fixed",
    "description": "`dockerfile` path is relative to cargo workspace root",
}
```

ziutech:
I was thinking whether this should be marked as a "change", because this might break someone's setup, but "fix" is probably correct.

ziutech:
Thanks!

PR #1578, opened 2024-10-23
Update owo-colors

jayvdb:
This removes one of the atty dependencies

### DIFF
--- a/src/shell.rs
+++ b/src/shell.rs
@@ -27,7 +27,7 @@ macro_rules! write_style {
             ColorChoice::Auto => write!(
                 $stream,
                 "{}",
-                $message $(.if_supports_color($stream.owo(), |text| text.$style()))*
+                $message $(.if_supports_color($stream.owo().unwrap(), |text| text.$style()))*
             ),
         }?;
     }};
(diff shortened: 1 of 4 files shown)


src/shell.rs
```diff
@@ -451,19 +451,19 @@ fn get_verbosity(
...
-    fn owo(&self) -> owo_colors::Stream {
+    fn owo(&self) -> Option<owo_colors::Stream> {
         Self::OWO
     }
 }
 
 impl Stream for io::Stdin {
     type TTY = io::Stdin;
-    const OWO: owo_colors::Stream = owo_colors::Stream::Stdin;
+    const OWO: Option<owo_colors::Stream> = None;
```
jayvdb, line 466: Another option is to do `const OWO: owo_colors::Stream = owo_colors::Stream::Stdout;` here even though that is wrong, but it avoids the `Option` and the `unwrap()`.

Or split the trait.  I was a bit worried as it is `pub`.  OTOH, probably nobody else is using the lib.

Emilgardis: Breaking the api is fine. We dont guarantee that its stable

https://github.com/cross-rs/cross/blob/4090beca3cfffa44371a5bba524de3a578aa46c3/src/lib.rs#L8

neuschaefer:
Superseded by #1708, I believe.

Issue #1518, opened 2024-06-27
Building aarch64 image fails due to gfortran not being available


astapleton:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

Building the aarch64 Docker image on Mac M1/M2 fails due to a missing package error for gfortran. ie. running this:
```
docker build -f Dockerfile.aarch64-unknown-linux-gnu -t ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main https://github.com/cross-rs/cross.git#main:docker
```
results in this error:
```
Unable to locate package gfortran-aarch64-linux-gnu
```
which then fails the image creation.

It seems like this was introduced in this commit: https://github.com/cross-rs/cross/commit/6ab0e7cfc8fd45e6947b0c2768c041ad4ba0c528 

I also reproduced this on an x86_64 Windows host when specifying `--platform linux/arm64` during the build step.

### Which operating system is the host (e.g computer cross is on) running?

- [X] macOS
- [X] Windows
- [ ] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [X] x86_64 / AMD64
- [ ] arm32
- [X] arm64 (including Mac M1)

### What container engine is cross using?

- [X] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.5

### Example

```
docker build -f Dockerfile.aarch64-unknown-linux-gnu -t ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main https://github.com/cross-rs/cross.git#main:docker
```

### Additional information / notes

Building with an older commit works:
```
docker build -f Dockerfile.aarch64-unknown-linux-gnu -t ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main https://github.com/cross-rs/cross.git#c87a52a60dbcde069714a8d2ed51570e7fc23cf9:docker
```

wenxb:
Seems to be installing `gfortran`

astapleton:
Still fails for me:
```
2.843 E: Unable to locate package gfortran-aarch64-linux-gnu
------
Dockerfile.aarch64-unknown-linux-gnu:15
--------------------
  14 |
  15 | >>> RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
  16 | >>>     g++-aarch64-linux-gnu \
[...]

PR #1597, opened 2024-12-18
Support `linux/arm64/v8` for most Ubuntu-based images

ianks:
Previously, many images only provided `linux/amd64` builds, which meant that Apple silicon folks were left to Rosetta or qemu emulation when cross-compiling. Although Rosetta's `x86_64` emulation is a remarkable feat of engineering, nothing beats the performance of using the `linux/arm64/v8` on Apple silicon.

This PR adds support to a bunch of Ubuntu-based images for `linux/arm64/v8` builds. I haven't been able to build all of them yet... mostly just tested the `x86_64-unknown-linux-gnu` container. But I plan to squash any issues that arise from CI. Theoretically, it shouldn't be too hard since the core logic is the same across all images.

### New Platforms Supported

- `aarch64-unknown-linux-gnu`
- `armv5te-unknown-linux-gnueabi`
- `armv7-unknown-linux-gnueabi`
- `i586-unknown-linux-gnu`
- `i686-unknown-linux-gnu`
- `mips-unknown-linux-gnu`
- `mips64-unknown-linux-gnuabi64`
- `mips64el-unknown-linux-gnuabi64`
- `mipsel-unknown-linux-gnu`
- `powerpc-unknown-linux-gnu`
- `powerpc64-unknown-linux-gnu`
- `powerpc64le-unknown-linux-gnu`
- `riscv64gc-unknown-linux-gnu`
- `s390x-unknown-linux-gnu`
- `sparc64-unknown-linux-gnu`
- `thumbv7neon-unknown-linux-gnueabihf`
- `x86_64-unknown-linux-gnu`

### DIFF
--- a/docker/Dockerfile.aarch64-unknown-linux-gnu
+++ b/docker/Dockerfile.aarch64-unknown-linux-gnu
@@ -12,10 +12,11 @@ RUN /xargo.sh
 
 FROM cross-base as build
 
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-aarch64-linux-gnu \
-    gfortran-aarch64-linux-gnu \
-    libc6-dev-arm64-cross
+ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-gnu-
+ENV CROSS_SYSROOT=/usr/aarch64-linux-gnu
+
+COPY apt-cross-essential.sh /
+RUN TARGET_ARCH=arm64 TARGET_TRIPLE=aarch64-linux-gnu /apt-cross-essential.sh
 
 COPY deny-debian-packages.sh /
 RUN TARGET_ARCH=arm64 /deny-debian-packages.sh \
@@ -34,8 +35,6 @@ RUN /linux-image.sh aarch64
 COPY linux-runner base-runner.sh /
 COPY toolchain.cmake /opt/toolchain.cmake
 
-ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-gnu-
-ENV CROSS_SYSROOT=/usr/aarch64-linux-gnu
 ENV CROSS_TARGET_RUNNER="/linux-runner aarch64"
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="$CROSS_TARGET_RUNNER" \
--- a/docker/Dockerfile.arm-unknown-linux-gnueabi
+++ b/docker/Dockerfile.arm-unknown-linux-gnueabi
@@ -15,7 +15,7 @@ FROM cross-base AS build
 ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-gnueabi-
 ENV CROSS_SYSROOT=/usr/arm-linux-gnueabi
 
-COPY essential.sh /
+COPY apt-cross-essential.sh /
 RUN TARGET_ARCH=armel TARGET_TRIPLE=arm-linux-gnueabi /essential.sh
 
 COPY deny-debian-packages.sh /
--- a/docker/Dockerfile.arm-unknown-linux-gnueabi
+++ b/docker/Dockerfile.arm-unknown-linux-gnueabi
@@ -10,12 +10,13 @@ RUN /cmake.sh
 COPY xargo.sh /
 RUN /xargo.sh
 
-FROM cross-base as build
+FROM cross-base AS build
 
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-arm-linux-gnueabi \
-    gfortran-arm-linux-gnueabi \
-    libc6-dev-armel-cross
+ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-gnueabi-
+ENV CROSS_SYSROOT=/usr/arm-linux-gnueabi
+
+COPY essential.sh /
+RUN TARGET_ARCH=armel TARGET_TRIPLE=arm-linux-gnueabi /essential.sh
 
 COPY deny-debian-packages.sh /
 RUN TARGET_ARCH=armel /deny-debian-packages.sh \
@@ -28,8 +29,6 @@ RUN /qemu.sh arm
 COPY qemu-runner base-runner.sh /
 COPY toolchain.cmake /opt/toolchain.cmake
 
-ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-gnueabi-
-ENV CROSS_SYSROOT=/usr/arm-linux-gnueabi
 ENV CROSS_TARGET_RUNNER="/qemu-runner arm"
 ENV CARGO_TARGET_ARM_UNKNOWN_LINUX_GNUEABI_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CARGO_TARGET_ARM_UNKNOWN_LINUX_GNUEABI_RUNNER="$CROSS_TARGET_RUNNER" \
--- a/docker/Dockerfile.sparc64-unknown-linux-gnu
+++ b/docker/Dockerfile.sparc64-unknown-linux-gnu
@@ -12,8 +12,14 @@ RUN /xargo.sh
 
 FROM cross-base as build
 
+ENV CROSS_TOOLCHAIN_PREFIX=sparc64-linux-gnu-
+ENV CROSS_SYSROOT=/usr/sparc64-linux-gnu
+
+COPY apt-cross-essential.sh /
+RUN TARGET_ARCH=sparc64 TARGET_TRIPLE=sparc64-linux-gnu /apt-cross-essential.sh
+
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-sparc64-linux-gnu \
+    g++- \
     gfortran-sparc64-linux-gnu \
     libc6-dev-sparc64-cross
 
@@ -34,8 +40,6 @@ RUN /linux-image.sh sparc64
 COPY linux-runner base-runner.sh /
 COPY toolchain.cmake /opt/toolchain.cmake
 
-ENV CROSS_TOOLCHAIN_PREFIX=sparc64-linux-gnu-
-ENV CROSS_SYSROOT=/usr/sparc64-linux-gnu
 ENV CROSS_TARGET_RUNNER="/linux-runner sparc64"
 ENV CARGO_TARGET_SPARC64_UNKNOWN_LINUX_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc \
     CARGO_TARGET_SPARC64_UNKNOWN_LINUX_GNU_RUNNER="$CROSS_TARGET_RUNNER" \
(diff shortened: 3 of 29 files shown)


docker/Dockerfile.aarch64-unknown-linux-gnu
```diff
@@ -12,10 +12,11 @@ RUN /xargo.sh
 
 FROM cross-base as build
 
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-aarch64-linux-gnu \
-    gfortran-aarch64-linux-gnu \
-    libc6-dev-arm64-cross
+ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-gnu-
+ENV CROSS_SYSROOT=/usr/aarch64-linux-gnu
```
ianks, line 16: I decided to move these variable above `apt-cross-essential.sh`, since I wanted to use `CROSS_TOOLCHAIN_PREFIX` to test for pre-installed gfortran (which was causing some headaches).

docker/linux-image.sh
```diff
@@ -156,7 +161,7 @@ main() {
         ;;
     x86_64)
         arch=amd64
-        kernel="${kversion}-amd64"
+        kernel="${kmajor}\.${kminor}\.${kpatch}-.*-amd64"
```
ianks, line 164: I don't know why, but apt was unable to find the original `5.10.0-26-amd64` package. Maybe it disappeared or something? Anyway, this fixed it and unblocked me

docker/qemu.sh
```diff
@@ -19,7 +19,7 @@ build_static_libffi () {
     tar --strip-components=1 -xzf "v${version}.tar.gz"
     ./configure --prefix="$td"/lib --disable-builddir --disable-shared --enable-static
     make "-j$(nproc)"
-    install -m 644 ./.libs/libffi.a /usr/lib64/
+    install -m 644 ./.libs/libffi.a /usr/local/lib/
```
ianks, line 22: `/usr/lib64` is not a thing on `arm64`, so just use `/usr/local/lib` instead. This potentially could cause issues with folks who building of their old Pentium 4's, but I'm not sure. Either way, I'd reckon it's a small subset of the userbase anyway.

Emilgardis:
This is awesome! Thank you, its all looking very good.

We have a problem though with how to publish the arm64 images, I laid out a strategy in https://github.com/cross-rs/cross/pull/849 but we could also use imagetools. We'd also probably want arm64 runners, but they are only available for team/enterprise orgs :/

The current ci wont set the platform correctly, itll alwo overwrite the uploaded images to not be multi-arch https://github.com/cross-rs/cross/blob/4090beca3cfffa44371a5bba524de3a578aa46c3/.github/workflows/ci.yml#L192

Ill have a look into this pr more closely

Emilgardis:
Actually, not true, it should build but would overwrite, lets do a try :D 
https://github.com/cross-rs/cross/blob/4090beca3cfffa44371a5bba524de3a578aa46c3/xtask/src/build_docker_image.rs#L179

/ci try --target aarch64-unknown-linux-gnu --target x86_64-unknown-linux-gnu

ianks:
@Emilgardis Ok I think I see the issue, `linux-image-5.10.0-26-*` is gone due to some security patch, so just bumping to `5.10.0-27` should resolve the issue. Can you try building again?

ianks:
> it should build but would overwrite

My (very limited) understanding was the `buildx` should automagically deal with pushing to the correct destination for each platform (so docker CLI can pull the right image later)... Does that sound right?

ianks:
/ci try --target aarch64-unknown-linux-gnu --target x86_64-unknown-linux-gnu

docker/linux-image.sh
```diff
@@ -424,7 +424,7 @@ EOF
 
     # need to reinstall the removed libgcc packages, which are required for apt
     if [[ "${arch}" == "${dpkg_arch}" ]]; then
-        apt-get install --no-install-recommends --assume-yes "${packages[@]}"
+        apt-get install --no-install-recommends --assume-yes "${libgcc_packages[@]}"
```
ianks, line 427: I stumbled upon this... which I assumed to have been a typo since `packages` is likely not defined (its only used for wildcard kernel image searches).

Emilgardis:
/ci try --target aarch64-unknown-linux-gnu --target x86_64-unknown-linux-gnu

Emilgardis:
> My (very limited) understanding was the `buildx` should automagically deal with pushing to the correct destination for each platform (so docker CLI can pull the right image later)... Does that sound right?

That sounds right, but it doesn't mean that both the amd64 and arm64/v8 images would be available, only the last built one would be tagged with what we want.

ianks:
> > My (very limited) understanding was the `buildx` should automagically deal with pushing to the correct destination for each platform (so docker CLI can pull the right image later)... Does that sound right?
> 
> That sounds right, but it doesn't mean that both the amd64 and arm64/v8 images would be available, only the last built one would be tagged with what we want.

I took a look, and I _think_ https://github.com/cross-rs/cross/pull/1597/commits/64283325b840e3c3ebfa6d9e7d5b952a8a63925b may do the trick. Basically, it just expands the matrix to include `platform` as a dimension, and passes the `--platform` to `build-docker-image`. Am I on the right track here???

Emilgardis:
That works, but it doesnt solve the problem of how to consolidate the two builds into one image

Emilgardis:
I have no preference, just want something that works. I've not heard of buildx supporting multiple platforms and then combining it, that's interesting and would be useful

Emilgardis:
Thinking about this a bit more, using the multi value --platform would work for all images, except when target=host, as we want to then use `Dockerfile.native` instead

ianks:
> Thinking about this a bit more, using the multi value --platform would work for all images, except when target=host, as we want to then use Dockerfile.native instead

Ok, I'll give the multi-platform string a whirl. The only thing I'm unsure about is if we need to supply the `--push` parameter for it to work. If so, that could be a nuisance because I think we want to test before actually pushing 🤔

Emilgardis:
Dont think thats needed, i only see that the `docker-container` driver is needed, which we should be using already. You should revert https://github.com/cross-rs/cross/pull/1597/commits/64283325b840e3c3ebfa6d9e7d5b952a8a63925b i think

ianks:
> Dont think thats needed, i only see that the `docker-container` driver is needed, which we should be using already. You should revert [6428332](https://github.com/cross-rs/cross/commit/64283325b840e3c3ebfa6d9e7d5b952a8a63925b) i think

To clarify, you _do_ want to use multiple platforms per job correct? Because before [6428332](https://github.com/cross-rs/cross/commit/64283325b840e3c3ebfa6d9e7d5b952a8a63925b), we were only passing one platform ([see here](https://github.com/cross-rs/cross/pull/1597#issuecomment-2554951496)). If so, I can proceed with revert and work on passing multiple platforms per job afterwards.

Emilgardis:
> To clarify, you _do_ want to use multiple platforms per job correct? Because before [6428332](https://github.com/cross-rs/cross/commit/64283325b840e3c3ebfa6d9e7d5b952a8a63925b), we were only passing one platform ([see here](https://github.com/cross-rs/cross/pull/1597#issuecomment-2554951496)). If so, I can proceed with revert and work on passing multiple platforms per job afterwards.

Interesting, that is not the intent of the code, it should build all passed platforms in one job per target in sequence. I don't think it's easy to do one job per target and platform, as we'd have to pass the built image somewhere else, however, as I [mentioned here](https://github.com/cross-rs/cross/pull/849#issuecomment-1304651603) I think the most correct approach is publishing the image as `<target>:<version>-<platform>` and then consolidate the new `<target>:<version>-*` images into one image tagged `<target>:<version>`

ianks:
@Emilgardis If possible, I'd like to get this PR merged as is then we can handle the CI aspect of it as a follow up if you're OK with that. I'm struggling to figure out the best way to do it and I don't want to balloon the scope of this PR any more

Emilgardis:
That's alright! The problem with merging it as is is that this will make people on aarch64 try to use an image that is not published. If you remove the changes to targets.toml and update provided_images.rs we can merge

ianks:
Done. I've pulled out those changes to a new branch to pick up after: https://github.com/oxidize-rb/cross/compare/more-arm64-docker-builds...oxidize-rb:cross:more-arm64-docker-builds-targets-branch?expand=1

Emilgardis (approved):
Thank you!

Emilgardis:
can ignore the centos failure, the other errors need to be looked at, seems like there's .sh files missing?

ianks:
Ok I _think_ I fixed it.

Emilgardis:
I don't see a new commit, maybe you forgot to push? :D

docker/Dockerfile.sparc64-unknown-linux-gnu
```diff
@@ -12,8 +12,14 @@ RUN /xargo.sh
...
+ENV CROSS_SYSROOT=/usr/sparc64-linux-gnu
+
+COPY apt-cross-essential.sh /
+RUN TARGET_ARCH=sparc64 TARGET_TRIPLE=sparc64-linux-gnu /apt-cross-essential.sh
+
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-sparc64-linux-gnu \
+    g++- \
     gfortran-sparc64-linux-gnu \
     libc6-dev-sparc64-cross
```
Emilgardis, line 24: missed this while reviewing, this seems wrong

Emilgardis:
@ianks did you manage to figure this out? If not and you're not able to continue I'd be happy to take over/help

astapleton:
Excited for this to merge. I was just looking into #1518 and this looks like it will address all the issues I experienced building an image.

docker/Dockerfile.arm-unknown-linux-gnueabi
```diff
@@ -10,12 +10,13 @@ RUN /cmake.sh
...
 
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-arm-linux-gnueabi \
-    gfortran-arm-linux-gnueabi \
-    libc6-dev-armel-cross
+ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-gnueabi-
+ENV CROSS_SYSROOT=/usr/arm-linux-gnueabi
+
+COPY apt-cross-essential.sh /
+RUN TARGET_ARCH=armel TARGET_TRIPLE=arm-linux-gnueabi /essential.sh
```
garasubo, line 19: it looks apt-cross-essential.sh no longer exists

garasubo:
@Emilgardis @ianks This PR looks great. I encountered this issue on my arm64 PC and was looking for a solution like this. Do you have any updates on this? It seems the last blocker is the build failure [here](https://github.com/cross-rs/cross/pull/1597#issuecomment-2567588350), and the reason seems to be the line I commented on. I can take over it if you’d like.

Emilgardis:
> @Emilgardis @ianks This PR looks great. I encountered this issue on my arm64 PC and was looking for a solution like this. Do you have any updates on this? It seems the last blocker is the build failure [here](https://github.com/cross-rs/cross/pull/1597#issuecomment-2567588350), and the reason seems to be the line I commented on. I can take over it if you’d like.

Emilgardis:
Im closing this pr due to inactivity, if someone wants to take over feel free to do so (and please consider attributing @ianks if you derive from the work in this pr)

garasubo:
I'll take over it, and now am working on #1630

PR #1609, opened 2025-01-17
Kill asmjs-unknown-emscripten

purplesyringa:
asmjs has been in limbo for years. Modern rustc does not even acknowledge it as an existing target anymore. As such, reap it out.

This commit also removes emscripten.sh, which has not been used anywhere even prior to this change.

My apologies for not running tests on this commit before pushing. They are all failing at this moment, and I don't think it's related to my changes.


Emilgardis (approved):
Thank you! I'll get this merged once CI is fixed

PR #1611, opened 2025-01-17
Update owo-colors

purplesyringa:
The goal here is to remove a transitive dependency on the deprecated `atty` crate, which trips `cargo deny` because it's unmaintained.

This currently does not eliminate the dependency completely, as we're still waiting on `color-eyre` to release [1], and merging this PR as-is would result in `cargo deny` being even more angry at using different versions of the same crate, so this PR is a draft for now.

The update is non-trivial because `owo-colors` no longer supports checking whether stdin supports colors, which is, well, reasonable.

Semantically, this cements the `Stream` trait as something that can be printed onto, rather than a generic-purpose stream. Uses of `as_tty` are replaced with a direct call to the `std` `is_terminal` function, and `Stream` (now called `OwoStream`) now exclusively handles mapping `std` I/O streams to `owo-colors` streams.

[1]: https://github.com/eyre-rs/eyre/issues/215

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,17 +90,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.2.0"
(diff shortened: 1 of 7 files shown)


neuschaefer:
`color-eyre` 0.6.4 (and 0.6.5) has by now been released, so I think it's possible to make new progress here.

Meanwhile cargo-deny is also warning about the [now-deprecated](https://rustsec.org/advisories/RUSTSEC-2025-0056) `adler` crate, which is also pulled in via color-eyre. The right version of `backtrace` to get adler2 but retain a low enough MSRV is 0.3.74

Emilgardis:
msrv can be bumped freely :)

purplesyringa:
Since #1708 was merged, I'm closing this. Thanks everyone!

PR #1630, opened 2025-02-19
Support `linux/arm64/v8` for most Ubuntu-based images (part 2)

garasubo:
This is a follow-up to #1597. Docker images for Arm64-based PCs has been added.


docker/Dockerfile.aarch64-unknown-linux-gnu
```diff
@@ -12,10 +12,11 @@ RUN /xargo.sh
...
 
-RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
-    g++-aarch64-linux-gnu \
-    gfortran-aarch64-linux-gnu \
-    libc6-dev-arm64-cross
+ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-gnu-
+ENV CROSS_SYSROOT=/usr/aarch64-linux-gnu
+
+COPY apt-cross-essential.sh /
+RUN TARGET_ARCH=arm64 TARGET_TRIPLE=aarch64-linux-gnu /apt-cross-essential.sh
```
Emilgardis, line 19: Hmm, looking at this again makes me suspicious of this approach, it makes it harder to debug with a glance whats extra in the image. Id prefer it if the script was removed or if the wanted packages were explicitly passed as arguments

Emilgardis (changes requested):

docker/Dockerfile.aarch64-unknown-linux-gnu
garasubo: reverted these changes

Emilgardis:
looks good, can you rebase the changes such that there's no merge commit?

docker/Dockerfile.aarch64-unknown-linux-gnu
Emilgardis: I think there will be some problems with this, due to what was [mentioned here](https://github.com/cross-rs/cross/pull/1597#discussion_r1889656345)

Emilgardis:
need to test with --platform arm64/v8 also

ianks:
Hey thanks for doing this! Things have been crazy where I live so I've been short on time ❤️

garasubo:
@Emilgardis 

> need to test with --platform arm64/v8 also

Do you want me to change `provided_images.rs` and related codes to build images for arm64/v8 in this PR? I plan to change them in another PR.

garasubo:
rebased the branch to remove merge commit from this PR.

Emilgardis:
> @Emilgardis
> 
> > need to test with --platform arm64/v8 also
> 
> Do you want me to change `provided_images.rs` and related codes to build images for arm64/v8 in this PR? I plan to change them in another PR.

If you think you can do it in this pr, feel free to do so. Take care of the comments i left in the previous pr about some ways to make it work.

garasubo:
@Emilgardis I caught up the discussion on the previous PR, and tried to change to enable to publish arm64 images, but it was harder than I expected. So, I'd like to merge this PR first, and then submit another PR. Could you review my change again?
I think I fixed the errors in the previous run and rebased the branch to remove merge commit from this PR.

docker/apt-cross-essential.sh
Emilgardis, line 1: This file is not used anymore, i think however it would be worth to incorporate the command checking it does since the previous pr mentioned it was needed to avoid issues with gfortran

Emilgardis (changes requested):
Going to try building some images myself, but overall it looks fine

docker/apt-cross-essential.sh
garasubo: I'm not sure what kind of issues with gfortran the previous pr were referring to. But it looks all Dockerfiles have `apt install gfortran-${TARGET_TRIPLE}`. Is there anything else that should be done? Otherwise, I'll just remove this file from the pr.

Emilgardis:
going to merge this, still some work left that is needed, but we'll work on that

PR #1664, opened 2025-04-12
feat: add riscv64gc-unknown-linux-musl image

### LINKED ISSUE #1197: Request: support for `riscv64-unknown-linux-musl`
### Checklist

- [X] This feature could be solved with a [custom docker image](https://github.com/cross-rs/cross#custom-docker-images) (optional)

### Describe your request

Simply support for `riscv64-unknown-linux-musl` when cross compiling
I tried to make my own docker image by modifying the aarch64 musl image, but i could not get it right

some systems use systems like alpine use musl and alpine is used as the basis for many lightweight/embeded llinux and riscv is a small architecture so it (appears to be) easy to emulate
support for riscv64 musl would be very nice for cross compiling (it is also a rust target)

a-wing:
Close #1197

* * *

Only use crosstools-ng

If use [musl-cross-make](https://github.com/richfelker/musl-cross-make), `binutils` version is to low too `ld` cannot link in riscv

I test [musl.cc](http://musl.cc/) [riscv64-linux-musl-cross.tgz](http://musl.cc/riscv64-linux-musl-cross.tgz), It's ok `binutils: 2.37`, `ld` can link, It's working

But, `musl-cross-make` is only support `binutils: 2.33.1`, `binutils: 2.33.2` is big changed, The patch is unavailable, migration this patch is very complex (a lot of conflict). BTW: These patches have nothing to do with riscv

### DIFF
--- a/.changes/1664.json
+++ b/.changes/1664.json
@@ -1,0 +1,5 @@
+{
+    "type": "added",
+    "description": "add riscv64gc-unknown-linux-musl support",
+    "issues": [1197]
+}
--- a/.github/ISSUE_TEMPLATE/b_issue_report.yml
+++ b/.github/ISSUE_TEMPLATE/b_issue_report.yml
@@ -66,6 +66,7 @@ body:
         - powerpc64-unknown-linux-gnu
         - powerpc-unknown-linux-gnu
         - riscv64gc-unknown-linux-gnu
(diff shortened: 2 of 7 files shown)


Emilgardis (approved):
Thanks! Looks good to me :)

PR #1665, opened 2025-04-12
fix Dockerfile warn FromAsCasing as => AS

a-wing:
> => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)                                                                                                                                     
> => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 13)


Emilgardis (approved):
Thanks!

PR #1666, opened 2025-04-13
fix Dockerfile warn LegacyKeyValueFormat ENV key value => ENV key=value

a-wing:
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 20) https://docs.docker.com/reference/build-checks/legacy-key-value-format/


Emilgardis (approved):
lgtm

PR #1670, opened 2025-04-18
[RFC] Shrink musl docker image size by ~50%

crazyscot:
Here is a proof of concept patchset which reduces the size of the musl images from ~2.4G to ~1.2G, on the two architectures I tried (x86_64 and aarch64).

Using aarch64 as a primary testbed, I got these results:
* Baseline (edge): image size 2.39G (794M gzipped)
* Patch 1 (strip the binaries in musl.sh): 1.55G
* Add patch 2 (flatten the output image): 1.31G
* Add patch 3 (additional tidyup script): 1.22G (348M gzipped).

A space/bandwidth saving of 47% uncompressed, 56% compressed.

I've tested the result with my use cases (x86_64 and aarch64). The crate I am cross-compiling depends on `ring` which has a C component.

There is almost certainly more squeezing that could be done, but the Law of Diminishing Returns kicks in somewhere. For example, I spotted copies of identical binaries in /usr/local/bin which could be symlinked together to save a few more MB, but coming up with a safe and robust solution there might take a little thought.

The majority of these savings are specific to the musl builds. I haven't analysed any of the non-musl images but there could be scope for improvement there as well.

If you like this approach, I'll happily copy & paste the tidyup & flatten part into the other dockerfiles. OTOH, there may be other ways to achieve this; happy to discuss.

### DIFF
--- a/docker/Dockerfile.aarch64-unknown-linux-musl
+++ b/docker/Dockerfile.aarch64-unknown-linux-musl
@@ -18,6 +18,13 @@ RUN /qemu.sh aarch64
 COPY musl.sh /
 RUN /musl.sh TARGET=aarch64-linux-musl
 
+COPY tidyup.sh /
+RUN /tidyup.sh
+
+FROM scratch AS final
+COPY --from=build / /
+CMD ["/bin/bash"]
+
 ENV CROSS_TOOLCHAIN_PREFIX=aarch64-linux-musl-
 ENV CROSS_SYSROOT=/usr/local/aarch64-linux-musl
 COPY musl-symlink.sh /
--- a/docker/Dockerfile.arm-unknown-linux-musleabi
+++ b/docker/Dockerfile.arm-unknown-linux-musleabi
@@ -22,6 +22,13 @@ RUN /musl.sh \
                       --with-float=soft \
                       --with-mode=arm"
 
+COPY tidyup.sh /
+RUN /tidyup.sh
+
+FROM scratch AS final
+COPY --from=build / /
+CMD ["/bin/bash"]
+
 ENV CROSS_TOOLCHAIN_PREFIX=arm-linux-musleabi-
 ENV CROSS_SYSROOT=/usr/local/arm-linux-musleabi
 COPY musl-symlink.sh /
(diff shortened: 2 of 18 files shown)


Emilgardis:
Thank you for the pr!

This is tempting, however wouldn't it make some types of troubleshooting harder?

crazyscot:
I've only stripped the toolchain binaries, not the target libc. If I had stripped the target libc and CRT, that I think would hamper certain deep debug scenarios. (But to be honest, I can't remember the last time I needed a debug libc.)

The only troubleshooting I can see this making harder is if you somehow needed to debug the cross-compiler itself. That is extremely rare, and I note that the toolchains are already stripped on other target images. For example, on aarch64-unknown-linux-gnu:edge:
```
root@1da000ab810e:/# file /usr/bin/aarch64-linux-gnu-gcc-9
/usr/bin/aarch64-linux-gnu-gcc-9: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=a18cab6773f8430914a40048e4fa10bb787ef473, for GNU/Linux 3.2.0, stripped
```

Emilgardis:
Alright that was mainly what my concern was. Let's skip stripping qemu and wine, otherwise I think I'd be happy to have this included for the other images too

crazyscot:
OK, I've updated the strip logic to apply only to the toolchain, and converted the hard links to symlinks.
I noticed that there were a couple of same-directory hard links, which I also converted to symlinks.

Flatten and tidyup is now applied to all the musl images.
For comparison, the aarch64 musl image is now 1.26GB.

By the way, it looks like the freebsd images would benefit from the same treatment. However, this isn't my use case; I don't have a ready way to test the output. I suspect that applies to all the builds that build their own gccs (netbed, solaris, illumos, dragonfly).

Emilgardis:
/ci try --target aarch64-unknown-linux-musl --target x86_64-unknown-linux-gnu

crazyscot:
I've cleaned up the Dockerfile warnings, but there isn't a lot I can do about the `502 Bad Gateway` from musl.libc.org...

Emilgardis (approved):
Thank you! Lets merge this :)

PR #1636, opened 2025-03-04
update ci tools to publish arm64 docker images

garasubo:
Update CI and related tools to publish multi-arch images
Ref: #1597, #1630

PR #1672, opened 2025-04-24
xtask/build_docker_image: Add `windows/amd64` platform handling

MarijnS95:
As listed at for example https://docs.docker.com/build/building/multi-platform/.

Allows using `--platform windows/amd64` rather than `--platform x86_64-pc-windows-msvc`.

---

I didn't manage to use this in GitHub Actions CI on `windows-latest` just yet because it complains that `--platform` is not understood, yet the help text suggests that `buildx` isn't installed in these images in the first place?: https://github.com/Traverse-Research/cpuinfo-rs/actions/runs/14644208913/job/41094014439#step:5:334


xtask/src/build_docker_image.rs
```diff
@@ -363,6 +363,7 @@ fn has_native_image(
     Ok(match target.sub.as_deref() {
         // FIXME: add additional subs for new Linux distros, such as alpine.
         None | Some("centos") => match (platform, target.name.as_str()) {
+            // TODO: "windows/amd64"?
```
Emilgardis, line 366: not needed, we will never have a msvc image available like this through a "native" extra dockerfile

Emilgardis (approved):
looks good, minor nit about the todo comment

Emilgardis:
regarding buildx/buildkit, there's experimental support for it, see this list for their proress: https://github.com/moby/buildkit/issues?q=is%3Aissue%20state%3Aopen%20label%3Aarea%2Fwindows-wcow

and this issue for the popular setup-buildx-action https://redirect.github.com/docker/setup-buildx-action/issues/292

MarijnS95:
@Emilgardis thanks, so it's `buildx` that doesn't support Windows just yet.

What's the right way to use this `cross` project for an `aarch64-pc-windows-msvc` target? As far as I'm aware Docker is "not intended" to emulate Windows on Linux despite having QEMU available to emulate the `aarch64` architecture, and at the same time `buildx` cannot be used on Windows yet (and if it is, what's the setup for emulating `aarch64` on an `x64` GitHub Actions runner)?

As far as I understand `cross` also uses the container (and/or emulation?) to cross-compile? Otherwise we still have a setup to cross-compile (for various Windows MSVC targets/architectures) with the native compiler using https://github.com/Jake-Shadle/xwin/, but that (obviously) isn't able to ~run~ emulate tests etc.

Emilgardis:
We have a msvc image here already, we just don't publish its not certain it's actually any good: https://github.com/cross-rs/cross-toolchains/blob/main/docker/Dockerfile.aarch64-pc-windows-msvc-cross, it does not use the container platform `windows`. it uses `linux`, so yes, it's cross compilation.

cross in general does not compile through emulation, we cross-compile. Although, we do emulation on aarch64 hosts for now because we don't have the infrastructure rn to make linux/arm64 images published (see #1636)

MarijnS95:
@Emilgardis thanks for confirming, I see now (should have opened the docker image before): it's emulating native MSVC in an Ubuntu Docker image on `wine`, hence should run in a Docker container on a Linux/Ubuntu GitHub runner :+1: 

Trying this now on my dev machine, it is bindmounting `~/.cargo` and conflicting with my aforementioned `xwin`+`clang`-based setup from `[target.aarch64-pc-windows-msvc]` + `[env]` in `~/.cargo/config.toml` (need to comment it out, it's fortunately not present on the CI though).

It's also surprising that `cross build --target=aarch64-pc-windows-msvc` first tries to install the `stable-x86_64-unknown-linux-gnu` (native host) toolchain before spawning a `docker` container and continuing with the Windows target there.

Finally, the `Dockerfile.aarch64-pc-windows-msvc` you linked looks to be missing `BINDGEN_EXTRA_CLANG_ARGS` so `bindgen` is failing with obvious errors like "`stdint.h` not found"...

PR #1674, opened 2025-04-27
freebsd: fix image build

aokblast:
Currently, this patch is tested by built the docker image but haven't test the build of any application upon the docker image.

When I use cross to build application, it will create the following Dockerfile in the target directory:

```
FROM ghcr.io/cross-rs/x86_64-unknown-freebsd:main
                ARG CROSS_DEB_ARCH=
                ARG CROSS_CMD
                RUN eval "${CROSS_CMD}"
```

I try to replace the "FROM" sentence to my local build image but when `cross build` is executed, it will recover to ghcr.io again.

FreeBSD change the compress algorithm of the packagesite to zstd. Also it change the download filename.

Besides fixing the download issue, I update the gcc version to 13

### DIFF
--- a/.changes/1674.json
+++ b/.changes/1674.json
@@ -1,0 +1,4 @@
+{
+    "type": "fixed",
+    "description": "use zstd to extract the compressed file as FreeBSD do so. also, bump the gcc version"
+}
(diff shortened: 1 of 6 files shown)


Emilgardis:
`/ci try --target *freebsd`

aokblast:
Emm, I think I cannot call it because I am not a member. Could you please help me run the CI again?

Emilgardis (changes requested):
the gcc bump requires an update in the readme file 👍🏼 You can use `cargo xtask target-info <targets>` to automatically extract the different versions if you want

otherwise this looks good!

Emilgardis:
Just missing a changelog entry now for the gcc bump 👍🏼

Emilgardis (commented):
LGTM!

PR #1683, opened 2025-05-23
fix(rustup): installed toolchains for legacy rustup

### LINKED ISSUE #1678: Incorrect toolchain check when running cross with active or default toolchain
### Checklist

- [x] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

When running any cross command, cross tries to download the stable version while it is already installed.

This is because when running `rustup toolchain list`, my toolchain stable has a suffix `(active, default)`, which makes the check fail. I believe adding the `--quiet` flag **at the end** of the command should fix the issue. 

I have tried both the latest stable 0.2.5 version and the lastet version on the main branch (9e2298e17170655342d3248a9c8ac37ef92ba38f at the time of writing)

Two months ago, on a version not released yet, the command `rustup toolchain list` was replaced with `rustup --quiet toolchain list`. This however does not fix the issue as the `--quiet` flag applies to the `rustup` command and not the `list` command.

_No response_

- [ ] macOS
- [ ] Windows
- [x] Linux / BSD

- [x] x86_64 / AMD64
- [ ] arm32

- [ ] docker
- [ ] podman

### cross version

cross 0.2.5

### Example

```shell-session
$ cross --help -v
+ cargo metadata --format-version 1
+ rustc --print sysroot
+ rustup toolchain list
+ rustup --verbose toolchain add stable-x86_64-unknown-linux-gnu --profile minimal
debug: read metadata version: '12'
debug: updating existing install for 'stable-x86_64-unknown-linux-gnu'
debug: toolchain directory: '/home/user/.rustup/toolchains/stable-x86_64-unknown-linux-gnu'
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
debug: creating temp file: /home/user/.rustup/tmp/yivvct2ihzgg02o9_file
debug: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
debug: downloading with reqwest
debug: deleted temp file: /home/user/.rustup/tmp/yivvct2ihzgg02o9_file
debug: toolchain is already up to date

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.86.0 (05f9846f8 2025-03-31)

info: checking for self-update
+ rustup target list --toolchain stable-x86_64-unknown-linux-gnu
+ rustup component list --toolchain stable-x86_64-unknown-linux-gnu
+ cargo --help -v
Rust's package manager

Usage: cargo [+toolchain] [OPTIONS] [COMMAND]
       cargo [+toolchain] [OPTIONS] -Zscript <MANIFEST_RS> [ARGS]...
...
```

Example of `rustup toolchain list` result

```shell-session
$ rustup toolchain list
stable-x86_64-unknown-linux-gnu (active, default)
nightly-x86_64-unknown-linux-gnu
```

_No response_

reneleonhardt:
Fixes #1678

Unfortunately some installations use legacy versions, my previous fix was only working for the current rustup >= 1.28.0.


src/rustup.rs
```diff
@@ -84,11 +86,17 @@ pub fn active_toolchain(msg_info: &mut MessageInfo) -> Result<String> {
...
     let out = rustup_command(msg_info, true)
-        .args(["toolchain", "list", "--quiet"]) // suppress " (active, default)" suffix
+        .args(["toolchain", "list"]) // --quiet would be available from 1.28.0
         .run_and_get_stdout(msg_info)?;
 
-    Ok(out.lines().map(|l| l.trim().to_owned()).collect())
+    // Emulate --quiet by removing suffixes like " (active, default)" or " (override)" manually
+    Ok(out
+        .lines()
+        .map(|l| SUFFIX_REGEX.replace(l.trim(), "").into_owned())
```
Emilgardis, line 98: Regex is not needed for this.

Emilgardis (changes requested):

src/rustup.rs
reneleonhardt: How? 🤔

Emilgardis: Please squash the commits and we can merge. If possible I'd like a test for this also, but no stress we can fo that later

reneleonhardt: I couldn't find any tests for rustup.rs and I know that you don't like new crates like visibility to allow doctests... 
https://github.com/cross-rs/cross/pull/1684

PR #1697, opened 2025-07-18
fix env var for custom toolchain compatibility

Jonathas-Conceicao:
I was having some trouble setting toolchain compatibility with nix flakes environment and noticed the env var the configuration reads is actually different from the one the [documention](https://github.com/cross-rs/cross/blob/51f46f296253d8122c927c5bb933e3c4f27cc317/docs/environment_variables.md?plain=1#L48) and other parts of [code](https://github.com/cross-rs/cross/blob/51f46f296253d8122c927c5bb933e3c4f27cc317/src/rustc.rs#L187) suggests. This fixes the environment variable loaded on configuration

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -189,7 +189,7 @@ impl Environment {
     }
 
     fn custom_toolchain_compat(&self) -> Option<String> {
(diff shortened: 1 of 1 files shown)


Emilgardis:
Oops, thats definitely wrong!

PR #1699, opened 2025-08-15
Add ability to set rust version explicitly, rather than `rustup` latest

bleggett:
The image build scripts always let the `rustup` binary choose the latest `rust` target version to install. 

This can break things when `rustup` is updated and pulls a newer version of rust that the components in this repo haven't yet been updated to build against, see https://github.com/japaric/xargo/pull/349 for example.

That can be fixed, but in general it would be good to allow a specific/fixed/pinned version to be specified, if desired, for whatever reason.

This PR adds the ability to request a specific rust target version by setting the `RUST_VERSION` container arg, such as

- `RUST_VERSION=1.88`

if unset, or `RUST_VERSION=latest`, fall back to current behavior of letting `rustup` decide what version to install (current behavior).

I updated the `ARG` default for `RUST_VERSION` to be `latest` in all the images, for the sake of explicitness.

### DIFF
--- a/docker/Dockerfile.aarch64-linux-android
+++ b/docker/Dockerfile.aarch64-linux-android
@@ -1,5 +1,6 @@
 FROM ubuntu:20.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
+ARG RUST_VERSION=latest
 
 COPY common.sh lib.sh /
 RUN /common.sh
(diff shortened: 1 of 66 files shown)


Emilgardis:
Thank you for the pr! However, I think I'd rather just remove xargo, it's not needed anymore with build-std working well. What do you think? see #692

bleggett:
> Thank you for the pr! However, I think I'd rather just remove xargo, it's not needed anymore with build-std working well. What do you think? see #692

Ah, missed that! That sounds better.

PR #1707, opened 2025-09-25
Add aarch64_be-unknown-linux-gnu target

neuschaefer:
Compared to [aarch64_be-unknown-linux-musl](https://github.com/cross-rs/cross/pull/1706), the GNU target has been supported in Rust for much longer.

TODO:
- [ ] .


Emilgardis:
Did you forget to add the dockerfile? :D

neuschaefer:
> Did you forget to add the dockerfile? :D

I forgot to write it, because I didn't know how this part of cross works :)

Emilgardis:
the dockerfile is essential for providing the cross-compilation tools needed for C (and other) deps. Just pure rust will work fine withut cross

docker/common.sh
```diff
@@ -42,6 +43,7 @@ if_centos install_packages \
     pkgconfig
 
 if_ubuntu install_packages \
+    adduser \
```
Emilgardis, line 46: why was this added?

Emilgardis (approved):
lgtm, one question :)

docker/common.sh
neuschaefer: Both bzip2 and adduser were necessary during the build. I'm rerunning it now to give a more detailed answer

neuschaefer: Both of them are due to crosstool-ng.sh, so I'm moving the installations there, and documenting the reasons in the commit message

PR #1722, opened 2025-11-08
Downstreamed Fix: Bad docker image name prefix pruning

KronsyC:
If a workspace name started with _ or ., the generated docker image name segment would be invalid due to `-_` or `-.` sequences, this code simply extends the postfix pruning code to also prune these prefixes


whycanty:
fuck my bad, this PR shouldn't exist. Im so sorry, messed up the git-foo

Issue #1618, opened 2025-01-27
Linker wrapper script for musl outputs garbage due to `set - x`


VorpalBlade:
### Checklist

- [x] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

https://github.com/cross-rs/cross/blob/main/docker/musl-gcc.sh uses `set -x`, which causes unexpected output. Rustc now issues a lint for that (on nightly). This breaks `-Dwarning` builds. See this rustc issue: https://github.com/rust-lang/rust/issues/136132

Cross should remove that `set -x` and then make a new release of the affected component(s). The docker images in this case I suppose? 

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [x] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [x] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [x] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.5 (66221abde 2024-11-19)

### Example

Cross compile any program to aarch64-unknown-linux-musl.

VorpalBlade:
> I think it's unfortunate that this is an error-by-default, but oh well, it's a simple thing to fix. PRs welcome, otherwise I'll get to it myself

I believe they are currently planning to roll back this *temporarily* due to the fallout of the it.

Emilgardis:
Ah yes, it was changed: https://github.com/rust-lang/rust/pull/136098

Noratrieb:
Yes, we rolled it back for now to fix the warnings on our side and in places like here before enabling it again to reduce fallout.

I see that this script was added 3 years ago to work around issues in "older Rust versions". Maybe it can just be removed now?

bjorn3:
> I think it's unfortunate that this is an error-by-default, but oh well, it's a simple thing to fix. PRs welcome, otherwise I'll get to it myself

It is a warning by default. OP used `-Dwarnings` upgrading any warning to an error: https://github.com/VorpalBlade/chezmoi_modify_manager/blob/12c5fea1442fcd7e465f827c9963475d088cf73e/.github/workflows/build.yml#L22

VorpalBlade:
[...]

PR #1728, opened 2025-12-12
Remove set -x from gcc wrappers and fix behavior with 0 arguments

jyn514:
Fixes https://github.com/cross-rs/cross/issues/1618.

Additionally fixes the behavior when invoked with 0 arguments:
`${@}` is a single word, regardless of how many arguments were passed to
the program. To expand it as an array, it must be `$@`.
See https://www.gnu.org/software/bash/manual/html_node/Shell-Expansions.html.

```
docker/musl-gcc.sh: line 73: @: unbound variable
docker/freebsd-gcc.sh: line 16: @: unbound variable
```

### DIFF
--- a/docker/freebsd-gcc.sh
+++ b/docker/freebsd-gcc.sh
@@ -7,10 +7,10 @@ set -euo pipefail
 
 main() {
     if [[ $# -eq 0 ]]; then
-        exec "${CROSS_TOOLCHAIN_PREFIX}gcc" "${@}"
+        exec "${CROSS_TOOLCHAIN_PREFIX}gcc"
     else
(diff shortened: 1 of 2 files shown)


jyn514:
I do want to note that the bug that `musl-gcc` exists to workaround was fixed in 1.65, which was released in November 2022, more than three years ago.

Emilgardis (approved):
let's merge this even though it's not required for newer versions :)

thank you!

jyn514:
thanks for the review! i'm hoping this in combination with a few other changes will let me turn https://github.com/rust-lang/rust/issues/136096 back on in a release or two :)

PR #1734, opened 2026-01-01
fix: upgrade dependencies for FreeBSD 15+ compatibility

seuros:
Happy new year

I upgraded all the crates while doing the refactoring. 

Upgrades several dependencies with breaking API changes:
- nix 0.26.2 → 0.30.1 (FreeBSD 15+ support)
- thiserror 1.0 → 2.0
- toml 0.7 → 0.9
- which 4.4 → 8.0
- const-sha1 0.2 → 0.3
- signal-hook 0.3 → 0.4
- directories 4.0 → 6.0

API migrations:
- nix: errno() → Errno::last_raw(), from_i32() → from_raw()
- const-sha1: ConstBuffer removed, use sha1() directly
- toml: Deserializer::new() → Deserializer::parse()
- which: enable default features for real-sys support

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
       - uses: ./.github/actions/setup-rust
         with:
           components: clippy
-          toolchain: 1.77.2 # MSRV, Minimally Supported Rust Version. Make sure to update README.md and clippy.toml
+          toolchain: 1.92.0 # MSRV, Minimally Supported Rust Version. Make sure to update README.md and clippy.toml
       - name: Run clippy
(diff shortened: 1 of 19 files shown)


seuros:
one of the failure is not related to my update.

The other is ! Should we upgrade to edition 2024 ?  I can upgrade the codebase.

Emilgardis:
You just need to update the version of cargo/rustc that we use in CI, do not have to upgrade the whole project to 2024. (cargo is perfectly capable of mixing editions between crates)

here's where you'd do that: https://github.com/cross-rs/cross/blob/e7b8115d0bd161b9791eea171461e3df27d0a78b/.github/workflows/ci.yml#L77

Emilgardis (approved):
lgtm! thank you!

PR #1748, opened 2026-02-06
improve /ci try

Emilgardis:
- add reason for failure in output
- make it possible to pass CROSS_DEBUG=true


Emilgardis:
wanted for #1580

PR #1752, opened 2026-02-25
Fix again FreeBSD image build

ydirson:
The previous fix used a suffix that seems to have been dropped since.

### DIFF
--- a/docker/freebsd-install.sh
+++ b/docker/freebsd-install.sh
@@ -121,8 +121,8 @@ setup_freebsd_packagesite() {
     pkg_source=$(freebsd_package_source "${url}")
 
     mkdir -p "${FREEBSD_PACKAGEDIR}"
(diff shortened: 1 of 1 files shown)


Emilgardis:
Thanks for the pr! I found this too, but I solved it this way instead as I was unable to get pkg to work correctly.

https://github.com/cross-rs/cross/pull/1580/changes/53708d20f50a95ee464a245cc8d4dc9f64f49149

/ci try *freebsd*

ydirson:
> Thanks for the pr! I found this too, but I solved it this way instead as I was unable to get pkg to work correctly.

Note my tests were quite minimal actually, I only verified that the setup script does work, and I'm able to install my build dependencies.

Emilgardis:
oops forgot to add the label for no ci :D

Issue #1737, opened 2026-01-03
i686-pc-windows-gnu broken for multiple applications


mcandre:
### Checklist

- [x] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

Hi,

I'm able to port my Docker utility application, tuggy, to dozens of platforms using cross. However, the target `i686-pc-windows-gnu` somehow breaks down, producing strange errors regarding the standard library.

https://github.com/mcandre/tuggy

### What target(s) are you cross-compiling for?

i686-pc-windows-gnu

### Which operating system is the host (e.g computer cross is on) running?

- [x] macOS
- [ ] Windows
- [ ] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [ ] x86_64 / AMD64
- [ ] arm32
- [x] arm64 (including Mac M1)

### What container engine is cross using?

- [x] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross 0.2.5 (4e64366 2025-04-04)

### Example

Automating cross builds with crit, basically a for loop over most of the available Rust target triples.

1. Clone tuggy.
2. Install crit with `make -f install.mk`.
3. Remove `i686-pc-windows-gnu` from the exception list in `makefile`.
4. Run `make port`.

### Additional information / notes

Sometimes a computer process fails due to memory or disk issues, especially containers. Ran `docker system prune -f`, no change in behavior. macOS host RAM and host disk stable, with plenty to spare.

mcandre:
Update:

tinyrick is also effected. https://github.com/mcandre/tinyrick

The breakage appears to coincide with upgrading Rust from 1.91 to 1.92.

Emilgardis:
Please provide some logs or error messages, hard to get the context without it.

m-z-b:
Might be seeing the same problem using cargo 1.93.0 and the sample hello program. Here's the error messages:

Cargo: 1.93.0 (083ac5135 2025-12-15)
Rustc: 1.93.0 (254b59607 2026-01-19)
Host:  WSL2 on Windows 10. 
Error messages:
```sh
/mnt/c/projects/Rust$ cargo init hello2
    Creating binary (application) package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
/mnt/c/projects/Rust$ cd hello2
[...]

Issue #1730, opened 2025-12-21
sparcv9-sun-solaris linking error `undefined reference to `__sysconf_xpg7'`


Wyvern:
sparcv9-sun-solaris  cross build got linking error:

```
error: linking with `sparcv9-sun-solaris2.10-gcc` failed: exit status: 1
  │
  ├ note:  "sparcv9-sun-solaris2.10-gcc" "-m64" "<1 object files omitted>" "-Wl,-z,ignore" "-Wl,-Bstatic" "/project/build-dir/sparcv9-sun-solaris/src/build/compiler_builtins/1cba6cf8fd5c5c75/deps/libcompiler_builtins-1cba6cf8fd5c5c75.rlib" "-Wl,-Bdynamic" "-lsendfile" "-llgrp" "-lsocket" "-lposix4" "-lpthread" "-lresolv" "-lgcc_s" "-lc" "-lm" "-lrt" "-lpthread" "-lsendfile" "-llgrp" "-L" "/tmp<sysroot>cJ0efRX/raw-dylibs" "-o" "/project/build-dir/sparcv9-sun-solaris/src/build/Img/26156b48f1a83f81/deps/Img" "-Wl,--strip-all" "-nodefaultlibs"
  ├ note: some arguments are omitted. use `--verbose` to show all linker arguments
  ├ note: /usr/local/lib/gcc/sparcv9-sun-solaris2.10/8.4.0/../../../../sparcv9-sun-solaris2.10/bin/ld: warning: -z ignore ignored.
  │       /project/build-dir/sparcv9-sun-solaris/src/build/Img/26156b48f1a83f81/deps/Img.Img.f3822acec0a0948b-cgu.0.rcgu.o: In function `_RNxC3Img12H8lQSjiMX4h2':
  │       Img.f3822acec0a0948b-cgu.0:(.text._RNxC3Img12H8lQSjiMX4h2+0x65c): undefined reference to `__sysconf_xpg7'
  │       /project/build-dir/sparcv9-sun-solaris/src/build/Img/26156b48f1a83f81/deps/Img.Img.f3822acec0a0948b-cgu.0.rcgu.o: In function `main':
  │       Img.f3822acec0a0948b-cgu.0:(.text.main+0x214): undefined reference to `__sysconf_xpg7'
  │       collect2: error: ld returned 1 exit status
  │       
  ├ note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
[...]

Issue #1599, opened 2024-12-30
Support Solaris 11


ahqsoftwares:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar request
- [ ] This feature could be solved with a [custom image](https://github.com/cross-rs/cross/blob/main/docs/custom_images.md#custom-images) (optional)

### Describe your request

Cross should also support solaris 11 as some rust programs cannot be built with solaris 10, example below:

https://github.com/leadlang/lead/actions/runs/12541529005/job/34970191833
error: linking with `x86_64-pc-solaris2.10-gcc` failed: exit status: 1
  |
[...]

PR #1755, opened 2026-03-11
Upgrade windows docker images base to Ubuntu 24.04

### LINKED ISSUE #1754: x86_64-pc-windows-gnu image 0.2.5: link failure — `GetHostNameW` missing from mingw-w64
## Summary

Building a Rust nightly project with `cross build --target x86_64-pc-windows-gnu` using the `0.2.5` image fails at link time with an undefined reference to `GetHostNameW`. All Rust compilation succeeds — only the final mingw linker step fails.

## Environment

- **Host**: `aarch64-apple-darwin` (Apple Silicon, Docker Desktop with Rosetta)
- **Rust toolchain**: `nightly-x86_64-unknown-linux-gnu` — `1.96.0-nightly (80282b130 2026-03-06)`
- **cross version**: `0.2.5`
- **Docker image**: `ghcr.io/cross-rs/x86_64-pc-windows-gnu:0.2.5`
- **CROSS_CONTAINER_OPTS**: `--platform linux/amd64` (required on ARM64 host)

## Error

```
error: linking with `x86_64-w64-mingw32-gcc` failed: exit status: 1
  ...
  /rust/lib/rustlib/x86_64-pc-windows-gnu/lib/libstd-949922d303bba3ab.rlib(std-*.rcgu.o):
    .../std/src/sys/net/hostname/windows.rs:15: undefined reference to `GetHostNameW'
  collect2: error: ld returned 1 exit status
```

The linker command does include `-lws2_32`, so the library is being linked — but the mingw-w64 version in the image doesn't export `GetHostNameW`.

## Root Cause

Rust nightly's `std` now uses `GetHostNameW` (in `std::sys::net::hostname::windows`). This function requires a newer mingw-w64 than what's bundled in the `0.2.5` image. The `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` images work fine — this only affects the Windows GNU cross-compilation image.

## Reproduction

```bash
# Any Rust project using tokio or std networking should reproduce this
cargo init --name repro && cd repro
echo 'tokio = { version = "1", features = ["full"] }' >> Cargo.toml
cat > src/main.rs << 'EOF'
#[tokio::main]
async fn main() {
    println!("hello");
}
EOF
CROSS_CONTAINER_OPTS="--platform linux/amd64" cross build --target x86_64-pc-windows-gnu
```

## Expected Behavior

The build should link successfully, as it does for the Linux targets.

## Suggested Fix

Update the mingw-w64 toolchain in the `x86_64-pc-windows-gnu` Docker image to a version that includes `GetHostNameW` in its `ws2_32` import library.

Raniz85:
Newer versions of Rust require a newer version of MingW.

Fixes #1754, #1737

### DIFF
--- a/docker/Dockerfile.i686-pc-windows-gnu
+++ b/docker/Dockerfile.i686-pc-windows-gnu
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04 AS cross-base
+FROM ubuntu:24.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
 
 COPY common.sh lib.sh /
--- a/docker/Dockerfile.x86_64-pc-windows-gnu
+++ b/docker/Dockerfile.x86_64-pc-windows-gnu
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04 AS cross-base
+FROM ubuntu:24.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
 
 COPY common.sh lib.sh /
@@ -16,6 +16,12 @@ COPY wine.sh /
 RUN /wine.sh
 
 RUN apt-get update && apt-get install --assume-yes --no-install-recommends g++-mingw-w64-x86-64 gfortran-mingw-w64-x86-64
+
+# Ensure we compile with the POSIX versions of mingw
+RUN update-alternatives --set x86_64-w64-mingw32-gcc  /usr/bin/x86_64-w64-mingw32-gcc-posix  && \
+    update-alternatives --set x86_64-w64-mingw32-g++  /usr/bin/x86_64-w64-mingw32-g++-posix  && \
+    update-alternatives --set x86_64-w64-mingw32-gfortran /usr/bin/x86_64-w64-mingw32-gfortran-posix 2>/dev/null || true && \
+    update-alternatives --set x86_64-w64-mingw32-gcj  /usr/bin/x86_64-w64-mingw32-gcj-posix  2>/dev/null || true
 
 # run-detectors are responsible for calling the correct interpreter for exe
 # files. For some reason it does not work inside a docker container (it works
@@ -49,4 +55,6 @@ ENV CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER="$CROSS_TOOLCHAIN_PREFIX"gcc"$CROS
     CROSS_CMAKE_SYSTEM_NAME=Windows \
     CROSS_CMAKE_SYSTEM_PROCESSOR=AMD64 \
     CROSS_CMAKE_CRT=gnu \
-    CROSS_CMAKE_OBJECT_FLAGS="-ffunction-sections -fdata-sections -m64"
+    CROSS_CMAKE_OBJECT_FLAGS="-ffunction-sections -fdata-sections -m64" \
+    PKG_CONFIG_ALLOW_CROSS=1 \
+    PKG_CONFIG_PATH="/usr/x86_64-w64-mingw32/lib/pkgconfig/"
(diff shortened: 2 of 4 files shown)


Raniz85:
Error is about XDG_RUNTIME_DIR, I'll have a look tomorrow

--

Raniz

Emilgardis:
have a look at https://github.com/cross-rs/cross/pull/1580 also, where we've been trying to bump the image

Raniz85:
This turned out to be quite the rabbit hole. Made deeper because I haven't worked with C++ for some 20 years :S

I've spent considerable time trying to figure out what goes wrong - with a lot of help from pushing Claude in various directions.

The issue seems to be that the rust-cmake-hello-world crate is incompatible with newer GCC.
The reason that Claude gave me is that re2 uses `call_once` which is reliant on thread locals and this doesn't work with MingW-GCC-13 across DLL-boundaries. I found some info online that seems to support this (among others, this: https://github.com/llvm/llvm-project/issues/161039).

The solution is to either force libstdc++ to be linked statically (I don't like this solution) or to upgrade RE2 (newer versions doesn't seem to use `call_once`, which turned out to be a bit of a rabbit hole itself :)

Either way, the fix needs to be done in https://github.com/cross-rs/rust-cmake-hello-world. 

I think that upgrading RE2 is the better option, but that will break the test for most (all?) other images right now since they don't support C++17, which Abseil (a dependency of RE2 since ~May 2024) requires - it could help the effort in #1850 though since 24.04 has toolchains that support C++17.

There's a PR open on _rust-cmake-hello-world_ to upgrade RE2, once that is merged, the CI run should pass for these images again (though it will break for pretty much all the other images that use GCC-8 or older).

One option would be to point test.sh to specific revisions of _rust-cmake-hello-world_ instead of _main_ until all images have been updated.

Raniz85:
Two questions:

1. Should I change test.sh to use different refs for rust-cmake-hello-world depending on which image we're testing?
2. Should we combine this PR with #1580?

Emilgardis:
I think using different refs makes sense, I switched the PR to be put on branch `24` so we can make the test script use that branch instead when needed

Emilgardis:
We can combine the prs, but we can also just rebase the other one on top of it :) small incremental changes are nice though also

Raniz85:
I agree :)

I'll update the test script to use the new ref and then we can retry the image tests

MichaIng:
Since Debian/Ubuntu dropped sjlj for dwarf already (which you see in your updated patch as well), this isn't a reason anymore to compile the mingw compilers from source. Using the distro packages also makes assuring the POSIX variant obsolete, since there are explicit packages for this now: `g++-mingw-w64-x86-64-posix` `gfortran-mingw-w64-x86-64-posix`

So that way, the `mingw.sh` could be skipped and removed.

I am working on this on my fork as a continuation of #1580: https://github.com/MichaIng/cross/pull/2

But for cargo tests with `--features=re2`, I get a null pointer dereference:
````
wine: Unhandled page fault on execute access to 0000000000000000 at address 0000000000000000
````
Hence I removed `cpp = true` from the `targets.toml` for the two `windows-gnu` targets.

Or am I missing something, and own mingw compiler builds are still needed, and probably solve this particular issue I see with `re2`?

__________

_What I am still fiddling with on my fork is Solaris: #1599 #1730
I recognized a little late that I am working on issues unrelated to Ubuntu 20.04 => 24.04, but as I am on it ... DilOS 2 libs are too old for recent Rust, <s>it seems DilOS 3 is inevitable</s> (not released, just saw the APT suite, but it is empty). But I am still trying to do some linker mapping from new/missing to old/existing symbols. But this is another topic. I'll try to get everything green on my fork, and open a new PR then._

PR #1770, opened 2026-04-17
fix(docker): bump Debian Bullseye kernel version to 5.10.0-39

taladar:
linux-image-5.10.0-34-amd64 (and arm64/686) are no longer available in the Debian Bullseye repositories. The current latest is 5.10.0-39.


Emilgardis:
closing in favor of #1580

PR #1778, opened 2026-06-03
fix: suggest running cargo directly for native targets

### LINKED ISSUE #1447: Improve error message on CI when target = MSVC or Darwin and target = host
With https://github.com/cross-rs/cross/pull/661, we now error on warnings in CI.

```
[cross] warning: `cross` does not provide a Docker image for target x86_64-apple-darwin, specify a custom image in `Cross.toml`.
[cross] error: Errors encountered before cross compilation, aborting.
[cross] note: Disable this with `CROSS_NO_WARNINGS=0`
```

To clear up confusion, we could suggest a fix if it's possible that just running `cargo` would work, we can also point to https://github.com/cross-rs/cross-toolchains.

See [discussion](https://matrix.to/#/!ZFqvHUjUqgONNObpSJ:matrix.org/$QnkcnKVAYkbxBNHMogkbnHudprD2JlVSs2gFnASNdXs?via=matrix.org&via=envs.net&via=matrix.asonix.dog)

LeonidasZhak:
## Summary

Adds helpful guidance when `cross` reports `NoCompatibleImages` for native targets (Windows MSVC, macOS Darwin).

## Problem

When users run `cross` for a target that doesn't need Docker (e.g. `x86_64-apple-darwin` on a macOS host), they get:

```
[cross] warning: `cross` does not provide a Docker image for target x86_64-apple-darwin, ...
[cross] error: Errors encountered before cross compilation, aborting.
```

In CI (where warnings are errors), this is confusing — the user may not realize that `cargo` can compile natively for that target without `cross`.

## Solution

After the existing warning, emit an additional note:

- **If target == host**: suggest running `cargo` directly instead of `cross`
- **If target is native but != host**: point to [cross-toolchains](https://github.com/cross-rs/cross-toolchains) for pre-built Docker images

## Changes

`src/lib.rs`: Added platform detection in the `NoCompatibleImages` error path to provide actionable guidance for native targets.

## Testing

- `cargo check` passes
- `cargo test` passes (1 test)

## Related

Fixes #1447

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -771,6 +771,28 @@ pub fn setup(
         Err(err) => {
             msg_info.warn(err)?;
 
+            let target_triple = target.triple();
+            let host_triple = host.triple();
+            let native_targets = [
+                "x86_64-pc-windows-msvc",
+                "x86_64-apple-darwin",
+                "aarch64-apple-darwin",
+            ];
+            if native_targets.contains(&target_triple) {
+                if target_triple == host_triple {
+                    msg_info.note(format_args!(
+                        "target `{target_triple}` is the same as the host; \
+                         try running `cargo` directly instead of `cross`."
+                    ))?;
+                } else {
+                    msg_info.note(format_args!(
+                        "`cross` does not natively support target `{target_triple}`. \
+                         See <https://github.com/cross-rs/cross-toolchains> for pre-built \
+                         Docker images for additional targets."
+                    ))?;
+                }
+            }
+
             return Ok(None);
         }
     };


Emilgardis:
I'm not sure I understand the meaning of putting the --help output in the commit message. Could you fix that? Otherwise this seems fine

LeonidasZhak:
Thanks, that was accidental. I cleaned up the commit message and force-pushed the branch; no code changes.

Emilgardis (approved):

Emilgardis:
Thank you!

Issue #1035, opened 2022-09-29
Windows Builds Fail Due to WineHQ Signing Key


Alexhuszagh:
### Checklist

- [X] I've looked through the [issues and pull requests](https://github.com/cross-rs/cross/issues?q=) for similar reports

### Describe your issue

Attempting to build images using Wine packages from WineHQ currently [fail](https://github.com/cross-rs/cross/actions/runs/3149116522/jobs/5120407159#step:10:1634), producing the output:

```
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 76F1A20FF987672F
Reading package lists...
W: GPG error: https://dl.winehq.org/wine-builds/ubuntu focal InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 76F1A20FF987672F
```

This is because the the Wine servers are being [synchronized](https://wiki.winehq.org/Ubuntu) and changing which sources files are being used.

> - W: GPG error: https://dl.winehq.org/ ... NO_PUBKEY 76F1A20FF987672F
>
> Currently, the Wine servers are being synchronized. This means that some users still getting the older .sources file. If you get this warning, edit the winehq-<distro>.sources file and replace */usr/share/keyrings/* to */etc/apt/keyrings/*.
> 
> For example: `sudo sed -i s@/usr/share/keyrings/@/etc/apt/keyrings/@ /etc/apt/sources.list.d/winehq-focal.sources`

It recommends moving the key to `/etc/apt/keyrings`, rather than `/usr/share/keyrings`, which should probably solve the issue.

### What target(s) are you cross-compiling for?

i686-unknown-linux-gnu, x86_64-pc-windows-gnu

### Which operating system is the host (e.g computer cross is on) running?

- [ ] macOS
- [ ] Windows
- [X] Linux / BSD
- [ ] other OS (specify in description)

### What architecture is the host?

- [X] x86_64 / AMD64
- [ ] arm32
- [ ] arm64 (including Mac M1)

### What container engine is cross using?

- [X] docker
- [ ] podman
- [ ] other container engine (specify in description)

### cross version

cross ba030a6 (latest main)

PR #1779, opened 2026-06-04
Update Docker images to Ubuntu-24.04 and Cargo dependencies [continuation]

MichaIng:
This is a continuation of #1580 with CI all turned green.


.github/workflows/try.yml
```diff
@@ -14,7 +14,7 @@ jobs:
           gh pr comment ${{ github.event.issue.number }} --body "<!--try-ack-comment-->
           Starting try run. [Link to action](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}?pr=${{ github.event.issue.number }})"
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
```
MichaIng, line 17: It took me a while to find the repo setting to grant `GITHUB_TOKEN` write access to PR comments, to leave `/ci try` related comments. The error message without it gave the impression access to `secrets` would be the issue, hence using the identical `github.token` payload was an attempt to fix this on my fork, before I found the repo setting which actually fixed it.

Hence changes to both workflows could be reverted, but `github.token` is used in some places already, and the modern/clearer way to access the token. Hence I suggest to leave it this way. Functionally, both are 100% identical, including their masking in logs etc.

Emilgardis: Agreed, dont know why we went eith the secret here, maybe just habit 😅

ci/shared.sh
```diff
@@ -45,6 +45,7 @@ function mkcargotemp {
     echo '# Cargo.toml
     [workspace]
     members = ["'"$(basename "$td")"'"]
+    resolver = "3"
```
MichaIng, line 48: With `edition = 2024`, the implicit default is `resolver = 3`, while otherwise the default is `resolver = 1`. When using `edition = 2024`, if any workspace does not declare `resolver` explicitly, a warning is emitted stating this. Image test logs in CI hence get very bloated with this. To mute them, I added it here for all test builds.

ci/test-cross-image.sh
```diff
@@ -7,9 +7,13 @@ set -eo pipefail
 if [[ -z "${TARGET}" ]]; then
     export TARGET="aarch64-unknown-linux-gnu"
 fi
-# ^^subst is not supported on macOS bash (bash <4)
+
 # shellcheck disable=SC2155
-export TARGET_UPPER=$(echo "$TARGET" | awk '{print toupper($0)}')
+export TARGET_UPPER=${TARGET_UPPER^^}
```
MichaIng, line 12: Since those old macOS runners are not available anymore (the oldest available x86_64 is `macos-15-intel`), there is no bash <4 in any case. Hence switched to `^^` and used `,,` below.

ci/test-cross-image.sh
```diff
@@ -7,9 +7,13 @@ set -eo pipefail
...
+
+# OCI does not support uppercase characters from orga/owner names
+IMAGE=${IMAGE,,}
+CROSS_IMAGE=${CROSS_IMAGE,,}
```
MichaIng, line 16: Parts of CI failed on my fork since my account name has uppercase letters, and the owner/orga name is used as basis for GitHub container registry access/uploads/image names/tags etc, failing all over the place. GitHub workflow files do not support any syntax to turn `github.repository_owner` and such to lowercase. It can be done in shell in a step, but applies to the current job only. Furthermore, `try.yml` always invokes the default branch version of `ci.yml`, which made it impossible to run vast parts of CI on a PR on my fork. Best solution was to assure relevant variables are lowercase in CI test scripts and Rust code. This makes life easier for people who are testing on their fork.

docker/crosstool-config/arm-unknown-linux-gnueabihf.config
```diff
@@ -42,3 +57,8 @@ CT_ISL_V_0_20=y
 CT_LIBICONV_V_1_15=y
 CT_NCURSES_V_6_1=y
 CT_CC_LANG_FORTRAN=y
+CT_BUILD_MANUALS=n
+CT_REMOVE_DOCS=y
+GLIBC_MAKEINFO_WORKAROUND=y
+CT_LOG_PROGRESS_BAR=n
+CT_LOG_ALL=y
```
MichaIng, line 64: The `arm-unknown-linux-gnueabihf` build was failing, and the only way to find out why was to unmute STDOUT (see removal of `silence_stdout` in `crosstool-ng.sh` below), disable the progress bar (actually a spinner) that otherwise still masks STDOUT with a flood or single character lines, and raise the log level to include actual object install steps. The latter is causing a lot of output lines, so one might want to revert this (remove `CT_LOG_ALL=y`), but on the other hand, this is for our CI only, hence no real reason to be frugal with informational output?

The reason for the failure was the attempt to install (build) a particular manual file from the `gcc` build for Android. The 1st commit (of #1580) lowered the target version from 2.31 to 2.17 (not sure what it has to do with the aim of the PR anyway), and this old gcc version manual has an incompatibility with recent `texinfo`. I found some flags to disable manual builds, but they are actually set by default (`CT_REMOVE_DOCS=y` which implies `CT_BUILD_MANUALS=n`), and have no effect on the gcc build, where manuals are always built and installed, without any native way to disable/skip. But turns out, `crosstool-ng` recognized this issue as well and added an additional flag `GLIBC_MAKEINFO_WORKAROUND=y` to solve exactly this: It patches the old gcc sources to skip manual builds.

docker/android-system.sh
```diff
@@ -126,6 +134,24 @@ main() {
...
+        ! -path './external/llvm/utils/lit/tests/shtest-encoding.py'\
+        -print | while read -r file
+    do
+        LC_ALL=C grep -Iq . "$file" || continue
+        2to3 -w -n "$file"
+    done
+    sed -i 's/lower <= value <= upper/lower <= int(value) <= upper/' ./bionic/libc/fs_config_generator.py
+
+    # Soong looks for a python2.7 executable explicitly
+    ln -s python3 /usr/bin/python2.7
```
MichaIng, line 153: There is a lot of Python 2-only code in these old Android sources, but Ubuntu 24.04 has no Python 2 available anymore. We could pull it from elsewhere, but luckily Ubuntu 24.04 (Python 3.12) is the last one which still contains `2to3` to convert scripts, most importantly `print 'foo'` => `print('foo')`. `2to3` fails on a bunch of scripts, some of them are in weird binary format (most likely embedded binary code), hence they need to be skipped. `.repo/repo/` is pure Python 3 already, but has a lot of scripts `2to3` cannot process, hence it is skipped entirely. All those which caused failures during the Android build can be migrated, with only a single manual `sed` needed (`value` is a numerical str, which was fine for Python 2, but not for Python 3).

An alternative solution would be to switch to newer Android sources, but I didn't want to cause a breaking change, if avoidable.

docker/android-system.sh
```diff
@@ -812,7 +841,7 @@ major = sys.version_info.major
 minor = sys.version_info.minor
 print(f"{major}.{minor}")')
     set +u
-    export PYTHONPATH="${PYTHON_TMPDIR}/lib/python${version}/site-packages/:${PYTHONPATH}"
+    export PYTHONPATH="${PYTHON_TMPDIR}/local/lib/python${version}/dist-packages/:${PYTHONPATH}"
```
MichaIng, line 844: The default path structure when using `python3 -m pip install foo --prefix bar` seems to have changed. `python3 "${PYTHON_TMPDIR}/scripts/build-system.py"` below fails immediately with a meaningful error message if `PYTHONPATH` is wrong, of course.

docker/crosstool-ng.sh
```diff
@@ -90,16 +82,14 @@ main() {
...
     }
 
-    while silence_stdout download; [ $? -eq 124 ]; do
+    while download; [ $? -eq 124 ]; do
         # Indicates a timeout, repeat the command.
         sleep "${sleep}"
     done
-    silence_stdout su "${username}" \
-        -c "CT_DEBUG_CT_SAVE_STEPS=1 ${crosstooldir}/bin/ct-ng build.${nproc}"
+    su "${username}" -c "${crosstooldir}/bin/ct-ng build.${nproc}"
```
MichaIng, line 92: `STOP` is a Makefile option, not an env var, hence it needs to be passed as command argument to be effective. `CT_DEBUG_CT_SAVE_STEPS` should have been a Kconfig flag (the `crosstool-ng` implementation of those), but somehow it has an effect only as env var, while the others like `CT_LOG_PROGRESS_BAR` have an effect only in the config file, not sure why. However, in this case, `CT_DEBUG_CT_SAVE_STEPS=y` causes more issues than that it helps:
- If the first `ct-ng` call fails, sometimes due to some temporary network issue (rate limiting), it does not return exit code 124, hence the 2nd full `ct-ng` call cannot continue at the expected `companion_tools_for_build` step. Even if it starts at the beginning, all steps which were finished before are basically no-ops that finish in no time. So there is no real benefit of starting at `companion_tools_for_build` vs starting at the beginning.
- Before (in `main`), no use of `CT_DEBUG_CT_SAVE_STEPS` was done anyway: It needs to be `=y`, not `=1`, and `ct-ng` only starts at a particular step when using `RESTART=<step>` as argument. Without it, `CT_DEBUG_CT_SAVE_STEPS=y` is just unnecessary overhead.

docker/Dockerfile.aarch64-linux-android
```diff
@@ -1,6 +1,7 @@
-FROM ubuntu:20.04 AS cross-base
+FROM ubuntu:24.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
 
+COPY ubuntu.sources /etc/apt/sources.list.d/
```
MichaIng, line 4: There was so much back and forth and mixed formats with APT sources, causing various errors. Hence I found it easiest to use the new `ubuntu.sources` introduced in #1580 consequently right from the start, which contains all native and foreign architectures, all suites, and components, and `deb-src`, hence nothing needs to be added any time later. Since Ubuntu Docker containers do not ship with `ca-certificates`, HTTPS cannot be used at the beginning, but it is enabled in `common.sh` and individual cases where it is not invoked, right after `ca-certificates` is installed. Whenever another Ubuntu version is used, a `sed` adjusts it right after copying `ubuntu.sources` in place. Full consistency, max security, full compatibility.

Note that the presence of foreign architectures in APT lists does not mean that they can be installed or respective `Packages` files are even pulled. `dpkg --add-architecture` is still needed for that, which is now the one and only needed step to enable additional architectures, even for those which are available in `ports.ubuntu.com` only.

docker/Dockerfile.i586-unknown-linux-gnu
```diff
@@ -14,6 +15,13 @@ RUN apt-get update && apt-get install --assume-yes --no-install-recommends \
     gfortran-i686-linux-gnu \
     libc6-dev-i386-cross
 
+# Make i686 libraries available on host to test target binary
+RUN ln -s /usr/i686-linux-gnu/lib/ld-linux.so.2 /lib/
+RUN ln -s /usr/i686-linux-gnu/lib/libgcc_s.so.1 /lib/
+RUN ln -s /usr/i686-linux-gnu/lib/libc.so.6 /lib/
+RUN ln -s /usr/i686-linux-gnu/lib/libstdc++.so.6 /lib/
+RUN ln -s /usr/i686-linux-gnu/lib/libm.so.6 /lib/
```
MichaIng, line 23: I am not sure why this is needed now, i.e. why `/usr/i686-linux-gnu/lib` is not in default library path for x86 binaries, or whether Ubuntu 20.04 cross compiling libraries (notably `libc6-i386-cross`) placed them below `/lib` directly, or created symlinks in `postinst`. These 5 symlinks turn CI green, but it might be better to do this with all libraries, like
```sh
for i in /usr/i686-linux-gnu/lib/*.so*; do ln -s "$i" /lib/; done
```
Same for the `i686-unknown-linux-gnu‎` target. Let me know, and I'll update it.
```suggestion
# Make i686 libraries available on host to test target binary
RUN for i in /usr/i686-linux-gnu/lib/*.so*; do ln -s "$i" /lib/; done
```
_Is shell syntax that complex fully supported in a Dockerfile? 😅 Else `sh -c '...'`_

docker/Dockerfile.i686-pc-windows-gnu
```diff
@@ -9,26 +10,22 @@ RUN /cmake.sh
 
 FROM cross-base AS build
 
-RUN dpkg --add-architecture i386 && apt-get update && \
-    apt-get install --assume-yes --no-install-recommends libz-mingw-w64-dev
+RUN apt-get install --assume-yes --no-install-recommends libz-mingw-w64-dev
```
MichaIng, line 13: `dpkg --add-architecture i386` is done in `wine.sh` already, where it fits better.

docker/Dockerfile.i686-pc-windows-gnu
```diff
@@ -9,26 +10,22 @@ RUN /cmake.sh
...
 
 COPY wine.sh /
 RUN /wine.sh
 
-# Build mingw tools using dwarf exceptions
-COPY mingw.sh /
-RUN bash mingw.sh
+RUN apt-get install --assume-yes --no-install-recommends g++-mingw-w64-i686-posix gfortran-mingw-w64-i686-posix
```
MichaIng, line 18: As I mentioned in https://github.com/cross-rs/cross/pull/1755#issuecomment-4616039673, the reason for compiling the mingw compilers manually (all which `mingw.sh` does) is obsolete, since Debian/Ubuntu changed the very same thing (switching to dwarf exceptions) their end. Hence I removed it completely and just install the packages directly.

docker/Dockerfile.i686-unknown-linux-gnu
```diff
@@ -22,18 +30,12 @@ RUN TARGET_ARCH=i386 /deny-debian-packages.sh \
...
 RUN /qemu.sh i386 softmmu
 
-COPY dropbear.sh /
-RUN /dropbear.sh
-
-COPY linux-image.sh /
-RUN /linux-image.sh i686
-
-COPY linux-runner base-runner.sh /
+COPY qemu-runner base-runner.sh /
```
MichaIng, line 33: This was done in #1580, switching from linux-runner to qemu-runner, but `targets.toml` was not adjusted, and `linux-runner` was still called via one of the ENV cars below. I finished/fixed this.

But I don't know whether/why the linux-runner (`qemu-system`) failed. Could be checked in the original PR, or just retested. Maybe there is a way to fix that instead of dropping the runner.

But on the other hand, the i586 image was switched to qemu-runner before the PR already, and I could imaging that the reason for doing this applies for the i686 image the same way.

Emilgardis:
Really nice, thank you for doing this. All of those fails seem spurious.

docker/Dockerfile.mips64el-unknown-linux-gnuabi64
```diff
@@ -1,6 +1,9 @@
-FROM ubuntu:20.04 AS cross-base
+FROM ubuntu:22.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
 
+COPY ubuntu.sources /etc/apt/sources.list.d/
+RUN sed -i 's/noble/jammy/g' /etc/apt/sources.list.d/ubuntu.sources
+RUN rm /etc/apt/sources.list
```
MichaIng, line 6: mipsel and mips64el were removed with Debian Trixie, hence `linux-runner.sh` need to use Debian Bookworm userland. The glibc (cross-compile) packages/headers provided by Ubuntu 24.04 are somewhat strict, and do not provide any ABI version supported by Debian Bookworm. Hence the Debian Bookworm userland in the `linux-runner` is unable to execute the builds. I hence reverted to Ubuntu 22.04 in those two cases.

The alternative would be to switch from `linux-runner` to `qemu-runner`, which however degrades performance. Should be fine to stay with an Ubuntu Jammy base image as long as it is maintained (by Canonical). mipsel and mips64el hosts are most likely running old distro versions in most cases, and this keeps compatibility of the resulting builds with those.

docker/Dockerfile.wasm32-unknown-emscripten
```diff
@@ -1,7 +1,10 @@
-FROM emscripten/emsdk:3.1.14
+FROM emscripten/emsdk:3.1.74
 WORKDIR /
 ENV DEBIAN_FRONTEND=noninteractive
 
+COPY ubuntu.sources /etc/apt/sources.list.d/
+RUN sed -i 's/noble/jammy/g' /etc/apt/sources.list.d/ubuntu.sources
+RUN rm /etc/apt/sources.list
```
MichaIng, line 7: `emscripten/emsdk:3.1.74` switched from Ubuntu Focal to Ubuntu Jammy as their base.

docker/Dockerfile.zig
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04 AS cross-base
+FROM ubuntu:24.04 AS cross-base
 ENV DEBIAN_FRONTEND=noninteractive
```
MichaIng, line 3: Seems I missed it here:
```suggestion

COPY ubuntu.sources /etc/apt/sources.list.d/
```

docker/dropbear.sh
```diff
@@ -28,7 +28,7 @@ main() {
 
     pushd "${td}"
 
-    download_mirrors "" "dropbear-${version}.tar.bz2" "${mirrors[@]}"
+    download_mirrors "releases" "dropbear-${version}.tar.bz2" "${mirrors[@]}"
```
MichaIng, line 31: If CI runs into some rate limiting, falling back to the mirror, it failed, as it was missing the `releases` path element. Not sure whether all mirrors use it, but the two applied here do, hence I found it better to pass it as dedicated path argument to `download_mirrors`, which also solves the prior `foo//bar` double slash in the URL, due to the empty path/1st argument.

docker/linux-image.sh
```diff
@@ -57,107 +57,79 @@ max_kernel_version() {
...
+    local debsource="deb https://deb.debian.org/debian trixie main"
+    debsource="${debsource}\ndeb https://security.debian.org trixie-security main"
 
-    local dropbear="dropbear-bin"
-
-    local -a deps
     local kernel=
-    local libgcc="libgcc-s1"
-    local ncurses=
+    local kernel_pkg_split=0
```
MichaIng, line 66: Made some cleanup here, since all qemu-system userland versions are now close enough to share the same package names, and not require different additional dependencies ... aside of the fact that Debian Forky/testing, hence also Sid/unstable split `linux-image-*` packages into 4 separate packages. Needed where we use Debian ports, `kernel_pkg_split` used as flag.

docker/linux-image.sh
```diff
@@ -57,107 +57,79 @@ max_kernel_version() {
...
         kernel="${kversion}-686"
-        deps=(libcrypt1:"${arch}")
-        ;;
-    mips)
-        # mips was discontinued in bullseye, so we have to use buster.
-        libgcc="libgcc1"
-        debsource="deb http://http.debian.net/debian/ buster main"
-        debsource="${debsource}\ndeb http://security.debian.org/ buster/updates main"
-        kernel='4.*-4kc-malta'
-        ncurses="=6.1*"
```
MichaIng, line 95: Already in `main` there is no mips target which uses the `linux-runner`, and it wouldn't be possible (was the reason for those switching to `qemu-runner`) for the same reason we cannot use a Debian Bookworm mipsel/mips64el runner on Ubuntu 24.04: the provided glibc ABI versions do not overlap. While getting an overview I was confused about the `mips` entry here, hence removing it from this script for clarity. This cannot work ever again for any mips target.

docker/linux-image.sh
```diff
@@ -57,107 +57,79 @@ max_kernel_version() {
...
-
-        dropbear="dropbear"
-        deps=(libcrypt1:"${arch}")
+        # there is no stable port
+        # https://deb.debian.org/debian-ports/pool-powerpc/main/l/linux/
+        kernel='7.*-powerpc'
+        debsource="deb https://deb.debian.org/debian-ports unstable main"
+        debsource="${debsource}\ndeb https://deb.debian.org/debian-ports unreleased main"
+        # Debian Forky splits linux-image-* packages into linux-binary-* and linux-modules-* and linux-base-*
+        kernel_pkg_split=1
```
MichaIng, line 99: This was confusing: Adding ancient Jessie sources but also Debian ports unstable, which means that all packages are pulled from unstable. Was it only for the kernel version? For which reason? Debian ports unstable ships a recent powerpc kernel, not sure why anyone would want Linux 4.9, overriding all APT security checks etc. I hence aligned this with powerpc64, sparc64, and other targets which are available from Debian ports, which implies a switch to Linux 7.x and split kernel packages.

docker/linux-image.sh
```diff
@@ -57,107 +57,79 @@ max_kernel_version() {
...
         arch=ppc64el
-        kernel='5.*-powerpc64le'
-        deps=(libcrypt1:"${arch}")
+        kernel="${kversion}-powerpc64le"
         ;;
     riscv64)
-        kernel='6.*-riscv64'
-        debsource="deb http://deb.debian.org/debian unstable main"
-        deps=(libcrypt1:"${arch}")
+        kernel="${kversion}-riscv64"
```
MichaIng, line 116: `riscv64` is in stable Debian Trixie, no unstable needed anymore.

docker/linux-image.sh
```diff
@@ -198,15 +170,11 @@ main() {
...
-
-    for key in *.asc *.key; do
-        apt-key add "${key}"
-        rm "${key}"
-    done
+    cd /etc/apt/trusted.gpg.d
+    curl --retry 3 -sSfL 'https://ftp-master.debian.org/keys/archive-key-{10,11,12,13}.asc' -O
+    curl --retry 3 -sSfL 'https://ftp-master.debian.org/keys/archive-key-{10,11,12,13}-security.asc' -O
+    curl --retry 3 -sSfL 'https://ftp-master.debian.org/keys/release-{10,11,12,13}.asc' -O
+    curl --retry 3 -sSfL 'https://www.ports.debian.org/archive_2026.key' -o archive_2026.asc
```
MichaIng, line 177: `apt-key` is deprecated, and was removed with Debian Trixie. Ubuntu 24.04 still ships it, but it is obsolete since a decade or so: even EOL Debian/Ubuntu versions support `*.asc` keys in `/etc/apt/trusted.gpg.d` natively, no `apt-key` and no `gpg --dearmor` needed.

docker/linux-image.sh
```diff
@@ -389,12 +370,6 @@ mount -t 9p -o trans=virtio target /target -oversion=9p2000.u || true
 exec dropbear -F -E -B
 EOF
 
-    if [[ "${arch}" == "riscv64" ]]; then
-        # Symlink dynamic loader to /lib/ld-linux-riscv64-lp64d.so.1
-        mkdir -p "${root}/lib"
-        ln -s /usr/lib/riscv64-linux-gnu/ld-linux-riscv64-lp64d.so.1 "${root}/lib/ld-linux-riscv64-lp64d.so.1"
-    fi
```
MichaIng, line 396: As mentioned in #1580: After setting up the runner userland `usrmerge`d above, this step for `riscv64`/trixie is redundant, and fails as the target file exists already.

docker/solaris.sh
```diff
@@ -58,7 +58,8 @@ main() {
     esac
 
     apt-key adv --batch --yes --keyserver keyserver.ubuntu.com --recv-keys 74DA7924C5513486
-    add-apt-repository -y 'deb http://apt.dilos.org/dilos dilos2 main'
+    # https://apt.dilos.org/dilos is currently down
+    add-apt-repository -y 'deb https://mirrors.dotsrc.org/mirrors/pub/dilos/apt dilos2 main'
```
MichaIng, line 62: DilOS seems dead. Their website and APT repo are down. The mirror however is still live.

docker/solaris.sh
```diff
@@ -136,6 +137,26 @@ EOF
...
+
+// Create strong aliases that the Rust compiler can see
+long __sysconf_xpg7(int name) {
+    return __sysconf_xpg5(name);
+}
+int __xnet7_socket(int domain, int type, int protocol) {
+    return __xnet_socket(domain, type, protocol);
+}
+_EOF_
+    "${target}-gcc" -c compat.c -o /compat.o
```
MichaIng, line 158: This was broken before. Sadly there is no other APT-based Solaris distro, at least none with an actual APT repo to pull packages from. And DilOS packages are too old to be fully compatible with recent Rust. So I went down the rabbit hole and tried all ways to map or alias modern symbols Rust is expecting with the older ones present in DilOS libc. This was the only way that worked: a dedicated C object, used as linker script, which is added in `test.sh` via `RUSTFLAGS+=' -C link-arg=/compat.o'` for Solaris targets.

However, it would better if this was done by default, passed via Dockerfile ENV? But I couldn't find a way to do that, the documented way does not work for me:
- https://github.com/cross-rs/cross/blob/main/docs/cargo_configuration.md?plain=1#L14-L24
- I tried to pass `ENV CARGO_TARGET_X86_64_PC_SOLARIS_RUSTFLAGS="-C link-arg=foo"` with `Dockerfile.x86_64-pc-solaris`, but the flags did not appear in the `cargo` calls of the tests. Not sure if I was doing something wrong, or if `RUSTFLAGS` is overwritten in our CI, or if it is broken?

Anyway, for now, applying the needed flag to `RUSTFLAGS` in `test.sh` for Solaris targets, turns CI green, and at least provides a way for users to solve these issues via custom linker flag. I saw the alternative is to let `cross` pull older Rust, with `cross +<version>`.

docker/wine.sh
```diff
@@ -7,25 +7,20 @@ set -euo pipefail
...
+    local version="10.*~noble-1"
 
     # workaround for wine server synchronization, see #1035
     # we need to ensure the keys are now stored in `/etc/apt/keyrings`,
     # which were previously stored in `/usr/share/keyrings`, and ensure
     # our sources list searches for the right location.
     mkdir -p /etc/apt/keyrings
-    mv winehq.key /etc/apt/keyrings/winehq-archive.key
-
-    wget -nc https://dl.winehq.org/wine-builds/ubuntu/dists/focal/winehq-focal.sources
```
MichaIng, line 26: I found it weird that `wget` is installed only to download a key and a list, while `curl` is present and used with consistent 3 retries everywhere else. I hence aligned that here, and reordered things to make more sense, like define the WINE version below the comment which talks about the hardcoded version.

The line `# workaround for wine server synchronization, see #1035` above is actually a bit misleading, as if installing the key for a 3rd party APT repo was a "workaround", rather than a totally normal step. It would be also possible to download the key to `/usr/share/keyrings`. Not common standard for non-packaged keys, but APT does not enforce or exclude any path, as long as it matches the `signed-by` in the APT list. Anyway, I left that as is.

src/docker/shared.rs
```diff
@@ -23,16 +23,10 @@ use std::{env, fs, time};
...
 
 pub use super::custom::CROSS_CUSTOM_DOCKERFILE_IMAGE_PREFIX;
-pub const CROSS_IMAGE: &str = {
-    if let Some(opt) = option_env!("CROSS_IMAGE") {
-        opt
-    } else {
-        "ghcr.io/cross-rs"
-    }
-};
+pub const CROSS_IMAGE: &str = env!("CROSS_IMAGE_LOWER");
```
MichaIng, line 26: Since a constant naturally cannot be altered (to turn `CROSS_IMAGE` lowercase if the repo owner/orga is not), and there is no real way to do that before assigning the env var with the type it is returned as, the `CROSS_IMAGE` is now obtained in `build.rs`, turned lowercase there, and passed as `CROSS_IMAGE_LOWER` to be used here.

Cargo.toml
```diff
@@ -28,43 +28,43 @@ members = ["xtask"]
...
 shell-words = "1.1.0"
 const-sha1 = "0.3"
 signal-hook = { version = "0.4" }
 directories = "6.0"
 walkdir = { version = "2.3.2", optional = true }
-tempfile = "3.3.0"
-owo-colors = { version = "4.0", features = ["supports-colors"] }
-semver = "1.0.16"
-is_ci = "1.1.1"
+tempfile = "=3.24.0"
```
MichaIng, line 48: Hardcoding tempfile `3.24.0` is needed, otherwise cargo pulls v3.27.0 which pulls in dependencies that depend on Rust v1.88.0. The next higher v3.25.0 depends on v1.87.0.

This `Cargo.toml` however declares `rust-version = "1.85.0"`, which we probably want to keep? Restoring MSRV 1.85.0 caused rustfmt to revert some formatting changes applied in #1580. I aligned `edition = "2024"` and `rust-version = "1.85.0"` across "cross" and xtask, for a consistent `cargo update` without warnings.

deny.toml
```diff
@@ -16,6 +16,8 @@ deny = []
 skip = [
   # home v0.5.11
   { name = "windows-sys", version = "0.59.0" },
+  # toml v0.9 uses winnow 0.7 directly, while toml_parser (sub-dep of toml) uses winnow 1.0
+  { name = "winnow", version = "0.7" },
```
MichaIng, line 20: CI checks for doubled 2nd level dependencies, if different versions are pulled in by 1st level dependencies. There is already the `windows-sys` exception caused by `home v0.5.11`, hence I added an exception for `winnow` the same way, as `toml v0.9` pulls in two versions of it with two of its own dependency tails 😄.

targets.toml
```diff
@@ -5,7 +5,7 @@
 
 [[target]]
 target = "x86_64-apple-darwin"
-os = "macos-13"
+os = "macos-15-intel"
```
MichaIng, line 8: The `macos-13` runner has been removed a longer while ago, the `macos-14-intel` runner quite recently. `macos-14` exists but is Apple Silicon. `macos-15-intel` is hence the oldest freely available hosted macOS x86_64 runner on GitHub.

For completeness: With an enterprise subscription, there would be `macos-14-large` available, which is x86_64 as well.

targets.toml
```diff
@@ -31,7 +31,7 @@ deploy = true
 
 [[target]]
 target = "x86_64-pc-windows-msvc"
-os = "windows-2019"
+os = "windows-2022"
```
MichaIng, line 34: `windows-2019` was removed a while ago. `windows-2022` is now the oldest available Windows runner.

targets.toml
```diff
@@ -410,14 +410,12 @@ run = true
...
 target = "x86_64-pc-windows-gnu"
 os = "ubuntu-latest"
-cpp = true
 std = true
 run = true
 
 [[target]]
 target = "i686-pc-windows-gnu"
 os = "ubuntu-latest"
-cpp = true
```
MichaIng, line 420: As I memtioned in https://github.com/cross-rs/cross/pull/1755#issuecomment-4616039673, `cargo --features=re2` causes a NULL pointer deference now. Not sure about the underlying reason, or whether switching back to self-compiled/patched mingw compilers would solve it. However, #1755 also runs into failing CI as far as I can see, even the same it seems.

Removing `cpp = true` => no `--features=re2` tests, fixes things. Of course I can re-enable it if someone wants to debug and find a different solution.

MichaIng:
> Really nice, thank you for doing this. All of those fails seem spurious.

@Emilgardis 
Right, I wanted to mention this: When building/testing all targets concurrently, there will be _always_ some failures due to network rate limiting, usually 502 HTTP response codes, failing TLS handshakes, 403 in some cases. You can get all green by selecting "rerun all failed jobs" from the runs GUI, sometimes even needed a second time afterwards.

Naturally, rerunning all jobs again (like with a new `/ci try`) will just run into the same or more network errors 😄.

Emilgardis (approved):
I want to merge this, @cross-rs/maintainers fyi

Emilgardis:
Thank you @MichaIng for continuing this. Ill let this be for a bit more and then merge 😊

MichaIng:
Yeah maybe someone else finds time to look through the code comments, as some polishing could be done, or things solved in different ways. But can be also done in follow-ups, if issues appear.

Maybe trigger a rerun of failed CI jobs, just to be sure all is green as expected? 🙂

Emilgardis:
I thought I started a rerun but apparently not, thank you :)

The changes are good and bring the project to a working state, and small fixes or refactoring can be worked on afterwards

MichaIng:
This one I meant: https://github.com/cross-rs/cross/actions/runs/26971729604

MichaIng:
You can trigger a rerun of failed jobs from the same workflow instance. That way, jobs already succeeded do not unnecessarily run again, and you can get sure all is green, when not running into these network rate limiting while a large number of target job run in parallel 🙂:

{1B3A131C-349C-428F-9976-1BA531ECC189}

Emilgardis:
I'm aware thank you, reran zig explicitly because the action will checkout the old code pre a2921b5

MichaIng:
Rerunning jobs from a workflow always pulls the latest code from the same branch. The only thing that remains on old code state is the workflow YAML files themselves. Hence, triggering a new workflow should only be needed when those are changed, and AFAIK as well when secrets are changed.

Emilgardis:
Aha, you're right, I was under the impression I had exclicitly set the checkout-ref though to be the current sha, seems we havent :D

image

MichaIng:
Luckily it is the PR ref, which made dev-test cycles on my fork a lot easier 😄. I still felt a bit guilty to cause GitHub so much runner load and traffic, as some targets required a lot of iterations to find out and solve all issues 😅.

EDIT: This rate limiting towards GitHub runners is really an issue. I wished VPS providers would whitelist or relax this for GitHub IPs. On my main project, we check for latest software versions once a day, also to check whether related projects are still alive, and whether URL structures or filename schemes changed. A particular one fails in like 70% of cases with 502 or 403, even that it is a tiny plain text document with like 20 characters body that is requested, causing probably less traffic/load then the error response, and of course a retry is then needed. And that rate limiting is not in the hands of the software dev/maintainer, but the VPS provider he uses 😞.

Emilgardis:
Thank you, going to merge this! (without CI since we've already done CI :) )

Emilgardis:
and thank you @JanDiederich and @itpetey for helping with this!

PR #1785, opened 2026-07-01
enh: add all Ubuntu ports architectures to sources

MichaIng:
If a particular architecture is not available on a particular Ubuntu version, APT emits a notice, but does not fail. However, those 3 added architectures are available on Ubuntu 24.04 and 22.04 ports both, hence on all currently use Ubuntu base images.

That only the ARM architectures were present looks more like an oversight than intended. An alternative would be to remove the `Architectures:` line. But then, APT would query the repo even if no matching `dpkg --add-architecture` was done, and throw a warning if it does not support the native one, i.e. x86_64.

### DIFF
--- a/docker/ubuntu.sources
+++ b/docker/ubuntu.sources
@@ -8,7 +8,7 @@ Signed-By: /usr/share/keyrings/ubuntu-archive-keyring.gpg
 Types: deb deb-src
 URIs: http://ports.ubuntu.com/ubuntu-ports
 Suites: noble noble-updates noble-backports noble-security
-Architectures: arm64 armhf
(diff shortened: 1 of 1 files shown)


Emilgardis:
would it maybe make more sense to alter this in the linux-image.sh script?

MichaIng:
I was also thinking whether to apply the target arch only in `common.sh`. But the original idea seems to have been to have a complete `ubuntu.sources` that works OOTB for all targets without the need to alter anything afterwards (aside of naturally the HTTPS <> `ca-certificates` chicken&egg issue). And since additional architectures do not have any implications, I think it is the right/simpler approach.

Emilgardis (approved):

PR #1792, opened 2026-07-31
fix(remote): explicitly check for symlink before calling read_link

### LINKED ISSUE #1791: [CROSS_REMOTE] fs::read_link fails with 'Invalid argument (os error 22)' when encountering UNIX domain sockets
### Problem Description

When running `cross build` with `CROSS_REMOTE=1`, `cross` copies project files to the remote container daemon via `copy_dir` in `src/docker/remote.rs`.

During `copy_dir` directory traversal, entries that return `false` for both `file.file_type()?.is_file()` and `file.file_type()?.is_dir()` fall into the `else if copy_symlinks` branch, assuming any non-file non-directory entry must be a symlink:

https://github.com/cross-rs/cross/blob/main/src/docker/remote.rs#L473-L481

```rust
if file.file_type()?.is_file() {
    fs::copy(&src_path, &dst_path)
        .wrap_err_with(|| format!("when copying file {src_path:?} -> {dst_path:?}"))?;
} else if file.file_type()?.is_dir() {
    fs::create_dir(&dst_path).ok();
    had_symlinks = copy_dir(&src_path, &dst_path, copy_symlinks, depth + 1, skip)?;
} else if copy_symlinks {
    had_symlinks = true;
    let link_dst = fs::read_link(src_path)?;
```

If the workspace contains other Unix non-regular files—specifically **UNIX domain sockets** (`*.sock`, `S_IFSOCK`) or FIFO pipes (e.g., daemon sockets from local tools like `.codegraph/daemon.sock`)—calling `fs::read_link` on a socket file on macOS (or Linux) returns `EINVAL` (`os error 22: Invalid argument`), causing `cross` to fail:

```
Error: 
   0: could not run container
   1: could not complete remote run
   2: when copying project
   3: Invalid argument (os error 22)

Location:
   src/docker/remote.rs:481
```

---

### Reproduction Steps

1. Create a socket file in a subfolder inside the Rust project workspace (e.g. `nc -lU daemon.sock` or running a daemon creating a `.sock` file).
2. Run `CROSS_REMOTE=1 DOCKER_HOST=... cross build --target x86_64-unknown-linux-gnu`.
3. `cross` attempts to copy the project workspace and crashes at `remote.rs:481`.

---

### Temporary Workaround

Add a [`CACHEDIR.TAG`](https://bford.info/cachedir/) file to the directory containing socket files with the content:
```
Signature: 8a477f597d28d172789f06886806bc55
```
This forces `is_cachedir()` to return `true`, skipping directory recursion for that folder.

---

### Suggested Solution

1. **Check for `is_symlink()` explicitly**: Before calling `fs::read_link(src_path)`, check `file.file_type()?.is_symlink()`. Sockets, FIFOs, and character/block devices should be skipped or ignored rather than assumed to be symlinks.
2. **Respect `.gitignore`**: `copy_files_nocache` currently traverses all directories regardless of git status or `.gitignore`. `cross` should respect `.gitignore` rules (merging both repository-level `.gitignore` and user-level global gitignore from `core.excludesfile` / `git check-ignore`) when gathering files to copy under `CROSS_REMOTE=1`.

cathaysia:
### Summary

Fixes an issue where `cross build` with `CROSS_REMOTE=1` crashes with `Invalid argument (os error 22)` when encountering non-regular, non-directory, non-symlink entries in the workspace (such as UNIX domain sockets `*.sock` or FIFO pipes).

### Details

In `src/docker/remote.rs`, `copy_dir` iterates over directory entries. Previously, any entry where `file_type.is_file()` and `file_type.is_dir()` returned `false` fell through to `else if copy_symlinks`, assuming it was a symlink and executing:

```rust
let link_dst = fs::read_link(src_path)?;
```

On macOS/Unix, executing `readlink` on a UNIX domain socket (e.g., daemon sockets created by local tools like `.codegraph/daemon.sock`) returns `EINVAL` (`os error 22: Invalid argument`), causing `cross` to fail.

This PR adds an explicit `file_type.is_symlink()` check before attempting `fs::read_link`. Sockets, FIFOs, and other special Unix devices will fall through to the `else` block and be safely ignored without crashing.

Fixes #1791

### DIFF
--- a/src/docker/remote.rs
+++ b/src/docker/remote.rs
@@ -470,13 +470,14 @@ where
 
         let src_path = file.path();
         let dst_path = dst.join(file.file_name());
-        if file.file_type()?.is_file() {
+        let file_type = file.file_type()?;
+        if file_type.is_file() {
             fs::copy(&src_path, &dst_path)
                 .wrap_err_with(|| format!("when copying file {src_path:?} -> {dst_path:?}"))?;
-        } else if file.file_type()?.is_dir() {
+        } else if file_type.is_dir() {
             fs::create_dir(&dst_path).ok();
             had_symlinks = copy_dir(&src_path, &dst_path, copy_symlinks, depth + 1, skip)?;
-        } else if copy_symlinks {
+        } else if file_type.is_symlink() && copy_symlinks {
             had_symlinks = true;
             let link_dst = fs::read_link(src_path)?;
 


Emilgardis (approved):
Thank you!

PR #1793, opened 2026-08-07
feat(remote): copy only build artifacts back from remote builds

cathaysia:
Instead of copying the entire target directory back from the container (including the incremental build cache), parse cargo's --message-format=json output for compiler-artifact filenames and copy only those files via docker cp. Falls back to the full directory copy when no artifacts are parsed.

after:

image

### DIFF
--- a/.changes/1793.json
+++ b/.changes/1793.json
@@ -1,0 +1,5 @@
+{
+    "type": "added",
+    "description": "add CROSS_REMOTE_COPY_FULL_TARGET_DIR to copy the full target directory back from remote builds when build-script artifacts are needed on the host",
+    "issues": [1793]
+}
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -33,6 +33,10 @@ In-depth documentation with examples can be found [here][env-examples].
 - `CROSS_REMOTE`: Inform `cross` it is using a remote container engine, and use
   data volumes rather than local bind mounts. See [Remote][docs-remote] for
   more information using remote container engines.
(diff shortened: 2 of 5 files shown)


cathaysia:
Can someone review this?

Emilgardis:
I like this idea, however, does this pickup loose artifacts from build-scripts? If not, there should be a way to always copy full targetdir

cathaysia:
Thanks for the review! Regarding build-script artifacts: rather than special-casing `OUT_DIR`, this PR now adds an opt-in env var that copies the entire target directory back to the host.

**New feature: `CROSS_REMOTE_COPY_FULL_TARGET_DIR`**

- When set, cross skips `--message-format=json` (and the per-artifact parsing) and copies the whole target directory back to the host (`docker cp -a` of the full target dir), so anything a `build.rs` generated — `OUT_DIR` contents, generated code, bindings — is available on the host.
- It is orthogonal to `CROSS_REMOTE_SKIP_BUILD_ARTIFACTS`: if skip is set, nothing is copied back.

Emilgardis:
Perfect, can you document this in https://github.com/cross-rs/cross/blob/main/docs/environment_variables.md please!

cathaysia:
> Perfect, can you document this in https://github.com/cross-rs/cross/blob/main/docs/environment_variables.md please!

Ok, I add the `CROSS_REMOTE_SKIP_BUILD_ARTIFACTS` to document.

Emilgardis (approved):
lgtm, ty!

PR #1799, opened 2026-08-26
Bump musl libc and gcc version for *-musl targets

### LINKED ISSUE #1794: Bump musl targets to musl 1.2.5 and GCC 14.3.0
Some of the existing musl targets in `cross` are still built with `musl-cross-make` using relatively old toolchain versions, currently around:

* musl 1.2.3
* GCC 9.2.0
* binutils 2.33.1

One practical issue with the current GCC version is its limited support for newer C/C++ standards, especially C++20. Some modern C/C++ dependencies now require newer compiler features, which can make them difficult or impossible to build with the existing musl images.

I am considering updating the older musl targets and would like to ask for maintainers' preference on two points:

1. **Which versions should we target?**

   Would `musl 1.2.5 + GCC 14.2.0` be a reasonable baseline, matching the newer musl targets already used in `cross`?

   Alternatively, would you prefer a different GCC or musl version for compatibility reasons?

2. **Should the existing musl targets continue using `musl-cross-make`, or should they be migrated to `crosstool-ng`?**

   The current `musl-cross-make` based setup is relatively simple and already works for these targets, so updating the versions there would result in a smaller change.

   On the other hand, `cross` already uses `crosstool-ng` for several newer GNU and musl targets, so migrating the older musl targets could make toolchain configuration more consistent in the long term.

My current thought is either:

```text
Option A:
keep musl-cross-make
musl 1.2.3 -> 1.2.5
GCC  9.2.0 -> 14.3.0
```

Before working on this, I would like to confirm which direction and version combination would be preferred.

Dirreke:
Close #1794 

This PR updates the existing *-musl targets to use a newer musl toolchain while keeping the previous toolchain available for compatibility.

 - Migrate the affected musl targets from musl-cross-make to crosstool-ng.
 - Upgrade the default toolchain from musl 1.2.3 / GCC 9.2.0 to musl 1.2.5 / GCC 14.3.0.
 - Keep the previous musl 1.2.3 / GCC 9.2.0 toolchain available through :legacy images.
 - Refresh the target/toolchain table in the README and document how to use the legacy images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,73 +200,81 @@ QEMU gets upset when you spawn multiple threads. This means that, if one of your
 unit tests spawns threads, then it's more likely to fail or, worst, never
 terminate.
 
-| Target                                 |  libc  |  GCC   | C++ | QEMU  | `test` |
-|----------------------------------------|-------:|-------:|:---:|------:|:------:|
-| `aarch64-linux-android` [1]            | 9.0.8  | 9.0.8  | ✓   | 6.1.0 |   ✓    |
-| `aarch64-unknown-linux-gnu`            | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
-| `aarch64-unknown-linux-gnu:centos` [7] | 2.17   | 4.8.5  |     | 4.2.1 |   ✓    |
-| `aarch64-unknown-linux-musl`           | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
-| `aarch64_be-unknown-linux-gnu`         | 2.36   | 14.2.0 | ✓   | 6.1.0 |   ✓    |
-| `arm-linux-androideabi` [1]            | 9.0.8  | 9.0.8  | ✓   | 6.1.0 |   ✓    |
-| `arm-unknown-linux-gnueabi`            | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
-| `arm-unknown-linux-gnueabihf`          | 2.31   | 8.5.0  | ✓   | 6.1.0 |   ✓    |
-| `arm-unknown-linux-musleabi`           | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
-| `arm-unknown-linux-musleabihf`         | 1.2.3  | 9.2.0  | ✓   | 6.1.0 |   ✓    |
-| `armv5te-unknown-linux-gnueabi`        | 2.31   | 9.4.0  | ✓   | 6.1.0 |   ✓    |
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -283,13 +304,17 @@ terminate.
...
-[6] libc = emscripten and GCC = clang
+[6] libc = emscripten and the GCC/Clang column shows the bundled Clang version.
+    GFortran is not included.
 
 [7] Must change
     `image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main-centos"` in
     `Cross.toml` for `[target.aarch64-unknown-linux-gnu]` to use the
     CentOS7-compatible target.
 
+[8] To use the legacy GCC 9.2 musl toolchain, set `image = "ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main-legacy"` in the
```
Emilgardis: we have a quick way to refer to these targets using just `-<sub>`
```suggestion
[8] To use the legacy GCC 9.2 musl toolchain, set `image = "-legacy"` in the
```

Emilgardis (commented):
Thank you for the PR! I'm open to merge this, however it will double CI time for musl. Not sure if I like that. Will have to think about it

Dirreke:
Thanks for the feedback! IMO, the `*-legacy` targets may not be necessary. For the `*-musl` targets, libc is statically linked, so the resulting binaries do not have a runtime dependency on the target system's libc version. Would it make sense to first try upgrading the existing `*-musl` targets without keeping the legacy images? This would also avoid doubling the musl CI time.

BTW, would it be possible to first run `/ci try` on the current changes, so I can check for any build or CI errors and finish the remaining work?

Emilgardis:
Of course! My bad forgot :)

/ci try --target *-musl*

And yes, im ok with omitting the legacy targets

Dirreke:
It seems that [https://www.musl-libc.org](https://www.musl-libc.org/releases/musl-1.2.5.tar.gz) is not stable.

Dirreke:
@Emilgardis Would you like to give it another try?

Emilgardis:
```
/ci try --target *-musl*
```

Dirreke:
Looks like CI didn’t start successfully.

Dirreke:
@Emilgardis , Please help try x86_64-unknown-linux-musl again,thanks.

Dirreke:
fixed

Dirreke:
@Emilgardis Could you help try again?

Dirreke:
Still network issue. Lets try it again tomorrow.

Dirreke:
@Emilgardis

PR #1801, opened 2026-09-07
feat(remote): support .dockerignore when copying files to remote host

### LINKED ISSUE #1800: Support .dockerignore
I think we should probably support .dockerignore files.

cathaysia:
Support `.dockerignore` files when copying project and mounted volumes to remote container data volumes.

### Changes
- Add `docker_ignore` module for `.dockerignore` parsing and pattern matching based on the `glob` crate:
  - Supports comments (`#`), blank lines, and UTF-8 BOM.
  - Supports negation / exception rules (`!`) in sequential order.
  - Supports directory-only patterns (`/` suffix).
  - Matches root-relative and recursive (`**`) patterns using `glob::MatchOptions { require_literal_separator: true, .. }` following Docker / Moby `patternmatcher` specifications.
  - Respects parent directory exclusion hierarchy (excluding a parent directory prunes children).
- Integrate `.dockerignore` filtering into directory copying (`copy_dir_with_rel` and `copy_files_filtered`) in `remote.rs`, skipping ignored files and directories during traversal.
- Integrate `.dockerignore` filtering into `Fingerprint` scanning so persistent volume incremental sync respects ignore rules.
- Add comprehensive unit tests covering parsing, pattern matching, exception rules, directory copying, and fingerprint scanning.

close #1800

### DIFF
--- a/src/docker/docker_ignore.rs
+++ b/src/docker/docker_ignore.rs
@@ -1,0 +1,368 @@
+use std::fs;
+use std::path::Path;
+use std::str::FromStr;
+
+use eyre::Context;
+
+use crate::errors::Result;
+use crate::file::PathExt;
+
+#[derive(Debug, Clone)]
+pub struct DockerIgnoreRule {
+    pub pattern: glob::Pattern,
+    pub is_exception: bool,
+    pub only_dir: bool,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct DockerIgnore {
+    pub rules: Vec<DockerIgnoreRule>,
+}
+
+impl DockerIgnore {
+    pub fn empty() -> Self {
+        Self { rules: Vec::new() }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.rules.is_empty()
+    }
+
+    pub fn from_dir(dir: &Path) -> Result<Self> {
+        let ignore_file = dir.join(".dockerignore");
+        if ignore_file.is_file() {
+            Self::from_path(&ignore_file)
+        } else {
+            Ok(Self::empty())
+        }
+    }
+
+    pub fn from_path(path: &Path) -> Result<Self> {
+        let content = fs::read_to_string(path)
+            .wrap_err_with(|| format!("when reading dockerignore file {path:?}"))?;
+        Self::parse(&content)
+    }
+
+    pub fn parse(content: &str) -> Result<Self> {
+        let content = content.strip_prefix('\u{feff}').unwrap_or(content);
+        let mut rules = Vec::new();
+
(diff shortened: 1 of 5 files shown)


cathaysia:
Hi, could someone please review this?

src/docker/docker_ignore.rs
```diff
@@ -0,0 +1,368 @@
...
+        let content3 = "target/**\n!target/keep.txt\n";
+        let di3 = DockerIgnore::parse(content3).unwrap();
+        assert!(!di3.is_ignored("target", true));
+        assert!(!di3.is_ignored("target/keep.txt", false));
+        assert!(di3.is_ignored("target/delete.txt", false));
+    }
+
+    #[test]
+    fn test_dockerignore_dir_exception() {
+        let content = "target/\n!target/keep.txt\nlogs/\n!logs/**/*.log\nbuild/\n";
```
Emilgardis: These strings should be readable, use "#

src/docker/docker_ignore.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub fn empty() -> Self {
+        Self { rules: Vec::new() }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.rules.is_empty()
+    }
+
+    pub fn from_dir(dir: &Path) -> Result<Self> {
+        let ignore_file = dir.join(".dockerignore");
```
Emilgardis: Should support .containerignore as an alternative filename

Emilgardis (changes requested):
I think this is fine, some minor nits. One thing this makes me consider is if we should just switch to use ignore which is based on walkdir to traverse the mount

src/docker/docker_ignore.rs
cathaysia: Ok, I add some logic just like this:

```python
if engine is podman and "containerignore" exists:
    return "containerignore"
elif "dockerignore" exists:
    return "dockerignore"
else:
    return "containerignore"
```

and print log to shell when using a fallback logic:

```
if ignore_file not is "containerignore" and engine == podman:
    log!()
if ignore_file is "containerignore" and engine != podman:
    log!()
```

cathaysia:
> I think this is fine, some minor nits. One thing this makes me consider is if we should just switch to use ignore which is based on walkdir to traverse the mount

Considering that most of the traversal work involves pruning, using walkedir may not be as beneficial as we thought. I changed the recursion to iteration to avoid potential stack overflow issues.

cathaysia:
Is there anything else here that needs to be modified?

Emilgardis:
No, just need to have a look at the changes once more and then can merge :)

Emilgardis (approved):
Thank you!

PR #1802, opened 2026-09-11
Update toml dependency version to 1.0

chirizxc:
Rust `Cargo.toml` support TOML 1.1


Emilgardis:
needs a cargo update -p toml

